PR #6, opened 2018-07-31
*kashike waves his wand*

kashike:


astei:
Yeah, magic numbers get the best of me. Anyway, 👍

PR #12, opened 2018-08-05
Add default config

Desetude:
This should make it a bit easier to start out.


astei:
We actually had a default config but it got lost during the shuffling around. It's fixed now.

PR #14, opened 2018-08-05
Add fallback server support.

### LINKED ISSUE #11: Implement fallback connection support
Velocity does not yet currently support attempts to fall back to other servers in the `try` list. Supporting this would be a very easy project to knock out for a new contributor.

DylanKeir:
Resolves #11 - I'm sure there's a better way to do this, but I'm not too familiar with the code base at the moment. However, it works.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
@@ -240,7 +240,7 @@ public class ConnectedPlayer implements MinecraftConnectionAssociation, Player {
             connect()
                     .whenComplete((status, throwable) -> {
                         if (throwable != null) {
-                            if (throwable instanceof ConnectException) {
+                            if (throwable instanceof ConnectException && connectedServer == null) {
                                 Optional<ServerInfo> nextServer = getNextServerToTry();
                                 if (nextServer.isPresent()) {
                                     createConnectionRequest(nextServer.get()).fireAndForget();
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
@@ -27,6 +27,7 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
 import javax.annotation.Nonnull;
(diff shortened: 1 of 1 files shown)


proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
```diff
@@ -239,6 +240,13 @@ public void fireAndForget() {
             connect()
                     .whenComplete((status, throwable) -> {
                         if (throwable != null) {
+                            if (throwable instanceof ConnectException) {
+                                Optional<ServerInfo> nextServer = getNextServerToTry();
+                                if (nextServer.isPresent()) {
+                                    createConnectionRequest(nextServer.get()).fireAndForget();
```
astei: This is problematic: This would trigger even when the player was trying to connect to another server. Can you add in a check to ensure that this could only be triggered when the player is not yet connected to another server?

Additionally, I have concerns about this logic being in `fireAndForget()` - this is going to cover only connection exceptions, not kicks from servers.

DylanKeir: I have pushed a new commit which would check that the player is not currently connected to a server. However, as for relocating it, I'm unsure as to a better place to put it as the exception is handled in `fireAndForget()`. If by you mean "not kicks from servers" you mean to handle if a player is kicked, move then back to a default server, I could do that.

electronicboy: it's more during connection, the two major reasons why you'd wanna try the next server is because a connection error occured, or because the server kicked during login for some reason, e.g. "wrong client version"

hugmanrique: I would move this logic to #handleConnectionException(ServerInfo info, Component disconnectReason)

astei: @hugmanrique is right here. You should do this in the `handleConnectionException` method. It's called any time there's a disconnect from the server and it's the best place to put this code.

astei:
@DylanKeir Are you still interested in this?

astei:
It seems you do not have any continued interest in doing this, so I'm closing this branch. Feel free to open a new PR if you want to continue contributing!

PR #20, opened 2018-08-06
GS4 Query handler

mikroskeem:
~~Waiting for suggestions while I work on this~~

This PR adds GameSpy 4 Query (which Minecraft servers support) handler to Velocity.


mikroskeem:
Currently stuck behind given things:
- No simple way to get supported game versions as strings (can be done using reflection, but I'd expect to see such thing in API)

Questions:
- Should Velocity expose plugins? (https://wiki.vg/Query#K.2C_V_section)
- [What should this contain?](https://github.com/astei/velocity/pull/20/files#diff-681276c1ad896b3d2f525cb545ab784eR109)
- Should [config](https://github.com/astei/velocity/pull/20/files#diff-222ba29d705ec9dd97deeab2a4f449e4R11) only expose query port or keep it as-is?

astei:
> No simple way to get supported game versions as strings (can be done using reflection, but I'd expect to see such thing in API)

No need to do that. Just create a new constant in `ProtocolConstants` and hoist the current 1.9-1.13 there.

> What should this contain?

I think just "Velocity" is fine, but feel free to add an option to configure the map name (see my next answer below).

> Should config only expose query port or keep it as-is?

I think keeping the config the way it is fine. Perhaps you could move it into a new table called `query` and you can put `enabled = false`, `map_name = "Velocity"` and `port = 25577`, i.e.

```
[query]
enabled = false
port = 25577
map_name = "Velocity"
```

proxy/src/main/java/com/velocitypowered/proxy/protocol/netty/GS4QueryHandler.java
```diff
@@ -0,0 +1,189 @@
...
+        int sessionId = queryMessage.readInt();
+
+        // Allocate buffer for response
+        ByteBuf queryResponse = ctx.alloc().buffer();
+        DatagramPacket responsePacket = new DatagramPacket(queryResponse, msg.sender());
+
+        switch(type) {
+            case QUERY_TYPE_HANDSHAKE: {
+                // Generate new challenge token and put it into the sessions cache
+                int challengeToken = random.nextInt();
```
astei: Preferred practice in Java 8 (and Velocity) is to use `ThreadLocalRandom`.

mikroskeem:
How about making `plugins` configurable and/or modifyable by an event (ProxyQueryEvent)?

mikroskeem:
But alright, I'll continue with this tomorrow.

astei:
@mikroskeem That might work... if we had a plugin and event system. I want to think about these carefully.

mikroskeem:
I think it is ready now. I'll squash commits before merging.

proxy/src/main/java/com/velocitypowered/proxy/protocol/netty/GS4QueryHandler.java
```diff
@@ -0,0 +1,189 @@
...
+        ctx.writeAndFlush(responsePacket);
+    }
+
+    @Override
+    public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws Exception {
+        logger.warn("Error while trying to handle a query packet from " + ctx.channel().remoteAddress(), cause);
+    }
+
+    private static void writeString(ByteBuf buf, String string) {
+        buf.writeBytes(string.getBytes(StandardCharsets.UTF_8));
```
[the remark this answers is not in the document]
Minecrell: Additionally, as far as I can tell, the query protocol appears to encode the characters as single bytes (so pretty much extended ASCII), not UTF-8, so you should use `StandardCharsets.ISO_8859_1` here.

proxy/src/main/java/com/velocitypowered/network/ConnectionManager.java
```diff
@@ -120,6 +127,24 @@ protected void initChannel(final Channel ch) {
                 });
     }
 
+    public void queryBind(final String hostname, final int port) {
+        final Bootstrap bootstrap = new Bootstrap()
+                .channel(datagramChannelClass)
+                .group(this.workerGroup)
+                .handler(new GS4QueryHandler())
+                .localAddress(hostname, port);
+        bootstrap.bind()
```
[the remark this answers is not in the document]
mikroskeem: It's done in above `bind(InetSocketAddress)` as well, so I followed the style. But I can remove it.

proxy/src/main/java/com/velocitypowered/proxy/protocol/netty/GS4QueryHandler.java
```diff
@@ -0,0 +1,189 @@
...
+
+    private final static short QUERY_MAGIC_FIRST = 0xFE;
+    private final static short QUERY_MAGIC_SECOND = 0xFD;
+    private final static byte QUERY_TYPE_HANDSHAKE = 0x09;
+    private final static byte QUERY_TYPE_STAT = 0x00;
+    private final static byte[] QUERY_RESPONSE_FULL_PADDING = new byte[] { 0x73, 0x70, 0x6C, 0x69, 0x74, 0x6E, 0x75, 0x6D, 0x00, (byte) 0x80, 0x00 };
+    private final static byte[] QUERY_RESPONSE_FULL_PADDING2 = new byte[] { 0x01, 0x70, 0x6C, 0x61, 0x79, 0x65, 0x72, 0x5F, 0x00, 0x00 };
+
+    // Contents to add into basic stat response. See ResponseWriter class below
+    private final static List<String> QUERY_BASIC_RESPONSE_CONTENTS = Arrays.asList(
```
astei: Suggest using `ImmutableList#of()` here.

mikroskeem: Tbh ImmutableSet would be even better choice here.

astei: That would work fine, too.

PR #22, opened 2018-08-06
Add .idea to .gitignore file

MatrixTunnel:


astei:
`.idea` should already be excluded considering the rest of the `.gitignore` names specific files to exclude.

MatrixTunnel:
It technically is but some files I have locally aren't being excluded as you can see here https://i.imgur.com/vPdupHC.png

Might just be something on my end ¯\\\_(ツ)\_/¯

electronicboy:
the .idea folder has stuff which is actually useful for sharing, it's more the concept of user-specific/"high churn" files which you don't want in the repo, configs for stuff like code styling are useful to others and are fin to be in the repo

MatrixTunnel:
Ah, I forgot about that. Thanks :)

PR #25, opened 2018-08-07
[WIP] Plugin API

hugmanrique:
Introduces an annotation based plugin API.

A plugin must set its main class on the manifest file (this can be discussed) and that main class needs to have the `@Plugin` annotation which includes metadata and dependency information.

This API was heavily inspired by Voxelwind and Sponge's plugin APIs. Thank you! ❤️ 

TODO:

- [x] Load plugins
- [x] Validate plugin IDs
- [ ] Inject logger and server getter methods to plugin main class (introduce Guice?)
- [ ] Add tests
- [x] Remove the stupid `@author` header (I'm sorry 😢)


api/src/main/java/com/velocitypowered/api/plugin/PluginDescription.java
```diff
@@ -0,0 +1,75 @@
...
+import java.nio.file.Path;
+import java.util.Collection;
+import java.util.Optional;
+import java.util.Set;
+import java.util.regex.Pattern;
+
+/**
+ * Represents metadata for a specific version of a plugin.
+ */
+public interface PluginDescription {
```
jamierocks: `PluginDescriptor` would be a better name for this.

proxy/src/main/java/com/velocitypowered/proxy/plugin/loader/JavaPluginLoader.java
```diff
@@ -0,0 +1,123 @@
...
+    private String getMainClassName(Path source) throws Exception {
+        try (JarInputStream in = new JarInputStream(new BufferedInputStream(Files.newInputStream(source)))) {
+            Manifest manifest = in.getManifest();
+
+            if (manifest == null) {
+                throw new InvalidPluginException("Jar does not contain a manifest");
+            }
+
+            Attributes attributes = manifest.getMainAttributes();
+            String className = attributes.getValue("Main-Class");
```
jamierocks: This isn't a good idea. We shouldn't be using any existing Java entries.

hugmanrique: Yeah, we've been discussing on Discord about removing the option to have multiple plugins on one Jar file.

I agree that the `Main-Class` attribute was an ugly idea, will change it to a `.toml` file.

api/src/main/java/com/velocitypowered/api/plugin/PluginContainer.java
```diff
@@ -0,0 +1,26 @@
...
+    default Optional<?> getInstance() {
+        return Optional.empty();
+    }
+
+    /**
+     * Returns the assigned logger to this {@link Plugin}.
+     *
+     * @return the assigned logger
+     */
+    Logger getLogger();
```
Minecrell: This should use a slf4j logger instead. JUL doesn't work well on top of Log4J 2.

proxy/src/main/java/com/velocitypowered/proxy/plugin/util/DirectedAcyclicGraph.java
```diff
@@ -0,0 +1,145 @@
+package com.velocitypowered.proxy.plugin.util;
+
+import com.google.common.collect.ImmutableList;
+
+import java.util.*;
+
+public class DirectedAcyclicGraph<T> {
```
Minecrell: Maybe you can make use of Guava's graph API instead here?

api/src/main/java/com/velocitypowered/api/plugin/PluginDescription.java
astei: This does sound more like an internal implementation detail.

api/src/main/java/com/velocitypowered/api/plugin/Plugin.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface Plugin {
+    /**
+     * The pattern plugin IDs must match. Plugin IDs may only contain
+     * alphanumeric characters, dashes or underscores, must start with
+     * an alphabetic character and cannot be longer than 64 characters.
+     */
+    Pattern ID_PATTERN = PluginDescription.ID_PATTERN;
```
astei: This is too much of an internal implementation detail. We should just document the ID requirement.

proxy/src/main/java/com/velocitypowered/proxy/plugin/util/DirectedAcyclicGraph.java
astei: Admittedly, he probably ripped most of this from Voxelwind. Guava 19 (which Voxelwind used) didn't have a graph API.

api/src/main/java/com/velocitypowered/api/plugin/PluginContainer.java
hugmanrique: I'm gonna need to include Log4J 2 on the API module then (check the TODO on the implementation)

Minecrell: It already uses log4j2, so all that is needed is `compile 'org.slf4j:slf4j-api:1.7.25'` in the API and `compile "org.apache.logging.log4j:log4j-slf4j-impl:${log4jVersion}"` in the proxy.

astei: No, you won't need to do that. Just include the SLF4J API in the API module and then the log4j connector on the proxy.

api/build.gradle
```diff
@@ -8,6 +8,7 @@ dependencies {
     compile "com.google.guava:guava:${guavaVersion}"
     compile 'net.kyori:text:1.12-1.6.0-SNAPSHOT'
     compile 'com.moandjiezana.toml:toml4j:0.7.2'
+    compile "org.apache.logging.log4j:log4j-api:${log4jVersion}"
```
jamierocks: this should be slf4j.

.idea/compiler.xml
```diff
@@ -0,0 +1,15 @@
...
+      <module name="velocity-api_test" target="1.8" />
+      <module name="velocity-native_main" target="1.8" />
+      <module name="velocity-native_test" target="1.8" />
+      <module name="velocity-proxy_main" target="1.8" />
+      <module name="velocity-proxy_test" target="1.8" />
+      <module name="velocity_main" target="1.8" />
+      <module name="velocity_test" target="1.8" />
+    </bytecodeTargetLevel>
+  </component>
+</project>
```
jamierocks: noop

proxy/src/main/java/com/velocitypowered/proxy/plugin/VelocityPluginManager.java
```diff
@@ -0,0 +1,112 @@
...
+import java.util.*;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+
+public class VelocityPluginManager implements PluginManager {
+    private static final Logger logger = LogManager.getLogger(VelocityPluginManager.class);
+
+    private final Map<String, PluginContainer> plugins = new HashMap<>();
+    private final Map<Object, PluginContainer> pluginInstances = new HashMap<>();
```
Minecrell: IMO, this should be an `IdentityHashMap`. Otherwise it's possible to mess this up pretty easily by implementing a weird `equals` method in the plugin main class. There should never be more than once instance of the plugin main class, so the identity check should do the job.

proxy/src/main/java/com/velocitypowered/proxy/plugin/util/PluginDependencyUtils.java
```diff
@@ -0,0 +1,67 @@
...
+    public static List<PluginCandidate> sortCandidates(List<PluginCandidate> candidates) {
+        // Create our graph, we're going to be using this for Kahn's algorithm.
+        MutableGraph<PluginCandidate> graph = GraphBuilder.directed().allowsSelfLoops(false).build();
+
+        // Add edges
+        for (PluginCandidate description : candidates) {
+            graph.addNode(description);
+
+            for (PluginDependency dependency : description.getDependencies()) {
+                Optional<PluginCandidate> in = candidates.stream().filter(d -> d.getId().equals(dependency.getId())).findFirst();
```
astei: If a plugin depends on a lot of dependencies and the user has lots of plugins installed, this could slow down quickly. Consider having an ID->candidate map?

theminecoder:
I'm going to make a slightly controversial suggestion here. I would like to see the plugin classes implement an a standard interface (I understand not wanting to extend a class) so that we aren't just throwing objects left and right with no idea they if they are the actual plugin instances or not.

This will help a lot with plugin inter-operation, esp when having plugins that provide apis on a per plugin basis. It's much easier to deal with a `Map<Plugin, OtherObject>` rather then having to use a `Map<Object, OtherObject>` along with another check with the plugin manager.

proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
```diff
@@ -96,9 +98,34 @@ public void start() {
...
+        logger.info("Loading plugins...");
+
+        try {
+            Path pluginPath = Paths.get("plugins");
+
+            if (Files.notExists(pluginPath)) {
+                Files.createDirectory(pluginPath);
+            } else {
+                if (!Files.isDirectory(pluginPath)) {
+                    logger.info("Plugin location {} is not a directory, continuing without loading plugins", pluginPath);
```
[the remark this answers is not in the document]
mikroskeem: This should be logged on a warning level indeed.

jamierocks:
> This will help a lot with plugin inter-operation, esp when having plugins that provide apis on a per plugin basis. 

That would provide 0 assistance on that front, either way you have to do casts.

astei:
I have updated this branch for the latest `master` and will continue the plugin API work from there. I want to thank you for your contribution even though you don't have time to continue working on this.

hugmanrique:
No problem 😛

PR #34, opened 2018-08-12
Plugin API

### LINKED ISSUE #10: Initial support for plugins
There is currently no plugin support in Velocity. Adding support should be fairly trivial.

astei:
**This isn't finished yet!**

Resolves #10 and supersedes #25 (thanks to @hugmanrique for his good work so far)


api/src/main/java/com/velocitypowered/api/event/EventManager.java
```diff
@@ -0,0 +1,62 @@
...
+     * @param event the event to post
+     * @return a {@link CompletableFuture} representing the posted event
+     */
+    CompletableFuture<Object> post(@Nonnull Object event);
+
+    /**
+     * Posts the specified event to the event bus and discards the result.
+     * @param event the event to post
+     */
+    default void fireAndForget(@Nonnull Object event) {
```
kashike: `postAndForget`

api/src/main/java/com/velocitypowered/api/event/EventManager.java
```diff
@@ -0,0 +1,62 @@
...
+        post(event);
+    }
+
+    /**
+     * Unregisters all listeners for the specified {@code plugin}.
+     * @param plugin the plugin to deregister listeners for
+     */
+    void unregisterPluginListeners(@Nonnull Object plugin);
```
kashike: drop `Plugin` from name

api/src/main/java/com/velocitypowered/api/event/EventManager.java
astei: We use `fireAndForget` in `ConnectionRequestBuilder`. I think this is more consistent.

api/src/main/java/com/velocitypowered/api/event/EventManager.java
astei: Good idea

api/src/main/java/com/velocitypowered/api/event/EventManager.java
kashike: You're not consistent with `post`, however - `post` -> `fire`, then, to be consistent

api/src/main/java/com/velocitypowered/api/event/connection/LoginEvent.java
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * This event is fired once the player has been authenticated but before they connect to a server on the proxy.
+ */
+public class LoginEvent implements ResultedEvent<ResultedEvent.ComponentResult> {
+    private final Player player;
+    private ComponentResult result;
+
+    public LoginEvent(@NonNull Player player) {
+        this.player = player;
```
hugmanrique: Missing `checkNotNull` check here.

api/src/main/java/com/velocitypowered/api/event/connection/PreLoginEvent.java
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * This event is fired when a player has initiated a connection with the proxy but before the proxy authenticates the
+ * player with Mojang or before the player's proxy connection is fully established (for offline mode).
+ */
+public class PreLoginEvent implements ResultedEvent<ResultedEvent.ComponentResult> {
+    private final InboundConnection connection;
+    private final String username;
+    private ComponentResult result;
+
+    public PreLoginEvent(InboundConnection connection, String username) {
```
hugmanrique: These should also be `@NonNull` and have the not null check too.

api/src/main/java/com/velocitypowered/api/util/UuidUtils.java
```diff
@@ -1,15 +1,18 @@
...
 
-public enum UuidUtils {
-    ;
+public class UuidUtils {
+    private UuidUtils() {
+        throw new AssertionError();
+    }
 
-    public static UUID fromUndashed(final String string) {
+    public static @NonNull UUID fromUndashed(final @NonNull String string) {
```
hugmanrique: When we had that discussion on Discord we talked that plugins should always assume the API returns non-null objects, and if nullability is unavoidable we should use an `Optional` instead.

api/src/main/java/com/velocitypowered/api/event/connection/PreLoginEvent.java
astei: I am waiting on adding in full checker framework support.

proxy/src/main/java/com/velocitypowered/proxy/command/VelocityCommandManager.java
```diff
@@ -0,0 +1,84 @@
...
+
+        try {
+            command.execute(source, actualArgs);
+            return true;
+        } catch (Exception e) {
+            throw new RuntimeException("Unable to invoke command " + cmdLine + " for " + source, e);
+        }
+    }
+
+    public Optional<List<String>> offerSuggestions(CommandSource source, String cmdLine) {
```
XakepSDK: Bad practice, return empty list instead of optional

proxy/src/main/java/com/velocitypowered/proxy/plugin/loader/JavaPluginLoader.java
```diff
@@ -0,0 +1,129 @@
...
+        }
+
+        JavaVelocityPluginDescription javaDescription = (JavaVelocityPluginDescription) description;
+        Optional<Path> source = javaDescription.getSource();
+
+        if (!source.isPresent()) {
+            throw new IllegalArgumentException("No path in plugin description");
+        }
+
+        Injector injector = Guice.createInjector(new VelocityPluginModule(server, javaDescription, baseDirectory));
```
XakepSDK: Allow plugin to implement their module for in-plugin injections.

astei: I don't know how I think about this. This is an area of active consideration.

XakepSDK: Maybe special field in `Plugin` annotation that holds list of modules.(`Class<Module>[]`). Every module instance created using some injector, then plugin injector created with these modules.

PR #39, opened 2018-08-15
Fix PacketMapping toString/equals/hashCode, do not handle Chat and Ta…

Leymooo:
…bCompleteResponse packets from backend

This PR will help to reduce overhead by decoding and encoding packets, since Velocity does not handle a Chat and TabCompleteResponse from backend server.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/protocol/StateRegistry.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/protocol/StateRegistry.java
@@ -13,6 +13,7 @@ import java.util.function.Supplier;
 import static com.velocitypowered.proxy.protocol.ProtocolConstants.*;
 
 public enum StateRegistry {
+        
     HANDSHAKE {
         {
             SERVERBOUND.register(Handshake.class, Handshake::new,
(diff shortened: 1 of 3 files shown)


MylesIsCool:
Would it be better to call them something else than map1 and 0? Just seems a bit confusing.

Leymooo:
i have no ideas how to name those methods (

proxy/src/main/java/com/velocitypowered/proxy/protocol/StateRegistry.java
```diff
@@ -282,29 +283,46 @@ public boolean equals(Object o) {
...
     }
     
-    private static PacketMapping map(int id, int version) {
+    /**
+     * Create a PacketMapping where packet decoding is <b>enabled</b>
+     * @param id Packet id
+     * @param version Protocol version
+     * @return PacketMapping with packet decoding enabled
+     */
+    private static PacketMapping map1(int id, int version) {
```
hugmanrique: As @MylesIsCool said, most packets Velocity handles are not decoded by default, so it would make sense to have a `map(int id, int version)` map that called `map(int id, int version, false)` (where `false` is `packetDecoding`).

proxy/src/main/java/com/velocitypowered/proxy/protocol/StateRegistry.java
```diff
@@ -260,19 +261,20 @@ public String toString() {
     public static class PacketMapping {
         private final int id;
         private final int protocolVersion;
-        private final boolean needPacketDecode;
+        private final boolean packetDecoding;
```
astei: Would `encodeOnly` be a better name for this?

Leymooo:
@astei now it good?

PR #40, opened 2018-08-16
Update kyori/event to 3.0.0, implement listener ordering, implement ignore cancelled

lucko:


api/src/main/java/com/velocitypowered/api/event/EventManager.java
```diff
@@ -19,10 +19,11 @@
      * Requests that the specified {@code handler} listen for events and associate it with the {@code plugin}.
      * @param plugin the plugin to associate with the handler
      * @param eventClass the class for the event handler to register
+     * @param postOrder the order in which events should be posted to the handler
      * @param handler the handler to register
      * @param <E> the event type to handle
      */
-    <E> void register(@NonNull Object plugin, @NonNull Class<E> eventClass, @NonNull EventHandler<E> handler);
+    <E> void register(@NonNull Object plugin, @NonNull Class<E> eventClass, @NonNull PostOrder postOrder, @NonNull EventHandler<E> handler);
```
astei: Would like an overload without the `PostOrder` that simply specifies a normal post order.

proxy/src/main/java/com/velocitypowered/proxy/plugin/VelocityEventManager.java
```diff
@@ -129,28 +142,35 @@ public boolean shouldRegister(@NonNull Object listener, @NonNull Method method)
...
-        public @NonNull PostOrder postOrder(@NonNull Object listener, @NonNull Method method) {
-            return PostOrder.NORMAL; // TODO: Allow customizing his
+        public int postOrder(@NonNull Object listener, @NonNull Method method) {
+            return method.getAnnotation(Subscribe.class).order().ordinal();
         }
 
         @Override
         public boolean consumeCancelledEvents(@NonNull Object listener, @NonNull Method method) {
-            return true;
+            return !method.getAnnotation(Subscribe.class).ignoreCancelled();
```
astei: Velocity's "cancelled event" is based on `ResultedEvent<T>`. As this presently stands, all "cancelled events" will still be called by the event system.

api/src/main/java/com/velocitypowered/api/event/Subscribe.java
```diff
@@ -11,4 +11,19 @@
...
+     * @return the order
+     */
+    PostOrder order() default PostOrder.NORMAL;
+
+    /**
+     * If cancelled events shouldn't be posted to this listener.
+     *
+     * @return if cancelled events should be ignored.
+     */
+    boolean ignoreCancelled() default false;
```
astei: Velocity's does not have a concept of a "cancelled event". It does, however, have the closely related "resulted event".

proxy/src/main/java/com/velocitypowered/proxy/plugin/VelocityEventManager.java
```diff
@@ -51,11 +52,11 @@ public void register(@NonNull Object plugin, @NonNull Object listener) {
     }
 
     @Override
-    public <E> void register(@NonNull Object plugin, @NonNull Class<E> eventClass, @NonNull EventHandler<E> handler) {
+    public <E> void register(@NonNull Object plugin, @NonNull Class<E> eventClass, @NonNull PostOrder postOrder, @NonNull EventHandler<E> handler) {
         Preconditions.checkNotNull(plugin, "plugin");
         Preconditions.checkNotNull(eventClass, "eventClass");
         Preconditions.checkNotNull(handler, "listener");
-        bus.register(eventClass, new KyoriToVelocityHandler<>(handler));
```
astei: Plugins may not respect checker-qual, so make sure to add in an explicit null check for the `postOrder`.

proxy/src/main/java/com/velocitypowered/proxy/plugin/VelocityEventManager.java
lucko: Ahh, my bad :)

PR #41, opened 2018-08-16
Permissions!

lucko:
My aims with this were, in no particular order:

* Easy/simple to use
* Straight forward to implement, delegating *all* handling to the plugin managing permissions
* Fast in terms of performance

I've borrowed a few concepts from other areas of the MC ecosystem, namely

* `Tristate` from [Sponge](https://github.com/SpongePowered/SpongeAPI/blob/stable-7/src/main/java/org/spongepowered/api/util/Tristate.java) & [LuckPerms](https://github.com/lucko/LuckPerms/blob/master/api/src/main/java/me/lucko/luckperms/api/Tristate.java).

* `PermissionSubject` from [Sponge](https://github.com/SpongePowered/SpongeAPI/blob/stable-7/src/main/java/org/spongepowered/api/service/permission/Subject.java), although only the name is copied. The interface is really just a simple version of Bukkit's `Permissible`.

* `PermissionFunction` fulfils the same purpose as Bukkit's `PermissibleBase`, except it's not an implementation. It allows other plugins to effectively implement a `PermissionSubject`'s hasPermission method.

Design choices:

* Why not just an event? Well, mainly performance reasons. Permission checks should be fast, and creating, scheduling and then joining a future isn't exactly fast, especially in comparison with just doing a direct call to the provider. Additionally, doing things this way allows providers to store a players permissions data directly on the provider instance, as opposed to doing map lookups.

Potential issues:

* If support for reloading individual plugins was ever implemented, we'd probably have to re-think the way plugin provided "services" are handled.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
@@ -2,6 +2,8 @@ package com.velocitypowered.proxy.connection.client;
 
 import com.google.common.base.Preconditions;
 import com.google.gson.JsonObject;
+import com.velocitypowered.api.permission.PermissionFunction;
+import com.velocitypowered.api.permission.PermissionProvider;
 import com.velocitypowered.api.proxy.ConnectionRequestBuilder;
 import com.velocitypowered.api.util.MessagePosition;
 import com.velocitypowered.api.proxy.Player;
@@ -35,12 +37,14 @@ import java.util.concurrent.CompletableFuture;
 
 public class ConnectedPlayer implements MinecraftConnectionAssociation, Player {
     private static final PlainComponentSerializer PASS_THRU_TRANSLATE = new PlainComponentSerializer((c) -> "", TranslatableComponent::key);
+    public static final PermissionProvider DEFAULT_PERMISSIONS = s -> PermissionFunction.ALWAYS_UNDEFINED;
 
     private static final Logger logger = LogManager.getLogger(ConnectedPlayer.class);
 
     private final GameProfile profile;
     private final MinecraftConnection connection;
     private final InetSocketAddress virtualHost;
+    private PermissionFunction permissionFunction = null;
     private int tryIndex = 0;
     private ServerConnection connectedServer;
     private ClientSettings clientSettings;
@@ -83,6 +87,10 @@ public class ConnectedPlayer implements MinecraftConnectionAssociation, Player {
     @Override
     public Optional<InetSocketAddress> getVirtualHost() {
         return Optional.ofNullable(virtualHost);
+    }
+
+    public void setPermissionFunction(PermissionFunction permissionFunction) {
+        this.permissionFunction = permissionFunction;
     }
 
     @Override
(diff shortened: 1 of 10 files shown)


proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
```diff
@@ -50,6 +58,10 @@ public ConnectedPlayer(GameProfile profile, MinecraftConnection connection, Inet
         this.profile = profile;
         this.connection = connection;
         this.virtualHost = virtualHost;
+
+        // load permissions provider
+        EventManager eventManager = VelocityServer.getServer().getEventManager();
+        this.permissionFunction = eventManager.fire(new PermissionsSetupEvent(this, DEFAULT_PERMISSIONS)).join().createFunction(this);
```
astei: I prefer if this did not use `.join()`, this blocks the Netty I/O threads and defeats the point of moving event listeners off the main thread (which was specifically intended to avoid executing or waiting on "alien" code).

astei: I don't doubt that LuckPerms would sanely handle this event. But I'm not so sure about other plugins, which might do things like run a SQL query when this event is fired, which was a major consideration in forcing all events to be run "asynchronously" in Velocity. It's still early days, so we could have a discussion to revisit this decision (perhaps there are valid performance considerations).

The way I see it: Velocity should do its very best to ensure the Netty I/O threads are runnable at all times doing useful work like processing packets. Using `join()` here defeats the point: it's blocking waiting on alien code to finish up.

lucko: I understand the aim, but as I said, I don't see a sane way to achieve it.

1) Call the event and store a `CompletableFuture<PermissionsSetupEvent>` instance on the `ConnectedPlayer`. I dislike this because it's messy, and you're really just moving the problem elsewhere - we'll have to block when hasPermission is eventually called.

2) Init the function lazily when the future completes - causes race conditions, not really an acceptable solution imo.

astei: Or simply initialize an `UNDEFINED`-only permission function in `ConnectedPlayer`, fire the event once `ConnectedPlayer` is constructed, wait for it to complete asynchronously, set the new permission function, and allow the normal login process to complete.

This should be ok: plugins handling permission setup should not rely on the player having permissions already.

lucko: And if the normal login process completes before permissions are loaded?

As you said: 
> But I'm not so sure about other plugins, which might do things like run a SQL query when this event is fired

Which:
> causes race conditions, not really an acceptable solution imo

astei: > And if the normal login process completes before permissions are loaded?

I did not mean "the normal login process completes concurrently with the event execution". Simply use `whenCompleteAsync` to get the event execution back on the Netty event loop and finish the login process there.

lucko: Okay, I've moved the loading into the `LoginSessionHandler`, so everything is loading off the netty threads now. :)

astei:
LGTM.

PR #45, opened 2018-08-21
Fallback to next server on connection error

### LINKED ISSUE #11: Implement fallback connection support

phase:
Taken from #14. I just needed a reason to clone it and check out the codebase :)
Tell me if something else is wanted with this feature.

Closes #11


astei:
Exactly what we needed, thanks!

PR #48, opened 2018-08-22
possibility to force port via start-argument

BukkitTNT:


mikroskeem:
No... just no.

Better introduce proper argument parser (jopt-simple for example) and then provide a way to force the port.

mikroskeem:
Or that... 😄

PR #51, opened 2018-08-24
Allow to enable online mode for player connection

Leymooo:
This will be useful for auth plugins on offline mode servers. Eg to allow player to not use '/login' command on join if he login using a licensed account

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/event/ResultedEvent.java
+++ b/api/src/main/java/com/velocitypowered/api/event/ResultedEvent.java
@@ -1,6 +1,7 @@
 package com.velocitypowered.api.event;
 
 import com.google.common.base.Preconditions;
+
 import net.kyori.text.Component;
 import net.kyori.text.serializer.ComponentSerializers;
 import org.checkerframework.checker.nullness.qual.NonNull;
@@ -104,6 +105,57 @@ public interface ResultedEvent<R extends ResultedEvent.Result> {
         public static ComponentResult denied(@NonNull Component reason) {
             Preconditions.checkNotNull(reason, "reason");
             return new ComponentResult(false, reason);
+        }
+    }
+    /**
+     * Represents an "allowed/allowed with online mode/denied" result with a reason allowed for denial.
+     */
+    class PreLoginComponentResult extends ComponentResult {
+
+        private static final PreLoginComponentResult ALLOWED = new PreLoginComponentResult(true, null);
+        private static final PreLoginComponentResult ALLOWED_ONLINEMODE = new PreLoginComponentResult(true);
+
+        private final boolean onlineMode;
+        
+        /**
+         * Allows to enable a online mode for the player connection, when velocity running in offline mode
+         * Does not have any sense if velocity running in onlineMode;
+         * @param onlineMode if true, offline uuid will be used for player connection to avoid collision
+         */
+        private PreLoginComponentResult(boolean onlineMode) {
+            super(true, null);
+            this.onlineMode = onlineMode;
+        }
+
+        private PreLoginComponentResult(boolean allowed, @Nullable Component reason) {
+            super(allowed, reason);
+            // Don't care about this
+            this.onlineMode = false;
+        }
+
+        public boolean isOnlineMode() {
+            return this.onlineMode;
+        }
+
+        @Override
+        public String toString() {
+            if (isOnlineMode()) {
+                return "allowed with online mode and offline uuid";
+            }
+            return super.toString();
+        }
+
+        public static PreLoginComponentResult allowed() {
+            return ALLOWED;
+        }
+
+        public static PreLoginComponentResult allowedOnlineMode() {
+            return ALLOWED_ONLINEMODE;
+        }
+        
+        public static PreLoginComponentResult denied(@NonNull Component reason) {
+            Preconditions.checkNotNull(reason, "reason");
+            return new PreLoginComponentResult(false, reason);
         }
     }
 }
--- a/api/src/main/java/com/velocitypowered/api/event/ResultedEvent.java
+++ b/api/src/main/java/com/velocitypowered/api/event/ResultedEvent.java
@@ -73,7 +73,7 @@ public interface ResultedEvent<R extends ResultedEvent.Result> {
         private final boolean allowed;
         private final @Nullable Component reason;
 
-        private ComponentResult(boolean allowed, @Nullable Component reason) {
+        protected ComponentResult(boolean allowed, @Nullable Component reason) {
             this.allowed = allowed;
             this.reason = reason;
         }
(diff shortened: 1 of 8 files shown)


api/src/main/java/com/velocitypowered/api/event/connection/PreLoginEvent.java
```diff
@@ -28,6 +30,14 @@ public String getUsername() {
         return username;
     }
 
+    public boolean isOnlineMode() {
+        return onlineMode;
+    }
+    
+    public void setOnlineMode(boolean onlineMode) {
```
astei: I oppose this approach since it goes against the very idea of resulted events. I would prefer this to be part of the result: that is, please use a new result class that represents "allowed", "allowed with fake game profile", "denied with reason".

api/src/main/java/com/velocitypowered/api/proxy/Player.java
```diff
@@ -27,6 +27,13 @@
      */
     UUID getUniqueId();
 
+    /**
+     * Sets the player's UUID.
+     * Can be set only in Login state
+     * @param uuid
+     */
+    void setUniqueId(UUID uuid);
```
astei: During the login process, the `Player` object is not available. It only becomes available _after_ the login process has completed (in other words, when `LoginEvent` is fired).

api/src/main/java/com/velocitypowered/api/util/GameProfile.java
```diff
@@ -26,6 +26,10 @@ public UUID idAsUuid() {
         return UuidUtils.fromUndashed(id);
     }
 
+    public void setUuid(UUID uuid) {
```
astei: This object should remain immutable.

Leymooo:
Now its ok?

api/src/main/java/com/velocitypowered/api/event/ResultedEvent.java
```diff
@@ -106,4 +107,55 @@ public static ComponentResult denied(@NonNull Component reason) {
             return new ComponentResult(false, reason);
         }
     }
+    /**
+     * Represents an "allowed/allowed with online mode/denied" result with a reason allowed for denial.
+     */
+    class PreLoginComponentResult extends ComponentResult {
```
astei: Do not put this in the `ResultedEvent` class. Convention is that if a result is event-specific, it is a subclass of that event.

api/src/main/java/com/velocitypowered/api/event/ResultedEvent.java
```diff
@@ -106,4 +107,55 @@ public static ComponentResult denied(@NonNull Component reason) {
...
+
+        private static final PreLoginComponentResult ALLOWED = new PreLoginComponentResult(true, null);
+        private static final PreLoginComponentResult ALLOWED_ONLINEMODE = new PreLoginComponentResult(true);
+
+        private final boolean onlineMode;
+        
+        /**
+         * Allows to enable a online mode for the player connection, when velocity running in offline mode
+         * Does not have any sense if velocity running in onlineMode;
+         * @param onlineMode if true, offline uuid will be used for player connection to avoid collision
```
astei: Huh? Note that if Velocity determines that a player with the same username is online, it disconnects the new connection.

astei: On a related note, this should really have a `GameProfile` instead. An empty fake game profile means Velocity should perform regular authentication, otherwise Velocity can proceed to login with the fake profile. (In other words, this means that a basic version of a plugin like SkinsRestorer can work on Velocity simply by checking if the proxy is in offline mode, and if so, modifying the result of the `PreLoginEvent` with the new game profile).

Leymooo: I mean that a returned UUID from mojang, can be same with uuid of other player that was generated for offline player. Chance is very small, but..

astei: Unlikely:

```
>>> import uuid
>>> uuid.UUID('652a2bc4-e8cd-405d-b7b6-98156ee2dc09').version
4
>>> 
```

UUID version 4 IDs are totally random and the chances of two UUID v4 iDs colliding is so minimal that for all intents an purposes, UUIDs are effectively random.

Leymooo: I think it better to use offline uuid, if velocity is running in offline mode.

astei: For sanity, yes, we should do that. At the very least though, plugins should be able to inject arbitrary profile properties.

astei: Here's a usecase for allowing arbitrary UUIDs. If Mojang login servers go down (this used to happen a lot, it's rare now), under certain circumstances a plugin may wish to bypass authentication and fake the player's UUID and their skin.

astei: Why don't we try something else? A new event (`GameProfileRequestEvent`) fired just after the pre-login event and a plugin can set an arbitrary game profile there. It's never invoked in online-mode since it doesn't make sense.

Leymooo:
I am updated a pr. I think we need to create something like a PostLoginEvent or PreLoginSuccessEvent where we can finally pass a GameProfile before LoginSuccess will be sended.

astei:
@Leymooo I already suggested that and a possible name for it (`GameProfileRequestEvent`).

Leymooo:
Done

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
```diff
@@ -45,11 +47,11 @@
 
     private static final Logger logger = LogManager.getLogger(ConnectedPlayer.class);
 
-    private final GameProfile profile;
     private final MinecraftConnection connection;
     private final InetSocketAddress virtualHost;
     private PermissionFunction permissionFunction = null;
     private int tryIndex = 0;
+    private GameProfile profile;
```
astei: This should still remain `final`.

proxy/src/main/java/com/velocitypowered/proxy/connection/client/LoginSessionHandler.java
```diff
@@ -153,28 +159,38 @@ private EncryptionRequest generateRequest() {
...
-                    return VelocityServer.getServer().getEventManager().fire(new LoginEvent(player));
-                })
-                // then complete the connection
-                .thenAcceptAsync(event -> {
-                    if (!event.getResult().isAllowed()) {
-                        // The component is guaranteed to be provided if the connection was denied.
-                        inbound.closeWith(Disconnect.create(event.getResult().getReason().get()));
-                        return;
-                    }
+    private void initializePlayer(GameProfile profile, boolean onlinemode) {
```
astei: `onlineMode`

api/src/main/java/com/velocitypowered/api/event/ResultedEvent.java
```diff
@@ -72,7 +73,7 @@ public static GenericResult denied() {
         private final boolean allowed;
         private final @Nullable Component reason;
 
-        private ComponentResult(boolean allowed, @Nullable Component reason) {
+        protected ComponentResult(boolean allowed, @Nullable Component reason) {
```
astei: This should be made package-private.

api/src/main/java/com/velocitypowered/api/event/connection/PreLoginEvent.java
```diff
@@ -46,4 +52,57 @@ public String toString() {
...
+
+        private static final PreLoginComponentResult ALLOWED = new PreLoginComponentResult((Component) null);
+        private static final PreLoginComponentResult ALLOWED_ONLINEMODE = new PreLoginComponentResult(true);
+
+        private final boolean onlineMode;
+        /**
+         * Allows to enable a online mode for the player connection, when Velocity running in offline mode
+         * Does not have any sense if velocity running in onlineMode;
+         * @param onlineMode if true, offline uuid will be used for player connection if Velocity run in offlineMode
+         */
```
astei: Move this Javadoc to `#allowedOnlineMode()` (but see below for a further clarification to make).

api/src/main/java/com/velocitypowered/api/event/connection/PreLoginEvent.java
```diff
@@ -46,4 +52,57 @@ public String toString() {
...
+            }
+            
+            return super.toString();
+        }
+
+        public static PreLoginComponentResult allowed() {
+            return ALLOWED;
+        }
+
+        public static PreLoginComponentResult allowedOnlineMode() {
```
astei: I would call this `forceOnlineMode()` and have an equivalent `forceOfflineMode()` function, too.

api/src/main/java/com/velocitypowered/api/util/GameProfile.java
```diff
@@ -40,6 +40,10 @@ public static GameProfile forOfflinePlayer(@NonNull String username) {
         return new GameProfile(id, username, ImmutableList.of());
     }
 
+    public static String createOfflineUUID(@NonNull String username) {
```
astei: Unused function, please remove it.

api/src/main/java/com/velocitypowered/api/event/connection/PreLoginEvent.java
Leymooo: I really can not understand why we need to have forceOfflineMode when proxy running in online mode.

astei: Okay, then we can omit it. If someone does need it we can add it.

api/src/main/java/com/velocitypowered/api/event/ResultedEvent.java
Leymooo: how i can make this? i can not call super() in PreLoginEvent.PreLoginComponentResult

astei: Oof... okay, then keep this protected. I don't know instinctively all the Java access levels off the top of my head!

astei:
Thanks! I'm mostly happy with this now.

PR #54, opened 2018-08-25
Allow spaces in server names

MatrixTunnel:


mikroskeem:
Wouldn't it interefere with tab completion support?

MatrixTunnel:
I didn't even know there was tab completion. I play on 1.13 so that's probably why it never worked for me

--

After testing, I am unable to get it working at all in 1.12 or below. It should work fine but it won't work for me :/

mikroskeem:
There's a tabcomplete handler below tho? 😄

kashike:
:-1: - server names shouldn't contain spaces.

astei:
I'm afraid to say that this is a feature we won't be supporting.

astei:
@MatrixTunnel On a related note, I considered adding in 1.13 tab complete support but decided against it because of the massive amount of work required to support even rudimentary Brigadier node parsing and writing. If you need any "rich" tab complete you are better off writing your commands as a Bukkit/Sponge plugin instead.

PR #74, opened 2018-09-11
Allow to edit GameProfile properties.

Leymooo:
Can be useful for skin change plugins


Cybermaxke:
The player will have to be respawned (includes sending tab list packets, etc.) on the client in order to update its skin.

Leymooo:
i know this. I make this to avoid using a reflection in plugins.

DoNotSpamPls:
Usually skin plugins circumvent this by vanishing and unvanishing the player instantly.

theminecoder:
The problem with this sort of functionality is unless you're only running one proxy, you need to replicate this change to any other proxy instance that may or may not use the profile at some point. This sort of stuff really should be left to the server level so the proxy doesn't need to special case data for some specific players.

astei:
I think @Cybermaxke is right, and his reasoning is the same reason why I'm leaning towards not including this in Velocity. Either way, you still need to handle the skin switching on the server-side, and if you've already got code to handle that, what is Velocity doing other than providing a very small convenience?

Cybermaxke:
Maybe can the field be internally updated by tracking packets send from the server? I don't see why you would manually modify it.

astei:
That's just extra complexity for what is ultimately a niche feature in my opinion. Velocity can not reasonably cater to 100% of users by being everything to everyone. That sets us up for disaster and will almost certainly lead to Velocity 2.0.0 being a thing within months.

Cybermaxke:
That's true, @Leymooo what is the reason that this should be updated in the proxy, instead of it being kept to server level?

Leymooo:
Check the  ChangeSkin plugin. It uses a reflection to update a field with properties. I thought it could be good to allow change the properties via api.

@Cybermaxke plugins need to update properties on proxy, in order to send a changed skin data to others servers on switch and spawn player using new properties. If plugin dont change the propertis, plugin need to send a new skin data via PM or smth and respawn player again after player was spawned on server after switch.

astei:
I'm leaning towards merging this now. Please rebase on upstream.

astei:
Thanks!

PR #76, opened 2018-09-11
ServerPreConnectEvent#getInfo -> getServer

kashike:


astei:
I'm not seeing any clear advantage from moving to `Objects.requireNonNull()` other than the fact that the latter is a JDK function. For compatibility (and convenience), I find it unlikely that we're going to drop Google Guava from the API.

astei:
On the other hand, the ServerPreConnectEvent#getInfo -> getServer rename is fine although it's breaking.

PR #84, opened 2018-09-13
fix compile

Leymooo:
Sorry for that. Ide shows that all is good.


MiniDigger:
maybe jenkins should be configured to run CI on PRs too? 
could avoid merging PRs that don't compile or don't pass tests.

supertassu:
or travis?

MiniDigger:
no need for travis when there is already a jenkins pipeline that is configured to build and deploy master and only build branches. should be trivial to expend that to PRs.

PR #85, opened 2018-09-14
Add CONNECT_TIMEOUT_MILLIS to fix timeout errors.

PurpleIsEverything:
For some reason having a read timeout handler is not enough. Without this option players will randomly timeout from backend servers.


astei:
Are we sure we want a 30-second connection timeout? I'd prefer something shorter, like the 5 seconds BungeeCord uses.

PurpleIsEverything:
True, though for some modded clients this option will need to be configurable.

astei:
I've merged this in. We can make the connection timeout configurable in the future (as an advanced option).

PR #89, opened 2018-09-16
[WIP] Introduce RegisteredServer API

astei:
This interface roughly maps the BungeeCord ServerInfo API. Accordingly, this is a breaking API change, as many of the server-related events and methods working with server info instances now provide/expect the RegisteredServer interface instead.

This resolves a minor, but potentially serious pain point for potential Velocity API users.


api/src/main/java/com/velocitypowered/api/proxy/ProxyServer.java
```diff
@@ -45,23 +46,24 @@
     int getPlayerCount();
 
     /**
-     * Retrieves a registered {@link ServerInfo} instance by its name. The search is case-insensitive.
+     * Retrieves a registered {@link RegisteredServer} instance by its name. The search is case-insensitive.
      * @param name the name of the server
      * @return the registered server, which may be empty
      */
-    Optional<ServerInfo> getServerInfo(String name);
+    Optional<RegisteredServer> getServerInfo(String name);
```
Cybermaxke: `getServerInfo` -> `getServer`/`getRegisteredServer`

astei: We don't need `getServer()` returning `ServerInfo`, considering you can do `proxy.getServer("test").map(RegisteredServer::getServerInfo)`.

I will rename `getServerInfo` to `getServer` though.

PR #94, opened 2018-09-19
Modinfo: Add getType and getModList methods.

PurpleIsEverything:
Once forced hosts are implemented this will allow caching of mod lists and the ability return them during proxy ping events.

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/proxy/server/ServerPing.java
+++ b/api/src/main/java/com/velocitypowered/api/proxy/server/ServerPing.java
@@ -317,7 +317,7 @@ public class ServerPing {
             return type;
         }
 
-        public List<Mod> getModList() {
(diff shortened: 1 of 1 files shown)


api/src/main/java/com/velocitypowered/api/proxy/server/ServerPing.java
```diff
@@ -312,6 +312,14 @@ public Modinfo(String type, List<Mod> modList) {
             this.type = Preconditions.checkNotNull(type, "type");
             this.modList = ImmutableList.copyOf(modList);
         }
+
+        public String getType() {
+            return type;
+        }
+
+        public List<Mod> getModList() {
```
kashike: `getMods` imo

astei:
I must've forgotten to add this in... Thanks!

PR #95, opened 2018-09-20
Title API

astei:


api/src/main/java/com/velocitypowered/api/util/title/Title.java
```diff
@@ -0,0 +1,51 @@
...
+    };
+
+    /**
+     * A title that, when sent to the client, will cause any existing title to be hidden. The title may be restored by
+     * a {@link TextTitle} with no title or subtitle (only a time).
+     */
+    Title HIDE = new Title() {
+        @Override
+        public String toString() {
+            return "hide title";
```
phase: Is there a reason to have these other than debugging?

astei: You would be correct: they're only there for debugging and consistency (since `TextTitle` has `toString()` as well).

astei:
Verified that this change works with 1.8-1.13.1.

PR #103, opened 2018-09-26
Allow to add custom packet handlers.

Leymooo:
This PR add a way to control packet sending and recieving.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/ConnectionInitEvent.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/ConnectionInitEvent.java
@@ -1,0 +1,38 @@
+package com.velocitypowered.proxy.connection;
+
+import com.velocitypowered.api.proxy.server.RegisteredServer;
+
+/**
+ * This event is used to allow add custom packet handlers to
+ * MinecraftConnection. Use this event only if you know what you do. This event
+ * can fired sync. Use this event only for adding custom handlers
+ */
+@Deprecated
+public class ConnectionInitEvent {
+
+    private final MinecraftConnection minecraftConnection;
+    private final RegisteredServer server;
+
+    public ConnectionInitEvent(MinecraftConnection connection, RegisteredServer server) {
+        this.minecraftConnection = connection;
+        this.server = server;
+    }
+
+    public MinecraftConnection getMinecraftConnection() {
+        return minecraftConnection;
+    }
+
+    public RegisteredServer getServer() {
+        return server;
+    }
+
+    /**
+     * @return true if connection is a server connection(Between proxy and
+     * server), false if connection is a client connection(Between client and
+     * proxy)
+     */
+    public boolean isBackendConnection() {
+        return getServer() != null;
+    }
+
+}
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/MinecraftConnection.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/MinecraftConnection.java
@@ -27,6 +27,7 @@ import java.security.GeneralSecurityException;
 
 import static com.velocitypowered.proxy.network.Connections.*;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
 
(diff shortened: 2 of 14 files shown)


Leymooo:
I dont test this pr. I open it for reviews.

proxy/src/main/java/com/velocitypowered/proxy/connection/ConnectionInitEvent.java
```diff
@@ -0,0 +1,38 @@
...
+    public RegisteredServer getServer() {
+        return server;
+    }
+
+    /**
+     * @return true if connection is a server connection(Between proxy and
+     * server), false if connection is a client connection(Between client and
+     * proxy)
+     */
+    public boolean isBackendConnection() {
```
hugmanrique: I would create two different subclasses for each kind of connection, most people will assume `getServer()` will never return null + it's cleaner imo.

proxy/src/main/java/com/velocitypowered/proxy/connection/MinecraftConnection.java
```diff
@@ -99,20 +151,76 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws E
 
     @Override
     public void channelWritabilityChanged(ChannelHandlerContext ctx) throws Exception {
+        if (pluginsSessionHandlers != null && !pluginsSessionHandlers.isEmpty()) {
+            synchronized (pluginsSessionHandlers) {
```
hugmanrique: You also need to synchronize the if statement above.

proxy/src/main/java/com/velocitypowered/proxy/connection/MinecraftConnection.java
```diff
@@ -85,6 +129,14 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
     @Override
     public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws Exception {
         if (ctx.channel().isActive()) {
+            if (pluginsSessionHandlers != null && !pluginsSessionHandlers.isEmpty()) {
```
hugmanrique: You need to synchronize this if condition to guarantee memory visibility. I assume methods such as `.write` could get called from different threads.

Leymooo: https://github.com/VelocityPowered/Velocity/pull/103/files#diff-98293b855c687bdaddb2c9b78e541088R168
there is a synchronizedList

Leymooo:
there is a example of usage a custom packet handler. https://github.com/Leymooo/VelocityPacketHandler

proxy/src/main/java/com/velocitypowered/proxy/connection/MinecraftConnection.java
```diff
@@ -49,6 +57,13 @@ public MinecraftConnection(Channel channel, VelocityServer server) {
 
     @Override
     public void channelActive(ChannelHandlerContext ctx) throws Exception {
+        if (pluginsSessionHandlers != null && !pluginsSessionHandlers.isEmpty()) {
```
astei: I see lines like this and my only reaction is _yuck!_ - this isn't particularly idiomatic.

I'd prefer something more like `firePluginSessionHandlers(handler::connected);`.

proxy/src/main/java/com/velocitypowered/proxy/connection/MinecraftConnection.java
```diff
@@ -99,20 +153,76 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws E
...
+            }
+        }
         if (sessionHandler != null) {
             sessionHandler.writabilityChanged();
         }
     }
 
+    public void addCustomSessionHandler(MinecraftSessionHandler handler) {
+        if (this.pluginsSessionHandlers == null) {
+            this.pluginsSessionHandlers = Collections.synchronizedList(new ArrayList<>());
```
astei: I would probably prefer using `CopyOnWriteArrayList` here. It is perfect since plugins are unlikely to register and unregister listeners frequently, but iterations occur often.

Mystiflow:
How come this was closed?

PR #77, opened 2018-09-12
[WIP] Add getMotd & getMaxPlayerCount to the ProxyServer API

phase:
I needed these for a plugin. Let me know if anything else is needed!

PR #106, opened 2018-10-01
Added ProxyConfig interface and implemented in VelocityConfiguration

Crypnotic:
This adds the preliminary code to have access to the proxy configuration values through the API. Picked up from #77 after inactivity for some time. No javadocs as of yet

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/proxy/config/ProxyConfig.java
+++ b/api/src/main/java/com/velocitypowered/api/proxy/config/ProxyConfig.java
@@ -1,0 +1,103 @@
+package com.velocitypowered.api.proxy.config;
+
+import com.velocitypowered.api.util.Favicon;
+import net.kyori.text.Component;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * Provides an interface to a proxy configuration
+ */
+public interface ProxyConfig {
+    /**
+     * Whether GameSpy 4 queries are accepted by the proxy
+     * @return queries enabled
+     */
+    boolean isQueryEnabled();
+
+    /**
+     * Get the port GameSpy 4 queries are accepted on
+     * @return the query port
+     */
+    int getQueryPort();
+
+    /**
+     * Get the map name reported to GameSpy 4 query services
+     * @return the map name
+     */
+    String getQueryMap();
+
+    /**
+     * Get the MOTD component shown in the tab list
+     * @return the motd component
+     */
+    Component getMotdComponent();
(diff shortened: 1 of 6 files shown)


proxy/src/main/java/com/velocitypowered/proxy/connection/client/StatusSessionHandler.java
```diff
@@ -41,7 +41,7 @@ public boolean handle(StatusRequest packet) {
                 new ServerPing.Version(shownVersion, "Velocity " + ProtocolConstants.SUPPORTED_GENERIC_VERSION_STRING),
                 new ServerPing.Players(server.getPlayerCount(), configuration.getShowMaxPlayers(), ImmutableList.of()),
                 configuration.getMotdComponent(),
-                configuration.getFavicon(),
+                configuration.getFavicon().get(),
```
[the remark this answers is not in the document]
Crypnotic: I wrote that with the assumption that the Favicon was either set or null. Is that not the case? If it is, does that need to be changed to make sure nothing weird happens?

proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
```diff
@@ -263,8 +264,8 @@ public int getLoginRatelimit() {
         return advanced.getLoginRatelimit();
     }
 
-    public Favicon getFavicon() {
-        return favicon;
+    public Optional<Favicon> getFavicon() {
+        return Optional.of(favicon);
```
astei: `Optional.ofNullable()`

astei: `Optional.ofNullable()`

api/src/main/java/com/velocitypowered/api/proxy/config/ProxyConfig.java
```diff
@@ -0,0 +1,41 @@
+package com.velocitypowered.api.proxy.config;
+
+import com.velocitypowered.api.util.Favicon;
+import net.kyori.text.Component;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+public interface ProxyConfig {
```
astei: Make sure to add in Javadoc.

Crypnotic: I went ahead and wrote javadoc for a methods, not the best so please feel free to make them not suck so much

PR #108, opened 2018-10-04
Add PROXY protocol support

LXGaming:


proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
```diff
@@ -41,6 +41,10 @@
     @ConfigKey("online-mode")
     private boolean onlineMode = true;
 
+    @Comment("Enables compatibility with the PROXY protocol.")
```
Leymooo: PROXY -> HAPROXY

proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
```diff
@@ -41,6 +41,10 @@
...
+    @ConfigKey("proxy-protocol")
+    private boolean proxyProtocol = false;
```
Leymooo: move this to Advanced

proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
electronicboy: PROXY is correct here, given that it's talking about the actual protocol itself and not a specific implementation

astei:
Going to give this a try today.

astei:
Verified this works. Thanks for the contribution! :)

PR #112, opened 2018-10-10
Add Forge mod list support

LXGaming:
Extends Forge support by adding the ability to process the [ModList](https://wiki.vg/Minecraft_Forge_Handshake#ModList) PluginMessage sent by the Client

### Additions
- PlayerModInfoEvent
- getModInfo method for Player

### Changes
ServerPing.ModList has been moved to it's own class

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/event/player/PlayerModInfoEvent.java
+++ b/api/src/main/java/com/velocitypowered/api/event/player/PlayerModInfoEvent.java
@@ -1,0 +1,23 @@
+package com.velocitypowered.api.event.player;
+
+import com.google.common.base.Preconditions;
+import com.velocitypowered.api.proxy.Player;
+import com.velocitypowered.api.util.ModInfo;
+
+public final class PlayerModInfoEvent {
+    private final Player player;
+    private final ModInfo modInfo;
+    
+    public PlayerModInfoEvent(Player player, ModInfo modInfo) {
+        this.player = Preconditions.checkNotNull(player, "player");
+        this.modInfo = Preconditions.checkNotNull(modInfo, "modInfo");
+    }
+    
+    public Player getPlayer() {
+        return player;
+    }
+    
+    public ModInfo getModInfo() {
+        return modInfo;
+    }
+}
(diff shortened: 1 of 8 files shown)


proxy/src/main/java/com/velocitypowered/proxy/protocol/util/PluginMessageUtil.java
```diff
@@ -76,4 +78,23 @@ public static PluginMessage rewriteMCBrand(PluginMessage message) {
         newMsg.setData(rewrittenData);
         return newMsg;
     }
+    
+    public static List<ModInfo.Mod> readModList(PluginMessage message) {
+        List<ModInfo.Mod> mods = Lists.newArrayList();
```
astei: `immutableList.builder()`

api/src/main/java/com/velocitypowered/api/event/player/PlayerModInfoEvent.java
```diff
@@ -0,0 +1,23 @@
+package com.velocitypowered.api.event.player;
+
+import com.google.common.base.Preconditions;
+import com.velocitypowered.api.proxy.Player;
+import com.velocitypowered.api.util.ModInfo;
+
+public final class PlayerModInfoEvent {
```
astei: Please add Javadoc.

proxy/src/main/java/com/velocitypowered/proxy/protocol/util/PluginMessageUtil.java
dualspiral: Instead of creating the list builder on each and every invocation of this, just create the builder after you verify the discminator is 2 and return the list then. `ImmutableList#of()` just returns a constant so is far cheaper than doing this each and every time you read a `FML|HS` packet.

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientPlaySessionHandler.java
```diff
@@ -147,6 +145,11 @@ public boolean handle(PluginMessage packet) {
             player.getConnectedServer().getConnection().write(PluginMessageUtil.rewriteMCBrand(packet));
         } else if (player.getConnectedServer().isLegacyForge() && !player.getConnectedServer().hasCompletedJoin()) {
             if (packet.getChannel().equals(VelocityConstants.FORGE_LEGACY_HANDSHAKE_CHANNEL)) {
+                List<ModInfo.Mod> mods = PluginMessageUtil.readModList(packet);
```
dualspiral: Once you have the mod list, there is no need to do it again as the mod list won't change for FML clients during the lifetime of their connection. Rather than go through the trouble of reading _every_ `FML|HS` packet and the overhead that brings, I suggest checking to see if the mod list has been set, and once it has, not doing all this.

Note that if it later transpires that we need to track the FML handshake, this can be incorporated into that.

PR #113, opened 2018-10-11
Enhance the Command System

FerusGrim:
I've made a bit of effort to flesh out the command system a bit more. The current implementation is very basic (though usable) and is missing a lot of key features - primarily child commands, which should clear future clutter of giant if-else nests.

Current implementation can be neatly seen in the updated `proxy` packages, but for brevity, I'll include some examples below.

**Registration at VelocityServer**
```java
        commandManager.register(commandManager.builder("velocity")
                .description("Display information about Velocity.")
                .permission("velocity.command.info")
                .executor(new ActualVelocityCommand())
                .build());

        commandManager.register(commandManager.builder("server")
                .description("Display information about or connect to other servers connected to Velocity.")
                .permission("velocity.command.info")
                .permissionState(PermissionState.PLAYER_ONLY)
                .executor(new ServerCommand(commandManager))
                .build());

        commandManager.register(commandManager.builder("shutdown", "end")
                .description("Shut Velocity down.")
                .permissionState(PermissionState.CONSOLE_ONLY)
                .executor(new ShutdownCommand(this))
                .build());
```

**Example `CommandExecutor`: Shutdown Command**
```java
public class ShutdownCommand implements CommandExecutor {

    private final VelocityServer server;

    public ShutdownCommand(VelocityServer server) {
        this.server = server;
    }

    @Override
    public void execute(@NonNull CommandSource source, @NonNull String[] args) {
        this.server.shutdown();
    }

    @Override
    public List<String> suggest(@NonNull CommandSource source, @NonNull String[] args) {
        return ImmutableList.of();
    }
}
```

There are a few further improvements I'd like to make, such as creating a more cohesive tab suggestion method, a more logical permissions setup, and, of course, any improvements suggested here. Before I sink any more time into this, however, I wanted to actually see if there would be more interest in this. It's entirely possible the command system in its current implementation is considered complete and I don't want to waste too much time on something no one wants. 👍 

Loving the direction that Velocity is taking, and I hope to see more improvements in the future. :)


Cybermaxke:
Maybe it would be worth looking at using brigadier ? Since it now open source and licensed under MIT.

https://github.com/Mojang/brigadier

FerusGrim:
@Cybermaxke I'll take a look at the viability of Brigadier. I haven't had a chance to particularly research it since its release.

astei:
I'm reluctant to merge this in, at least at this point. I'm looking to stabilize the API and do an initial 1.0.0 release, and this is a breaking change.

I'm thinking that a revamped command system would be in the cards with Velocity 2.0.0, amongst other changes I'd like to make.

kashike:
Big :-1: from me as-is - this makes using other external systems, such as Brigadier, a massive pain.

api/src/main/java/com/velocitypowered/api/command/Command.java
```diff
@@ -1,44 +1,107 @@
...
 public interface Command {
     /**
-     * Executes the command for the specified {@link CommandSource}.
-     * @param source the source of this command
-     * @param args the arguments for this command
+     * Gets the CommandManager for this command.
+     * @return the CommandManager
      */
-    void execute(@NonNull CommandSource source, @NonNull String[] args);
+    CommandManager getCommandManager();
```
kashike: Why the heck does the command store this? :-1:

lucko:
My opinion is that it's best for the command API in Velocity to be as minimal as possible (i.e. the current system).

If plugin authors want to use a more comprehensive command framework, then they can evaluate the available options, and pick one of them.

* [EngineHub/intake](EngineHub/Intake)
* [aikar/commands](https://github.com/aikar/commands)
* [Mojang/brigadier](https://github.com/Mojang/brigadier)
* [cbeust/jcommander](https://github.com/cbeust/jcommander)

to name just a few of the popular libraries available.

#### Key benefits to this approach IMO :)

* Plugins that don't care about using a framework (or that use their own built-in framework) don't have to work around / learn a clunky API just to register a command.
* Plugins that *do* want to use a framework have the choice between a variety of purpose-built systems, and can pick the one which suits them the best - be that annotation, functional/builder style, etc etc.
* Velocity team don't have to develop/maintain/support a system that isn't crucial for the operation of a "Minecraft proxy".

@FerusGrim Just to be clear, I'm not commenting on your work here, merely just my opinion on what I think should be included. 😃

FerusGrim:
Fair enough, @lucko. That's partly why I posted the PR so quickly - I wasn't sure if it was something anyone even wanted. :)

DoNotSpamPls:
While the code here is very good, I don't think it's a good fit, as others have said. It'd be much better if you put your free time to work on the [issues that are currently present](https://github.com/VelocityPowered/Velocity/issues), rather than fixing an issue no one really has at the moment.

kashike:
I'm going to go ahead and close this pull request for the reasons @lucko gave above.

PR #114, opened 2018-10-12
Extract channel initialiser and transport type out

kashike:


proxy/src/main/java/com/velocitypowered/proxy/network/ConnectionManager.java
```diff
@@ -1,89 +1,49 @@
...
-                        connection.setState(StateRegistry.HANDSHAKE);
-                        connection.setSessionHandler(new HandshakeSessionHandler(connection, server));
-                        ch.pipeline().addLast(Connections.HANDLER, connection);
-                        
-                        if (server.getConfiguration().isProxyProtocol()) {
-                            ch.pipeline().addFirst(new HAProxyMessageDecoder());
-                        }
-                    }
-                })
+                .childHandler(new ServerChannelInitializer(this.server))
```
Leymooo: Does we really need to create a new object every time?  It's also does not allow to rewrite a ChannelInitializer.

proxy/src/main/java/com/velocitypowered/proxy/network/ConnectionManager.java
```diff
@@ -1,89 +1,52 @@
...
-    public ConnectionManager(VelocityServer server) {
+    public ConnectionManager(final VelocityServer server) {
         this.server = server;
         this.transportType = TransportType.bestType();
-        this.bossGroup = transportType.createEventLoopGroup(true);
-        this.workerGroup = transportType.createEventLoopGroup(false);
+        this.bossGroup = this.transportType.createEventLoopGroup(TransportType.Type.BOSS);
+        this.workerGroup = this.transportType.createEventLoopGroup(TransportType.Type.WORKER);
+        this.serverChannelInitializer = new ServerChannelInitializerHolder(new ServerChannelInitializer(this.server));
+        this.serverChannelInitializer.set(new ServerChannelInitializer(this.server));
```
Leymooo: why do you call `set` here?

kashike: Shhhh.

Issue #110, opened 2018-10-08
Reload command


astei:
In essence:

* Reload the `velocity.toml` in-place as much as reasonably (and safely) possible.
* Fire a `ProxyReloadEvent` to allow plugins to reload their own configurations.

astei:
See #139 for most recent work on this.

LartyHD:
Does anyone else work here?

astei:
I work on Velocity in whatever time I have between my contracted work and college.

Anyway, yes, this is fixed. I've been busy with finals and so looking and caring for the issue tracker hasn't been a big priority for me.

PR #120, opened 2018-10-23
[WIP] Reload (sub)command

mikroskeem:
Implementation for #110 

TODO:
- [x] Fire ProxyReloadEvent
- [x] What should happen when Velocity configuration fails to load?
- [x] What should happen when plugin configuration fails to load?
- [x] Should we show different message when configuration fails to load?
- [x] Think about better command name than `reload` (conflicts with servers)
- [ ] Merge server entries
- [ ] Reconfigure Ratelimiter
- [ ] Re-bind client and query listeners (?)

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
@@ -7,6 +7,7 @@ import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import com.velocitypowered.api.event.EventManager;
 import com.velocitypowered.api.event.proxy.ProxyInitializeEvent;
+import com.velocitypowered.api.event.proxy.ProxyReloadEvent;
 import com.velocitypowered.api.event.proxy.ProxyShutdownEvent;
 import com.velocitypowered.api.plugin.PluginContainer;
 import com.velocitypowered.api.plugin.PluginManager;
@@ -57,6 +58,7 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.checkerframework.checker.nullness.qual.EnsuresNonNull;
 import org.checkerframework.checker.nullness.qual.MonotonicNonNull;
+import org.checkerframework.checker.nullness.qual.Nullable;
 import org.checkerframework.checker.nullness.qual.RequiresNonNull;
 
 public class VelocityServer implements ProxyServer {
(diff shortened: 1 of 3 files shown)


astei:
> What should happen when Velocity configuration fails to load?

Retain the existing configuration and warn the user.

> What should happen when plugin configuration fails to load?

We can't control what plugins want to do.

> Think about better command name than reload (conflicts with servers)

Perhaps we could turn `/velocity` into a large command with subcommands (think of `/velocity about`, `/velocity reload`, `/velocity dump`, ...)

astei:
I committed c977ddec which might be of use for you.

api/src/main/java/com/velocitypowered/api/proxy/ProxyServer.java
```diff
@@ -131,4 +132,9 @@
      * @return the proxy config
      * */
     ProxyConfig getConfiguration();
+
+    /**
+     * Reloads proxy configuration and fires {@link ProxyReloadEvent}
+     */
+    void reloadConfiguration();
```
mikroskeem: Should this throw checked exception? Because then there's a way to force checking if reload failed (and if yes, then catching exception and figuring out what happened).

proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
```diff
@@ -90,6 +91,18 @@ public VelocityConfiguration getConfiguration() {
         return configuration;
     }
 
+    @Override
+    public void reloadConfiguration() {
+        // TODO: reload configuration
+
+        // Give a chance for plugins to reload as well
+        try {
+            eventManager.fire(new ProxyReloadEvent()).get();
```
mikroskeem: Not entirely sure if this should be blocking

api/src/main/java/com/velocitypowered/api/proxy/ProxyServer.java
astei: I think it's better to not have this in the API.

proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
astei: It shouldn't be. Use `fireAndForget()`.

astei:
Well, @mikroskeem has no time to work on this, so I am working on this now. See my [branch](https://github.com/VelocityPowered/Velocity/tree/reload-command) where I'm taking a slightly different approach.

Issue #116, opened 2018-10-14
Native forced host support


astei:
Basically, add hosts that can direct to a specified set of initial connection servers that can be configured in `velocity.toml` as follows:

```toml
[forced-hosts]
"ftb.example.com" = ["ftb-lobby-1", "ftb-lobby-2"]
```

Shybella:
Yes, that's a neat idea.

astei:
Pulled #123 which resolves this issue

PR #123, opened 2018-10-23
[WIP] Forced hosts configuration support

mikroskeem:
Solves #116 

TODO:
- [x] Configuration bits
- [x] More fool-proof configuration reading
- [x] Hook forced hosts up to Velocity internals
- [x] Should invalid server entries get removed internally?

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
@@ -2,6 +2,7 @@ package com.velocitypowered.proxy.config;
 
 import com.google.common.base.MoreObjects;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableList;
 import com.moandjiezana.toml.Toml;
 import com.velocitypowered.api.proxy.config.ProxyConfig;
 import com.velocitypowered.api.util.Favicon;
@@ -65,6 +66,9 @@ public class VelocityConfiguration extends AnnotatedConfig implements ProxyConfi
     @Table("[servers]")
     private final Servers servers;
 
+    @Table("[forced-hosts]")
+    private final ForcedHosts forcedHosts;
+
     @Table("[advanced]")
     private final Advanced advanced;
 
@@ -76,15 +80,16 @@ public class VelocityConfiguration extends AnnotatedConfig implements ProxyConfi
     @Ignore
     private Favicon favicon;
 
-    public VelocityConfiguration(Servers servers, Advanced advanced, Query query) {
+    public VelocityConfiguration(Servers servers, ForcedHosts forcedHosts, Advanced advanced, Query query) {
         this.servers = servers;
+        this.forcedHosts = forcedHosts;
         this.advanced = advanced;
         this.query = query;
     }
 
     private VelocityConfiguration(String bind, String motd, int showMaxPlayers, boolean onlineMode,
             boolean announceForge, PlayerInfoForwarding playerInfoForwardingMode, byte[] forwardingSecret,
-            Servers servers, Advanced advanced, Query query) {
+            Servers servers, ForcedHosts forcedHosts, Advanced advanced, Query query) {
         this.bind = bind;
         this.motd = motd;
         this.showMaxPlayers = showMaxPlayers;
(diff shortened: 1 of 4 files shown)


proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
```diff
@@ -118,6 +118,14 @@ public void start() {
             servers.register(new ServerInfo(entry.getKey(), AddressUtil.parseAddress(entry.getValue())));
         }
 
+        for (Map.Entry<String, List<String>> entry : configuration.getForcedHosts().entrySet()) {
+            for (String server : entry.getValue()) {
+                if (!servers.getServer(server).isPresent()) {
+                    logger.warn("Server '{}' for forced host '{}' does not exist", server, entry.getKey());
```
mikroskeem: IIRC invalid server entries cause premature disconnection from proxy, e.g if 3rd server is invalid and there are 7 servers in list, then player would get disconnected on 3rd server connection attempt.

So should list be copied and rewritten here or should logic be changed where disconnection happens?

astei: Your concern is unfounded. Velocity will try to connect to all seven servers, and only after all attempts are exhausted will it disconnect the server.

I would prefer this were a fatal error that would prevent the configuration from being loaded.

proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
```diff
@@ -118,6 +118,16 @@ public void start() {
             servers.register(new ServerInfo(entry.getKey(), AddressUtil.parseAddress(entry.getValue())));
         }
 
+        for (Map.Entry<String, List<String>> entry : configuration.getForcedHosts().entrySet()) {
```
astei: This check should be performed in `VelocityConfiguration.validate()`.

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
```diff
@@ -344,7 +347,11 @@ private void handleConnectionException(RegisteredServer rs, @Nullable Component
     }
 
     Optional<RegisteredServer> getNextServerToTry() {
-        List<String> serversToTry = server.getConfiguration().getAttemptConnectionOrder();
+        if (serversToTry == null) {
+            Optional<List<String>> forcedHosts = getVirtualHost().map(vHost -> server.getConfiguration().getForcedHosts().get(vHost.getHostString()));
```
astei: This code will trigger the Optional Checker in Checker Framework. Prefer a `virtualHost != null` check and `server.getConfiguration().getForcedHosts().getOrDefault(virtualHost, server.getConfiguration().getAttemptConnectionOrder())`.

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
```diff
@@ -62,6 +64,7 @@
     private final InetSocketAddress virtualHost;
     private GameProfile profile;
     private PermissionFunction permissionFunction = null;
+    private List<String> serversToTry = null;
```
astei: I forgot! This should be marked `@MonotonicNonNull`.

proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
```diff
@@ -421,6 +443,61 @@ public String toString() {
...
+        private ForcedHosts() {}
+
+        private ForcedHosts(Toml toml) {
+            if (toml != null) {
+                Map<String, List<String>> forcedHosts = new HashMap<>();
+                for (Map.Entry<String, Object> entry : toml.entrySet()) {
+                    if (entry.getValue() instanceof String) {
+                        forcedHosts.put(stripQuotes(entry.getKey()), ImmutableList.of((String) entry.getValue()));
+                    } else if (entry.getValue() instanceof List) {
+                        forcedHosts.put(stripQuotes(entry.getKey()), ImmutableList.copyOf((List<String>) entry.getValue()));
```
astei: Make sure the list is not empty.

PR #130, opened 2018-11-03
Backend encryption

creeper123123321:


proxy/src/main/java/com/velocitypowered/proxy/util/EncryptionUtils.java
```diff
@@ -10,10 +10,12 @@
 import java.security.PublicKey;
 import java.security.SecureRandom;
 import java.security.spec.X509EncodedKeySpec;
+import java.util.Random;
 import javax.crypto.Cipher;
 
 public enum EncryptionUtils {
   ;
```
DoNotSpamPls: what is this

creeper123123321: ``public *enum* EncryptionUtils``

astei:
For the record, an empty enumeration (one with no values) is totally valid, [although some may be surprised as to what to expect](https://stackoverflow.com/questions/25137490/zero-instance-enum-vs-private-constructors-for-preventing-instantiation).

For the future, we will follow the typical convention of declaring a `final` class with a private constructor that throws an exception when instantiated.

astei:
I'm not totally sure if there is a general use case for enabling Minecraft protocol-level encryption support, since the encryption request is generally sent when the server is set to online-mode authentication. This could lead to confusion for people who don't know that Velocity has to be able to fake handshake data anyway.

astei:
Closing this PR, given that there has been no response in over two weeks and the potential confusion this feature could generate.

PR #131, opened 2018-11-08
Fix null Mod ID due to gson deserialization.

PurpleIsEverything:


hugmanrique:
As far as I can tell, this only changes the field name which 1. won't fix the null issue and 2. Is unneeded, because the id field is a member of the ModInfo class which you can already tell it's gonna be the mod's ID.

PurpleIsEverything:
Gson deserialization of the Mod class is dependent on field names. If you look at the raw JSON from the server ping you'd see that it uses modid as it's key instead of id.

This is not something I just changed randomly, I tested this before this PR was made this is the required change. The only reason I didn't change the method call was for API consistency.

PR #132, opened 2018-11-12
Optimize GameProfile

Yeregorix:
Hi !
I've decided to make a small PR to modify `GameProfile`. My main problem was that `Player#getUniqueId` was parsing the UUID from a String for every method call which can impact plugins performance. I've optimized this method and also few other pieces of code to minimize the usage of `UuidUtils#toUndashed` and `UuidUtils#fromUndashed`. To finish, I've added an interface to generalize the concept of object with an unique id: `Identifiable`, which can be really useful for some developers.
This is my first PR for the Velocity project, I try to do my best but if something is wrong tell me how I can improve it. 😃


mikroskeem:
What else could be identified by UUIDs in Velocity? :thinking:

api/src/main/java/com/velocitypowered/api/util/GameProfile.java
```diff
@@ -36,6 +48,36 @@ public String getName() {
     return properties;
   }
 
+  public GameProfile setUniqueId(UUID id) {
```
astei: There is no Javadoc here, but more importantly these names are misleading since the class is immutable and `set<something>` is typically used for mutable objects. Prefer `with<something>` instead.

api/src/main/java/com/velocitypowered/api/util/Identifiable.java
```diff
@@ -0,0 +1,11 @@
+package com.velocitypowered.api.util;
+
+import java.util.UUID;
+
+/**
+ * Represents an object that can be identified by its UUID
+ */
+public interface Identifiable {
+
+	UUID getUniqueId();
```
astei: Not in code style

api/src/main/java/com/velocitypowered/api/util/Identifiable.java
```diff
@@ -0,0 +1,11 @@
+package com.velocitypowered.api.util;
+
+import java.util.UUID;
+
+/**
+ * Represents an object that can be identified by its UUID
+ */
+public interface Identifiable {
```
astei: What purpose does this interface really have? I don't see any real reason for this.

api/src/main/java/com/velocitypowered/api/util/GameProfile.java
```diff
@@ -8,24 +8,36 @@
...
   }
 
-  public String getId() {
-    return id;
+  public GameProfile(String undashedId, String name, List<Property> properties) {
+    this(UuidUtils.fromUndashed(Preconditions.checkNotNull(undashedId, "undashedId")), undashedId,
+        Preconditions.checkNotNull(name, "name"), ImmutableList.copyOf(properties));
   }
 
-  public UUID idAsUuid() {
```
astei: Breaking API change. Please revert.

api/src/main/java/com/velocitypowered/api/util/GameProfile.java
```diff
@@ -8,24 +8,36 @@
 /**
  * Represents a Mojang game profile. This class is immutable.
  */
-public final class GameProfile {
+public final class GameProfile implements Identifiable {
 
-  private final String id;
-  private final String name;
+  private final UUID id;
```
astei: To negate the need for a custom `GameProfileSerializer`:

* Keep `id` as `String`
* Add a new `UUID` field (say `javaUuid`) with `transient` keyword.

api/src/main/java/com/velocitypowered/api/util/GameProfile.java
Yeregorix: `idAsUuid` sounded a bit unclear to me and I didn't know if you wanted to keep the API unchanged or not but I will revert it.

api/src/main/java/com/velocitypowered/api/util/Identifiable.java
Yeregorix: No reason yet in your implementation but it can be useful for some plugins to generalize some method related to these objects. In Velocity, the only affected classes are `Player` and `GameProfile` but as an example, a plugin might want to have a generic method to retrieve some data from its database according the id of the object passed as argument, whether it's a `Player`, a `GameProfile` or any custom object implementing `Identifiable`. Another example, it can be used with an `instanceof` to check whether an object has an unique id attached or not. A such interface can't hurt, but only help plugin developers. 😄 You will say, okay, so let them add their own interface but it would loose the interest to have a common a interface for every plugin and the API.

api/src/main/java/com/velocitypowered/api/util/GameProfile.java
Yeregorix: The problem is that it would allow to create an invalid `GameProfile` object when the `id` is not a valid undashed uuid which can cause an exception when calling `GameProfile#idAsUuid`. Do you think such objects should be allowed ? It does not only affect gson, it was also possible by calling the constructor with an invalid parameter.

astei: If that's a concern I would probably validate the ID. This can be done trivially.

astei: Actually, you are correct - this just hasn't come up in practice. I'm fine with keeping the `GameProfileSerializer`.

api/src/main/java/com/velocitypowered/api/util/GameProfile.java
Yeregorix: Should I deprecate this method in favor of `getUniqueId` or remove `getUniqueId` and thus probably `Identifiable` by the way, or just keep both methods without deprecating ?

astei: Deprecation would be acceptable if we keep `getUniqueId()`, but I'm still not sold on the need for the `Identifiable` interface.

api/src/main/java/com/velocitypowered/api/util/Identifiable.java
astei: I'm still not sold.

> a plugin might want to have a generic method to retrieve some data from its database according the id of the object passed as argument

There is already a standard object plugins can use to represent an UUID. It is called `java.util.UUID`. And getting the UUID is trivially done, given that there are only two objects in the Velocity API that even have a UUID:

```java
public static UUID getUuid(Object o) {
  if (o instanceof Player) return ((Player) o).getUniqueId();
  if (o instanceof GameProfile) return ((GameProfile) o).idAsUuid();
  throw new IllegalArgumentException("Provided object " + o + " doesn't have a UUID");
}
```

If a plugin needs a custom object for representing a user, it's probably going to implement a method similar to what I've provided for Velocity above.

There is no direct use of `Identifiable` in the Velocity API I can see providing convenience and I honestly don't think there is a use of this interface where a plain `java.util.UUID` would equally suffice.

Even more embarrassingly, let's assume you adjusted `ProxyServer#getPlayer()` to take an `Identifiable`. Does this code make sense?

```java
Player p = event.getPlayer();

// ... a lot of code ...
Player p1 = server.getPlayer(p); // but you already have the Player!
```

> You will say, okay, so let them add their own interface but it would loose the interest to have a common a interface for every plugin and the API.

But the common interface we could define here simply does not exist. An example of a common interface would be [PermissionFunction](https://jd.velocitypowered.com/com/velocitypowered/api/permission/PermissionFunction.html) - a common interface for plugins to define what permissions some subject has. It's useful - this is the foundation of the Velocity permisisons system. Where's the common interface and use for `Identifiable`? None, as far as I can tell.

Yeregorix:
```java
Player p = event.getPlayer();

// ... a lot of code ...
Player p1 = server.getPlayer(p); // but you already have the Player!
```

I agree this code is almost useless I more thinking of an example like this one:
Imagine another plugin added a custom CommandSource also implementing Identifable.
Now, in my plugin, I'm implementing this command:

```java
  @Override
  public void execute(CommandSource source, String[] args) {
    if (source instanceof Identifiable) { // Can't be replaced without sacrificing compatibility with others plugins
      User user = User.of((Identifiable) source); // Can be replaced by User.ofId(((Identifiable) source).getUniqueId())
      source.sendMessage(TextComponent.of("Hello " + user.getDisplayName() + " !"));
    } else {
      source.sendMessage(TextComponent.of("Sorry, but this command is not available for a source of your type :(."));
    }
  }
```

Once again it's a detail. If you really don't want this interface, just tell me and I will remove it. To finish it makes things slightly more homogenous. I don't see any reason to call two identical methods (`Player#getUniqueId` and `GameProfile#idAsUuid`) with differents names.

astei: I'm still not sold on it. I'd prefer if `Identifiable` was removed.

Yeregorix: Done. I would have tried 😶

api/src/main/java/com/velocitypowered/api/util/GameProfile.java
Yeregorix: I saw a message saying that you're fine with keeping the `GameProfileSerializer` but this message is no longer here. Do you still want to keep it or not ?

astei: We can keep it.

PR #136, opened 2018-11-15
Refactor Forge support to track the handshake & include keepalives

dualspiral:
This commit adds a basic Forge handshake state tracker for determining where we are in the forge cycle. Right this second, the only useful states are the beginning and end states, but because of the way the forge handshake works, we need to keep abreast of everything.

We also send keepalives just before server switch, and just after the FML handshake has completed.

Any future work on unstable connections will be done on top of this, so it makes sense to get this in place now.


dualspiral:
I've moved nearly all Forge code into Forge specific classes. There are a couple of lines here and there that would just be a pain to move, but for the most part, this should aid maintainability.

I've dealt with my biggest bugbear, that is that I'd added legacy forge bools and whatnot into the connection classes. These have all been replaced with enums/constants/classes that represent connection "types" and "phases", that should make it much easier to extend support for other connection types in the future, mostly thinking Forge 1.13+.

This still isn't ready for merge, I've not fully tested yet and there are still some debugging bits and bobs in here, but this is the direction I was thinking of when refactoring Forge support.

proxy/src/main/java/com/velocitypowered/proxy/connection/MinecraftConnection.java
```diff
@@ -279,27 +278,27 @@ public void setAssociation(MinecraftConnectionAssociation association) {
...
 
   public void setNextProtocolVersion(int nextProtocolVersion) {
     this.nextProtocolVersion = nextProtocolVersion;
   }
+
+  /**
+   * Gets the detected {@link ConnectionType}
+   * @return The {@link ConnectionType}
+   */
+  public ConnectionType getConnectionType() {
```
kashike: You can drop `Connection` from the name - `connection.getType()` is clear enough, don't need `connection.getConnectionType()`.

proxy/src/main/java/com/velocitypowered/proxy/connection/MinecraftConnection.java
```diff
@@ -279,27 +278,27 @@ public void setAssociation(MinecraftConnectionAssociation association) {
...
+    return connectionType;
+  }
+
+  /**
+   * Sets the detected {@link ConnectionType}
+   * @param connectionType The {@link ConnectionType}
+   */
+  public void setConnectionType(ConnectionType connectionType) {
```
kashike: You can drop `Connection` from the name - `connection.getType()` is clear enough, don't need `connection.getConnectionType()`.

dualspiral:
[velocity-proxy-1.0-SNAPSHOT-all.zip](https://github.com/VelocityPowered/Velocity/files/2609438/velocity-proxy-1.0-SNAPSHOT-all.zip)

For testing only - as requested in a private conversation.

proxy/src/main/java/com/velocitypowered/proxy/connection/ConnectionType.java
```diff
@@ -0,0 +1,69 @@
...
+ * The types of connection that may be selected.
+ */
+public interface ConnectionType {
+
+  /**
+   * Indicates that the connection has yet to reach the
+   * point where we have a definitive answer as to what
+   * type of connection we have.
+   */
+  ConnectionType UNDETERMINED =
```
astei: I would strongly recommend these to be factored out into a `ConnectionTypes` class. Keeping these in the interface clutters things up.

proxy/src/main/java/com/velocitypowered/proxy/connection/backend/BackendConnectionPhase.java
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * <p>Note that Forge phases are found in the enum
+ * {@link LegacyForgeHandshakeBackendPhase}.</p>
+ */
+public interface BackendConnectionPhase {
+
+  /**
+   * The backend connection is vanilla.
+   */
+  BackendConnectionPhase VANILLA = new BackendConnectionPhase() {};
```
astei: Same for `ConnectionType`

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientConnectionPhase.java
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * <p>Note that Forge phases are found in the enum
+ * {@link LegacyForgeHandshakeClientPhase}.</p>
+ */
+public interface ClientConnectionPhase {
+
+  /**
+   * The client is connecting with a vanilla client (as far as we can tell).
+   */
+  ClientConnectionPhase VANILLA = new ClientConnectionPhase() {};
```
astei: See `ConnectionType` comment.

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
```diff
@@ -64,6 +66,7 @@
   private static final PlainComponentSerializer PASS_THRU_TRANSLATE = new PlainComponentSerializer(
       c -> "", TranslatableComponent::key);
   static final PermissionProvider DEFAULT_PERMISSIONS = s -> PermissionFunction.ALWAYS_UNDEFINED;
+  private static final Random RANDOM = new Random();
```
astei: Prefer `ThreadLocalRandom` in all cases.

PR #137, opened 2018-11-18
Add option to get command label in Command#execute.

bhopahk:
This adds another method to [Command](https://jd.velocitypowered.com/com/velocitypowered/api/command/Command.html) which supplies the label of the command being run. Additionally, both the new and old execute methods default to do nothing.

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/command/Command.java
+++ b/api/src/main/java/com/velocitypowered/api/command/Command.java
@@ -16,7 +16,16 @@ public interface Command {
    * @param source the source of this command
    * @param args the arguments for this command
    */
-  void execute(CommandSource source, String @NonNull [] args);
+  default void execute(CommandSource source, String @NonNull [] args) {}
+
+  /**
+   * Executes the command for the specified {@link CommandSource}.
+   *
+   * @param source the source of this command
+   * @param label the label for this command
+   * @param args the arguments for this command
+   */
+  default void execute(CommandSource source, @NonNull String label, String @NonNull [] args) {}
 
   /**
    * Provides tab complete suggestions for a command for a specified {@link CommandSource}.
--- a/api/src/main/java/com/velocitypowered/api/command/Command.java
+++ b/api/src/main/java/com/velocitypowered/api/command/Command.java
@@ -16,16 +16,17 @@ public interface Command {
    * @param source the source of this command
    * @param args the arguments for this command
    */
+  @Deprecated
   default void execute(CommandSource source, String @NonNull [] args) {}
 
   /**
    * Executes the command for the specified {@link CommandSource}.
    *
    * @param source the source of this command
-   * @param label the label for this command
+   * @param alias the alias for this command
    * @param args the arguments for this command
    */
-  default void execute(CommandSource source, @NonNull String label, String @NonNull [] args) {}
+  default void execute(CommandSource source, @NonNull String alias, String @NonNull [] args) {}
 
   /**
    * Provides tab complete suggestions for a command for a specified {@link CommandSource}.
--- a/proxy/src/main/java/com/velocitypowered/proxy/command/VelocityCommandManager.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/command/VelocityCommandManager.java
@@ -58,6 +58,7 @@ public class VelocityCommandManager implements CommandManager {
       }
 
       command.execute(source, actualArgs);
+      command.execute(source, split[0], actualArgs);
       return true;
     } catch (Exception e) {
       throw new RuntimeException("Unable to invoke command " + cmdLine + " for " + source, e);


proxy/src/main/java/com/velocitypowered/proxy/command/VelocityCommandManager.java
```diff
@@ -58,6 +58,7 @@ public boolean execute(@NonNull CommandSource source, @NonNull String cmdLine) {
       }
 
       command.execute(source, actualArgs);
+      command.execute(source, split[0], actualArgs);
```
hugmanrique: Calling the same method with different params might be confusing for API callers. As @astei said on Discord, you already have the command label when you register the command, so I don't fully get the point of this. What's your proposed use case?

bhopahk: To allow multiple commands being sent to the same listener to be sorted from there. A more specific example is having a single class route different commands to the same place without having all of them be subcommands of or different classes. IE /ban and /tempban. The only different between them is that tempban needs to look for one more argument of time and without being able to find out the label of the command being run, I would be required to create two almost identical listeners.

Perhaps a better solution would be to have a single execute which gives the label and allow people to ignore it (as done in the Bukkit/Spigot API) or perhaps rename the execute with a label to something more identifiable.

hugmanrique: Have you thought about creating a `BanCommand` abstract class/interface that subclasses could extend i.e. `TempBanCommand` has to implement the method `ban(Player admin, String hacker, String[] args)` where it could grab the ban duration, reason, etc.

bhopahk: Yeah, though I am creating a set of my commands which are unrelated to velocity (to allow for me to create implementations in more apis easily) so i am sending commands through a single class to route them to the common version of the command and I was hoping to avoid having two different sets of each command. To your point though, it would absolutely work it is another set of classes that I was hoping to avoid writing.

Leymooo: You can use one class but different objects. Like
```java
public class LabeledCommand implements Command {

    private final String label;
    public  LabeledCommand(ProxyServer server, String label) {
        this.server = server;
        this.label = label;
    }

    @Override
    public void execute(@NonNull CommandSource source, String[] args) {
       if (label.equals("ban") {
         //
       } else if ...{
       //
       } else {
       }
    }
}
```
```java
        server.getCommandManager().register(new  LabeledCommand(server, "ban"), "ban");
        server.getCommandManager().register(new  LabeledCommand(server, "tempban"), "tempban");
        server.getCommandManager().register(new  LabeledCommand(server, "kick"), "kick");
```

bhopahk: I suppose the above is a valid solution, though that requires many instances of it would I would prefer to avoid.

dualspiral:
I think you're trying to solve the problem in the wrong way, and by doing so, leaking an implementation detail that you could run afoul with.

When you register your "ban" command, you may not be the only person registering that. In other Minecraft systems, for example, they also get registered with a prefix, their plugin ID. Now, I haven't studied the Velocity system, but if Velocity did this (now, or in the future) and ran your command with, say `/pluginid:ban`, the label would be `pluginid:ban` - how would your command deal with that? You now need to know about _implementation_ details of Velocity, things that might change at any time.

Java is an object oriented language - what is wrong with multiple instances of an object?  Why don't you want to do that? That's what these languages are _designed_ for. Sure, it takes a bit of memory, but it's not that much and it's not going to cause Velocity to run out of memory. The simplest solution is to do as @Leymooo suggested, though it's not necessarily the best. You could also create a `BanCommand` which your `TempBanCommand` inherits from and adds to the ban logic, or extract common methods, etc. etc.

Ultimately, I think this is a bad idea and is not going to solve your problem in the way you think it will. Just because Bukkit did it doesn't mean it's the right thing to do. However, if the label is going to be exposed, you might as well break the API, using default interfaces with two methods is going to confuse novice developers who are trying to learn Java as they write a plugin because it's not clear which one you use, and those who implement both will end up with the command executed _twice_.

api/src/main/java/com/velocitypowered/api/command/Command.java
```diff
@@ -16,7 +16,16 @@
...
    * @param args the arguments for this command
    */
-  void execute(CommandSource source, String @NonNull [] args);
+  default void execute(CommandSource source, String @NonNull [] args) {}
+
+  /**
+   * Executes the command for the specified {@link CommandSource}.
+   *
+   * @param source the source of this command
+   * @param label the label for this command
```
kashike: `alias` instead of `label`

api/src/main/java/com/velocitypowered/api/command/Command.java
```diff
@@ -16,7 +16,16 @@
...
+   * @param args the arguments for this command
+   */
+  default void execute(CommandSource source, @NonNull String label, String @NonNull [] args) {}
```
kashike: `alias` instead of `label`

bhopahk:
Closed in favor of [CommandRanEvent](https://github.com/VelocityPowered/Velocity/pull/138). Less confusing for beginners and gives more functionality.

PR #138, opened 2018-11-18
Implement CommandEvent.

Bennyboy1695:
This implements a CommandRanEvent which allows you to track commands being run as well as cancel and replace the Command and arguments.

I placed the event into the player package as i wasn't sure where else it would fit with the current packages available.


mikroskeem:
Event name implies that command is processed and already evaluated by Velocity, how about simply `CommandEvent`?

>I placed the event into the player package as i wasn't sure where else it would fit with the current packages available.

I'd create a separate package, `command` for example

api/src/main/java/com/velocitypowered/api/event/command/CommandEvent.java
```diff
@@ -0,0 +1,130 @@
...
+
+import com.google.common.base.Preconditions;
+import com.velocitypowered.api.event.ResultedEvent;
+import com.velocitypowered.api.proxy.Player;
+import java.util.Optional;
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+/**
+ * This event is fired when a player runs a command!.
```
kashike: Remove the `!`.

phase:
`CommandExecutedEvent` would be my preference.

kashike:
`CommandExecuteEvent` - command hasn't been executed at the point the event will be posted.

proxy/src/main/java/com/velocitypowered/proxy/command/VelocityCommandManager.java
```diff
@@ -56,8 +64,14 @@ public boolean execute(@NonNull CommandSource source, @NonNull String cmdLine) {
...
         return false;
       }
-
-      command.execute(source, actualArgs);
+      CommandExecuteEvent event = new CommandExecuteEvent(source, alias, actualArgs);
+      server.getEventManager().fire(event)
+              .thenAcceptAsync(cee -> {
+                CommandExecuteEvent.CommandResult commandResult = cee.getResult();
+                if (commandResult.isAllowed()) {
+                  command.execute(source, actualArgs);
```
astei: This is incorrect, since it will not reflect the result a plugin may choose to set.

Bennyboy1695: What would the prefered result be ?

astei: Okay. Right now your `CommandResult` doesn't actually let people replace the command (although the fields and getters exist). 

Either add this functionality or remove the redundant fields.

bhopahk:
In reference to [this](https://ptb.discordapp.com/channels/472484458856185878/472486015525847050/561374383478669312), I am not sure entirely what is still required for this PR, it seemed to somewhat die, but if it was updated to fix the conflicts are you willing to accept it?

astei:
I will need to re-implement this PR for Velocity 1.1.0 anyway. I'm closing this issue for now.

PR #145, opened 2018-11-22
Improvements in VelocityCommand

LartyHD:
Added too many args check to SubCommand: "Info".
Added a permission check: If a source does not have permission for all subcommands, it does not have permission on the whole command because i think the usage: "/velocity <>" is really bad.


proxy/src/main/java/com/velocitypowered/proxy/command/VelocityCommand.java
```diff
@@ -113,6 +116,11 @@ private Info(ProxyServer server) {
 
     @Override
     public void execute(CommandSource source, String @NonNull [] args) {
+      if (args.length != 0) {
```
supertassu:
```suggestion
      if (args.length == 0) {
```

dualspiral: No, what he’s suggesting is right, he’s saying that `/velocity info` doesn’t accept any arguments, which is true.

PR #149, opened 2018-11-24
Send only one respawn packet on server switch

Leymooo:
We can simply set dimension in JoinGame packet and send only one respawn packet. Useless, but why not?

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientPlaySessionHandler.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientPlaySessionHandler.java
@@ -314,11 +314,8 @@ public class ClientPlaySessionHandler implements MinecraftSessionHandler {
       // Most notably, by having the client accept the join game packet, we can work around the need
       // to perform entity ID rewrites, eliminating potential issues from rewriting packets and
       // improving compatibility with mods.
+      joinGame.setDimension(joinGame.getDimension() == 0 ? -1 : 0);
       player.getConnection().delayedWrite(joinGame);
-      int tempDim = joinGame.getDimension() == 0 ? -1 : 0;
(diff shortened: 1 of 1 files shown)


dualspiral:
Fairly sure that this would break mods, with this, the client and server may not be in sync with the dimension ID (client might think ID -1, while the server might think ID 0), and so any client/server mod combo that relies on that will fail.

You could get away with only sending two packets if the incoming ID is the same as the outgoing one.

Leymooo:
Why it should break mods? Velocity send 2 respawn packets to clear chunk cache(first with different dim and second with joinGame dim), this will trigger a world loading screen and clear chunk cache.

My pr also will trigger world loading screen and will clear chunk cache.

Just some code from MC.
```java
    public void handleJoinGame(SPacketJoinGame packetIn)
    {
        ....
        this.gameController.player.dimension = packetIn.getDimension();
        ....
    }
    public void handleRespawn(SPacketRespawn packetIn)
    {
        ...
        if (packetIn.getDimensionID() != this.gameController.player.dimension)
        {
              //some code to clear chunk cache and show world loading screen.
        }
       ...
    }
```

dualspiral:
Ok, sure. I don’t have my MCP workspace to hand to look at the specifics, though mods do funny things that no one ever expects. It would make sense to test this with a modpack or two to ensure compatibility. I would also assume that this is the same behaviour since 1.8.

My comment was otherwise valid for the code as it stood so was not totally baseless, (the wrong ID was being used in the response packets and thus was the original reasoning for my comment, even if my ultimate statement was wrong), but I see that’s been rectified now so no harm done.

astei:
You might want to update the comment.

If this is safe I'm fine with merging this in.

astei:
Closing due to inactivity.

PR #158, opened 2018-12-04
Some event bus related fixes

lucko:
* The first commit addresses a small bug I came across
* The other commits make some small optimisations to the usage of SimpleEventBus

The commit messages should be fairly self explanatory.


astei:
Sorry, things are slow because this is the week before finals. Reviewing now...

PR #165, opened 2019-01-12
Some fixes

xxDark:


xxDark:
> 
> 
> The capacity of the StringBuilders shouldn't be changed, it's there to prevent large unwanted data from being sent to the proxy.

On windows, malloc for example aligns the allocated memory to the size of the memory pag, and Java does the same thing - it reserves some space for realloc, so it makes no sence to set initial capacity for message with unknown length.

proxy/src/main/java/com/velocitypowered/proxy/command/VelocityCommandManager.java
```diff
@@ -13,10 +13,13 @@
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
+import java.util.regex.Pattern;
+
 import org.checkerframework.checker.nullness.qual.NonNull;
 
 public class VelocityCommandManager implements CommandManager {
 
+  private static final Pattern SPLIT_PATTERN = Pattern.compile(" ");
```
Leymooo: Revert this. String#split is faster. 
https://stackoverflow.com/a/14901754
https://github.com/SpigotMC/BungeeCord/pull/2539

lucko: Guava's `Splitter` works too.
https://google.github.io/guava/releases/23.0/api/docs/com/google/common/base/Splitter.html

xxDark: > 
> 
> Guava's `Splitter` works well too.
> https://google.github.io/guava/releases/23.0/api/docs/com/google/common/base/Splitter.html

Gonna make a quick JMH benchmark.

xxDark: > 
> 
> > Guava's `Splitter` works well too.
> > https://google.github.io/guava/releases/23.0/api/docs/com/google/common/base/Splitter.html
> 
> Gonna make a quick JMH benchmark.

Yeah, Guava's `Splitter` is much faster then Java's one.

xxDark:
> 
> 
> The capacity of the StringBuilders shouldn't be changed, it's there to prevent large unwanted data from being sent to the proxy.

Benchmark                    Mode  Cnt         Score        Error  Units
BuilderBenchmark.testDefault  thrpt   25  16446379,016 ± 635051,541  ops/s
BuilderBenchmark.testPredict  thrpt   25   2398699,209 ± 150939,535  ops/s

astei:
Thanks!

xxDark:
@astei please merge my latest commit to fix builds on Jenkis

astei:
I fixed it, don't worry.

PR #166, opened 2019-01-12
Plugin Messages API (Untested)

xxDark:


api/src/main/java/com/velocitypowered/api/proxy/ProxyServer.java
```diff
@@ -19,7 +20,22 @@
 /**
  * Provides an interface to a Minecraft server proxy.
  */
-public interface ProxyServer {
+public abstract class ProxyServer {
```
md678685: Uh, what? You should be able to pass a `ProxyServer` instance wherever you should need it. This doesn't need to be turned into a singleton.

api/src/main/java/com/velocitypowered/api/proxy/ProxyServer.java
```diff
@@ -19,7 +20,26 @@
...
+
+  private static ProxyServer instance;
```
Leymooo: We dont want to use static access to ProxyServer, just pass a instance to VelocityChannelIdentifier.

proxy/src/main/java/com/velocitypowered/proxy/messages/VelocityChannelIdentifier.java
```diff
@@ -0,0 +1,243 @@
...
+
+/**
+ * A dummy {@link ChannelIdentifier} for communication w/ Velocity
+ */
+public final class VelocityChannelIdentifier implements ChannelIdentifier {
+
+    private VelocityChannelIdentifier() { }
+
+    private static final byte[] API_MISMATCH = new byte[]{VelocityReplyCodes.API_MISMATCH};
+    public static final ChannelIdentifier INSTANCE = new VelocityChannelIdentifier();
```
Leymooo: I think its better to create a getter in VelocityServer.

api/src/main/java/com/velocitypowered/api/proxy/ProxyServer.java
xxDark: > 
> 
> Uh, what? You should be able to pass a `ProxyServer` instance wherever you should need it. This doesn't need to be turned into a singleton.

Gonna fix it, when will deal with code style issues :(

xxDark:
P.S Minecraft limits the amount of data being sent in payload packet, that's why I decided to comment the code where I write all player uuids/names.
To bypass this 'restriction (?)', I can make a 'trick' that splits a data in multiple payload packets. 🤔

astei:
@xxDark It wouldn't be hard to just split up the plugin message packets, no? The 32KiB limit is pretty annoying, though.

astei:
Note: we'll probably also need to produce a portable (Bukkit/Sponge) API for this once we're done.

astei:
@xxDark still interested in working on this? I might be willing to take over if you've lost interest.

xxDark:
Yup, I'm a bit busy rn.

PR #169, opened 2019-01-28
register player connection only when switched to PLAY state

Leymooo:
This PR will fix a error caused when plugins try to send a title, message, header and footer, etc during login event. https://pastebin.com/Uwu9Q48f .


proxy/src/main/java/com/velocitypowered/proxy/connection/client/LoginSessionHandler.java
```diff
@@ -223,41 +221,33 @@ private void initializePlayer(GameProfile profile, boolean onlineMode) {
...
-      if (!server.registerConnection(player)) {
+      if (!server.canRegisterConnection(player)) {
         player.disconnect(VelocityMessages.ALREADY_CONNECTED);
         return CompletableFuture.completedFuture(null);
       }
 
+      // Fire DisconnectEvent if player will be disconnected/kicked during login and next events. 
+      // This will protect us from bad plugins, those will store Player object from events
+      // and remove it on disconnect
+      mcConnection.setSessionHandler(new InitialConnectSessionHandler(player));
```
astei: This isn't required, since `LoginSessionHandler` will tear down the player. Look at the implementation of `disconnected()`.

Leymooo: Oh. did not see it

PR #176, opened 2019-03-02
Add matchPlayer and matchServer

Crypnotic:
Would be great to not have to implement these methods in every plugin that might use them. Bungeecord has implemented matchPlayer, but not matchServer. 

Edit: all checkstyle warnings fixed

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
@@ -50,13 +50,16 @@ import java.nio.file.Paths;
 import java.security.KeyPair;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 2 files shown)


proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
```diff
@@ -442,6 +445,38 @@ public void broadcast(Component component) {
...
+    }
+
+    return getAllPlayers().stream().filter(p -> p.getUsername()
+            .regionMatches(true, 0, partialName, 0, partialName.length()))
+            .collect(Collectors.toList());
+  }
+
+  @Override
+  public Collection<RegisteredServer> matchServer(String partialName) {
+    if (partialName == null) {
```
hugmanrique: You can use `Objects.requireNonNull`

proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
```diff
@@ -442,6 +445,38 @@ public void broadcast(Component component) {
     }
   }
 
+  @Override
+  public Collection<Player> matchPlayer(String partialName) {
+    if (partialName == null) {
```
hugmanrique: Same thing herez otherwise the PR looks good to me! 😀

Issue #183, opened 2019-03-16
Much error when use velocity


zhourui123:
forge 1.12.2
spongeforge 7.15
velocity latest
plugin: bungeequack

The following problems are all generated after the use of velocity, So I think this has something to do with velocity.
1. Sometimes connect player from ServerA to ServerB , the ServerA will print "abc moved too quickly! 716.0050468618773,-18.5,-952.4585183223486",then print NPC exception, see gist "server A log" line 129.

2. Server B crash more easily when player connect from Server A.  see gist "server B log". And player will kick back to server A: You were moved from the server you were on because you were kicked

The logs are all at the same time, so they can be viewed in combination

https://gist.github.com/zhourui123/d372e2dff1fee3ffe816826d9373d8ca

astei:
What does "Velocity latest" even mean? Use the build number provided by `/velocity version`.

To me the most interesting (and bizarre parts) are:

```
[19:50:06] [Netty Epoll Server IO #1/INFO] [FML]: Client protocol version 2
[19:50:06] [Netty Epoll Server IO #1/INFO] [FML]: Client attempting to join with 5 mods : minecraft@1.12.2,mfbl@1.0.0.0,FML@8.0.99.99,forge@14.23.5.2768,mcp@9.42
[19:50:06] [Netty Epoll Server IO #1/INFO] [FML]: Attempting connection with missing mods [spongeapi, sponge, spongeforge, common, spongycord, war, mercurius_updater] at CLIENT
[19:50:06] [Netty Epoll Server IO #1/INFO] [FML]: Unexpected packet during modded negotiation - assuming vanilla or keepalives : net.minecraft.network.play.client.CPacketPlayer$Position
[19:50:06] [Netty Epoll Server IO #1/INFO] [FML]: Unexpected packet during modded negotiation - assuming vanilla or keepalives : net.minecraft.network.play.client.CPacketPlayer$Rotation
[19:50:06] [Server thread/WARN] [net.minecraft.network.NetHandlerPlayServer]: abc moved too quickly! 719.4082733512457,-20.5,-950.3579596036736
[19:50:06] [Server thread/FATAL] [net.minecraft.server.MinecraftServer]: Error executing task
java.util.concurrent.ExecutionException: java.lang.NullPointerException
	at java.util.concurrent.FutureTask.report(FutureTask.java:122) ~[?:1.8.0_202]
[...]

PR #191, opened 2019-04-21
An Easter basket bearing gifts!

astei:
It's been a while since I've been tackling major Velocity issues. After much work, I'm finally able to present a nice Easter basket full of gifts for Velocity users:

* Velocity will now wait until `JoinGame` is sent before actually initiating a server switch. (Should fix a number of issues such as #183).
* Velocity is now using Gradle 5!
* A number of nits have been fixed thanks to the switch to Error Prone that accompanied the switch.

Issues left to look at:

* ~~What if Forge connection fails?~~


dualspiral:
> What if Forge connection fails?

Once the Forge handshake has started, the client reverts to a frozen state so as far as it is concerned, it should not be receiving any packets from the old server. It also should not be receiving most packets from the new server until the handshake has completed, such as position and rotation packets. The only safe thing to do at that point is to kick the client if the handshake fails.

astei:
Good to know. I'm wondering if there exists any early cases where Velocity can possibly recover. Can you reset the Forge handshake while it is still handshaking?

dualspiral:
> Can you reset the Forge handshake while it is still handshaking?

No, it only accepts the reset packet during the `DONE` state - https://github.com/MinecraftForge/MinecraftForge/blob/1.12.x/src/main/java/net/minecraftforge/fml/common/network/handshake/FMLHandshakeClientState.java#L200

I am unsure of if there is a way to ask the client if a modlist is good without the handshake. I looked into how the server list works, it's all client side. We may be stuck here, unfortunately.

astei:
Very well then. We can only hope for something better in Forge 1.13.

PR #195, opened 2019-04-27
Fix DataFormatException with java decompression

Willsr71:
In certain situations a client will fail to connect with an exception of `DataFormatException: unknown compression method` or `DataFormatException: incorrect header check`, this fixes that.

This was tested with a Direwolf20 modpack client against a SpongeVanilla server, the unknown compression method error happened with a localhost connection, the incorrect header check happened with the client connecting from outside the network.


astei:
LGTM but you must fix your Checkstyle error first.

PR #210, opened 2019-05-27
1.14.2 support

Leymooo:


api/src/main/java/com/velocitypowered/api/network/ProtocolVersion.java
```diff
@@ -28,8 +28,11 @@
   MINECRAFT_1_13_1(401, "1.13.1"),
   MINECRAFT_1_13_2(404, "1.13.2"),
   MINECRAFT_1_14(477, "1.14"),
-  MINECRAFT_1_14_1(480, "1.14.1");
-  
+  MINECRAFT_1_14_1(480, "1.14.1"),
+  MINECRAFT_1_14_2(485, "1.14.2");
+
+  public static ProtocolVersion[] values = values();
```
Cybermaxke: Make this private, a mutable array should not be exposed in the api.

Cybermaxke: Well, since it's only used to initialize some things. There isn't really a benefit of storing this in a field.

PR #212, opened 2019-05-31
Remove protocol linking

Leymooo:


astei:
Has this been tested? What benefits can we expect from this (i.e. lower memory usage)?

astei:
I see. Much less duplication if a packet ID hasn't changed. I'm fine with this.

Leymooo:
I tested a join from 1.8, 1.12.2 and 1.14.2 before create a pr. If you want i can test it more, but i sure it should be ok.

astei:
I think that's fine. I'll merge this in.

PR #214, opened 2019-06-04
Add ServerChatEvent

MrIvanPlays:
Adds ServerChatEvent. Can be used for manipulating server-sent messages, and (probably) be generated by the proxy with(out) changes

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/event/server/ServerChatEvent.java
+++ b/api/src/main/java/com/velocitypowered/api/event/server/ServerChatEvent.java
@@ -1,0 +1,108 @@
+package com.velocitypowered.api.event.server;
+
+import com.google.common.base.Preconditions;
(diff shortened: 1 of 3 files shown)


astei:
I won't be merging this in. I believe this was influenced by SpigotMC/BungeeCord#2615, and while there are some factors that make this more viable on Velocity (notably immutable chat components), there remain major barriers to actually adding the event.
 
* Chat packets are frequently sent to players from the server. If we're talking to hundreds of players, that's a lot of object allocation (and context-switching as Velocity events fire on an async thread).
* Velocity doesn't deserialize the chat packet by default. In the general, Velocity doesn't (and shouldn't have to) process every packet headed towards the proxy.
* This event isn't particularly user-friendly because the message actually sent to the player's client is a JSON message, requiring an extra cumbersome deserialization (and possible reserialization).

PR #219, opened 2019-06-17
1.7 Protocol Support

Gabik21MC:
This PR implements support for the _evil_ 6 year old minecraft version.
I am down to support this if there are issues.

I tested a join from 1.7, 1.8, 1.13 and 1.14
I tested server switching for 1.7 and 1.8 
But for server switching nothing had to be changed apart from the clearAll for tablist.


api/src/main/java/com/velocitypowered/api/network/ProtocolVersion.java
```diff
@@ -13,6 +13,7 @@
 public enum ProtocolVersion {
   UNKNOWN(-1, "Unknown"),
   LEGACY(-2, "Legacy"),
+  MINECRAFT_1_7(5, "1.7"),
```
Leymooo: 1.7 - 1.7.10 have multiple protocol versions (1.7-1.7.5, 1.7.6-1.7.10 if i not mistaken)

proxy/src/main/java/com/velocitypowered/proxy/protocol/ProtocolUtils.java
```diff
@@ -194,12 +203,95 @@ public static void writeProperties(ByteBuf buf, List<GameProfile.Property> prope
...
+
+    byte[] ret = new byte[len];
+    buf.readBytes(ret);
+    return ret;
+  }
+
+  // Legacy 1.7
+
+  /**
+   * Legacy.
```
hugmanrique: Please add meaningful javadocs or don't add them.

astei: Checkstyle requires Javadocs for non-trivial methods.

proxy/src/main/java/com/velocitypowered/proxy/protocol/ProtocolUtils.java
```diff
@@ -194,12 +203,95 @@ public static void writeProperties(ByteBuf buf, List<GameProfile.Property> prope
...
+   * @param b array
+   * @param buf buf
+   * @param allowExtended forge
+   */
+  public static void writeArray17(byte[] b, ByteBuf buf, boolean allowExtended) {
+    // (Integer.MAX_VALUE & 0x1FFF9A ) = 2097050 - Forge's current upper limit
+    if (allowExtended) {
+      Preconditions
+          .checkArgument(b.length <= (Integer.MAX_VALUE & 0x1FFF9A),
```
Cybermaxke: Put `Integer.MAX_VALUE & 0x1FFF9A` in a constant value, and use it for the message.

proxy/src/main/java/com/velocitypowered/proxy/tablist/VelocityTabListLegacy.java
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  @Override
+  public void clearHeaderAndFooter() {
+  }
+
+  @Override
+  public void addEntry(TabListEntry entry) {
+    super.addEntry(entry);
+    nameMapping.put(entry.getProfile().getName(), entry.getProfile().getId());
```
Cybermaxke: Maybe the display name should be used instead, the display name is what is actually should be displayed on the client. This also supports (legacy) color formatting.

Gabik21MC: The display name is a textcomponent and may be null. It is also transferred as a json, so I thought it would be better to fall back to non colorized and guarantee compatibility setting the displayname from the backend server.

Gabik21MC:
Should I remove the unnessecary diff caused by formatting?

proxy/src/main/java/com/velocitypowered/proxy/tablist/VelocityTabList.java
```diff
@@ -20,8 +21,8 @@
 
 public class VelocityTabList implements TabList {
 
-  private final MinecraftConnection connection;
```
Leymooo: protected?

proxy/src/main/java/com/velocitypowered/proxy/tablist/VelocityTabListLegacy.java
Cybermaxke: Then use the display name if present (converted to legacy formatting, with 16 char limit), and fall back to the profile name. That still guarantees the compatibility.

dualspiral:
Has this been tested with Forge for 1.7.10? It might seem a daft question, but I would imagine that the vast majority of 1.7.x servers will be forge based.

The reason I ask is that when I added Forge support to Velocity, [I made the assumption that we can detect that Forge is running on the client in 1.8-1.12.](https://github.com/VelocityPowered/Velocity/blob/master/proxy/src/main/java/com/velocitypowered/proxy/connection/client/HandshakeSessionHandler.java#L114). Forge in 1.8+ sends a `%FML%` marker in the client->server handshake to indicate that the connecting client is a Forge client.

However, **this is not the case in 1.7.10**. A forge client will simply dive into the handshake when server login completes with no other warning, this turned out to be a bit of a headache when I was looking at how to get this on BungeeCord - we were having all sorts of race conditions. I don't remember exactly what we did, but looking through the old Bungee patch, it seems to be us listening for a REGISTER packet for the "FML|HS" channel that alerted us to it, which is well past the login sequence anyway. 

It looks to me that you haven't detected this, and so Velocity won't know to send a reset packet to the client when switching servers, which I think will mean that the switch will hang (client not sending a hello back to the server on server switch) and the switch will fail.

(Also, don't bother testing this on any Forge version for 1.7.2, it's missing the reset packet meaning that server switching won't work.)

Gabik21MC:
I'll look into that finding a solution, might take a few days. Haven't really been dealing with Forge on the server side, so I have to get into this. Thank you for the detailed description of the problem :)

proxy/src/main/java/com/velocitypowered/proxy/protocol/ProtocolUtils.java
```diff
@@ -194,12 +203,93 @@ public static void writeProperties(ByteBuf buf, List<GameProfile.Property> prope
...
+    buf.writeBytes(b);
+  }
+
+  /**
+   * Reads a Minecraft-style VarInt from the specified {@code buf}.
+   *
+   * @param buf buf to write
+   * @return read int
+   */
+  public static int readVarShort(ByteBuf buf) {
```
Cybermaxke: This isn't really a `VarShort`, since it can encode higher numbers than a short. This should get a different name, maybe `readExtendedShort` (reflects the javadoc of forge 1.7.10), or `readExtendedForgeShort` (because it's only used by forge anyway).
The same applies to `writeVarShort`.

proxy/src/main/java/com/velocitypowered/proxy/protocol/ProtocolUtils.java
```diff
@@ -194,12 +203,93 @@ public static void writeProperties(ByteBuf buf, List<GameProfile.Property> prope
...
 
+  private static final int FORGE_MAX_ARRAY_LENGTH = Integer.MAX_VALUE & 0x1FFF9A;
+
+  /**
+   * Reads an byte array for legacy version 1.7 from the specified {@code buf}
+   *
+   * @param buf the buffer to read from
+   * @return the read byte array
+   */
+  public static byte[] readArray17(ByteBuf buf) {
```
Cybermaxke: `readByteArray17` to be consistent with `readByteArray`

proxy/src/main/java/com/velocitypowered/proxy/tablist/VelocityTabListLegacy.java
```diff
@@ -0,0 +1,109 @@
...
+            entry.setLatency(item.getLatency());
+          }
+        } else {
+          UUID uuid = UUID.randomUUID(); // Use a fake uuid to preserve function of custom entries
+          nameMapping.put(item.getName(), uuid);
+          entries.put(uuid, (VelocityTabListEntry) TabListEntry.builder()
+              .tabList(this)
+              .profile(new GameProfile(uuid, item.getName(), ImmutableList.of()))
+              .latency(item.getLatency())
+              .build());
```
Cybermaxke: Since the names can contain legacy formatting characters. If one is present, the legacy name should be converted to a component as the display name, and the normal (profile) name should (maybe) be stripped of these characters. If no formatting characters are present, no display name.
Always providing a display name is also an option.

proxy/src/main/java/com/velocitypowered/proxy/connection/backend/BackendPlaySessionHandler.java
```diff
@@ -103,8 +103,12 @@ public boolean handle(PluginMessage packet) {
     }
 
     if (PluginMessageUtil.isMcBrand(packet)) {
-      PluginMessage rewritten = PluginMessageUtil.rewriteMinecraftBrand(packet,
-          server.getVersion());
+      PluginMessage rewritten;
+      if (playerConnection.getProtocolVersion().compareTo(ProtocolVersion.MINECRAFT_1_8) >= 0) {
+        rewritten = PluginMessageUtil.rewriteMinecraftBrand(packet, server.getVersion());
```
astei: Could we just add a `ProtocolVersion` parameter for `rewriteMinecraftBrand` instead?

proxy/src/main/java/com/velocitypowered/proxy/tablist/VelocityTabListLegacy.java
Gabik21MC: Is there already a method somewhere to strip colors?

astei: You can use the `PlainComponentSerializer`.

Gabik21MC:
Alright so connecting to a forge server works fine, but switching to a forge server (does not matter if switching from vanilla or forge) causes the connection to be in transition state forever and I have no clue why

astei:
https://github.com/VelocityPowered/Velocity/pull/219#issuecomment-502747488 should give you a clue as to why.

proxy/src/main/java/com/velocitypowered/proxy/tablist/VelocityTabListLegacy.java
```diff
@@ -0,0 +1,113 @@
...
+          if (entry != null) {
+            entry.setLatency(item.getLatency());
+          }
+        } else {
+          UUID uuid = UUID.randomUUID(); // Use a fake uuid to preserve function of custom entries
+          nameMapping.put(strippedName, uuid);
+          entries.put(uuid, (VelocityTabListEntry) TabListEntry.builder()
+              .tabList(this)
+              .profile(new GameProfile(uuid, strippedName, ImmutableList.of()))
+              .latency(item.getLatency())
```
Cybermaxke: The display name still needs to be applied here, which is the text component returned by the legacy component serializer. It's only necessary to apply this if the stripped name and the original don't match (so contains formatting codes).

TerriblePanda:
hope to merga it soon.

Gabik21MC:
Forge is now working, but the first connecting server must be a forge server as explained in the commit message.

proxy/src/main/java/com/velocitypowered/proxy/connection/client/HandshakeSessionHandler.java
```diff
@@ -114,6 +114,10 @@ private ConnectionType checkForForge(Handshake handshake) {
     if (handshake.getServerAddress().endsWith(LegacyForgeConstants.HANDSHAKE_HOSTNAME_TOKEN)
         && handshake.getProtocolVersion().compareTo(ProtocolVersion.MINECRAFT_1_13) < 0) {
       return ConnectionTypes.LEGACY_FORGE;
+    } else if (handshake.getProtocolVersion().compareTo(ProtocolVersion.MINECRAFT_1_7_6) <= 0) {
+      // 1.7 Forge will not notify us during handshake. UNDETERMINED will listen for incoming
+      // forge handshake attempts. Vanilla clients will remain UNDETERMINED.
```
dualspiral: You should set the connection type as `VANILLA` as soon as you are confident that this is the case, rather than leaving this as undetermined.

dualspiral:
> but the first connecting server must be a forge server as explained in the commit message

I was able to do it with a vanilla server just fine on Bungee. I think all you need to do is check for the register `FML|HS` packet from the server (or even the first hello, server sends a message first) and, if this is a server switch rather than first server, send the reset packet.

To that end, maybe create a `VANILLA_17` connection type that deals with this behaviour for 1.7.10 clients only. It's suboptimal that a recieving server has to be a Forge server simply because many networks will have vanilla lobbies so they can use their Bukkit 1.7/Spigot plugins.

proxy/src/main/java/com/velocitypowered/proxy/connection/client/HandshakeSessionHandler.java
Gabik21MC: Whoops that comment is wrong. It is set to `VANILLA` after the `JoinGame` packet

astei:
As this PR adds a new non-patch version of Minecraft, it will need to be a part of Velocity 1.1.0. If you can, please rebase onto `dev/1.1.0` so it can be a part of Velocity 1.1.0.

astei:
Please make sure your merge conflicts are fixed. (I was working on some plugin messaging stuff earlier.)

Gabik21MC:
I am unsure if this is a correct solution. I can only confirm that it's working.

NycuRO:
@Gabik21MC hey

Gabik21MC:
Just for clarification: Vanilla (1.7.2-1.7.10) and Forge (1.7.10) is working now and Forge also works with the first server being Vanilla.

mpztilera:
Thats a great change to Velocity, if this would be merged I would switch to Velocity, because I need 1.7.10 support

astei:
Merging this in now - thank you!

PR #228, opened 2019-07-02
Fix NPE's when a command isn't known by velocity.

Cybermaxke:
Only affects `dev/1.1.0` since https://github.com/VelocityPowered/Velocity/commit/7834acd67f84e83834a8888f81e570cc435d2e98


astei:
Fixed in 39c505f

PR #229, opened 2019-07-02
Wait for player disconnect events on shutdown.

Cybermaxke:
Fixes https://github.com/VelocityPowered/Velocity/issues/178

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
@@ -57,6 +57,7 @@ import java.util.Optional;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ThreadLocalRandom;
 import net.kyori.text.Component;
 import net.kyori.text.TextComponent;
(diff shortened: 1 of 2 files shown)


proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
```diff
@@ -553,7 +555,15 @@ void teardown() {
...
     }
     server.unregisterConnection(this);
-    server.getEventManager().fireAndForget(new DisconnectEvent(this));
+    server.getEventManager().fire(new DisconnectEvent(this))
+            .thenRun(() -> this.teardownFuture.complete(null));
+  }
+
+  public void awaitTeardown() {
+    try {
+      teardownFuture.get();
```
Cybermaxke: Maybe a timeout? In case a plugin takes too much time to handle the event.

astei: There should be a timeout. Further, it needs to be over all player disconnects, not just individual player disconnects.

astei:
Please make sure to fix your Checkstyle issues.

Gabik21MC:
I have some questions regarding this. `ConnectionManager#shutdown` closes the endpoints and is called after the disconnect logic. Shouldn't this be done beforehand?

Also the boss and worker event loop groups aren't shut down at any point or am I wrong?

Cybermaxke:
@Gabik21MC You can't send disconnect packets if the endpoint is already closed.

astei:
This looks good to me. I would probably also add an explicit Netty group teardown but I can do that myself. Thanks!

(This has been a bug for a while, so I'll likely backport it to 1.0.2.)

PR #237, opened 2019-07-21
Update TerminalConsoleAppender to 1.2.0

Minecrell:
Update to the proper release. Seems like you have already switched to Jansi :)


astei:
Yep! It will help us because our JAR sizes have bloated for 1.1.0 because we're now shading Configurate.

PR #239, opened 2019-08-06
Show correct modinfo on server ping

alexstaeding:
This PR adds support for retrieving the `ModInfo` part of a ping from a specified server to forward it to the client via the proxy. This solves an issue with modded servers having a red X shown along with "Incompatible: 0 mods present" instead of the correct amount.

PR #240, opened 2019-08-06
Ping passthrough

astei:
A more general (and flexible) version of #239.

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/proxy/server/ServerPing.java
+++ b/api/src/main/java/com/velocitypowered/api/proxy/server/ServerPing.java
@@ -182,6 +182,19 @@ public final class ServerPing {
       return this;
     }
 
+    /**
+     * Uses the modified {@code mods} list in the response.
+     * @param mods the mods list to use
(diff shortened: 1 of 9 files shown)


astei:
Tested with passthrough disabled and with forwarding all information. Mod-only forwarding will be tested as soon as I'm back in college.

Leymooo:
Why ping all servers instead of first server in list like in bungee?

astei:
Because there is no reason not to parallelize the work in question. This will speed up the response to the user if a server is slow in responding to the request by allowing all the other requests to go through.

This is also required for the mods-only pass-through to work: it has to find the first server with a mod list it can pass to the client.

LGCMcLovin:
what if first server is a lobby with no mods, and you want a specific set of servers to be checked?

astei:
It will try to pick the first mod list it finds. So if you have a vanilla server first, it will move onto the next item in the list, because it has no mod list.

astei:
Mod list forwarding works. I am merging this in.

PR #254, opened 2019-10-20
Make AES crypto operations no-copy

Ichbinjoe:
All AES implementations being used are 'copy safe', where the source and
destination arrays may be the same. Lets save ourself a copy and reap
the performance wins!

### DIFF
--- a/native/src/main/java/com/velocitypowered/natives/encryption/JavaVelocityCipher.java
+++ b/native/src/main/java/com/velocitypowered/natives/encryption/JavaVelocityCipher.java
@@ -35,48 +35,22 @@ public class JavaVelocityCipher implements VelocityCipher {
   }
 
   @Override
-  public void process(ByteBuf source, ByteBuf destination) throws ShortBufferException {
+  public void process(ByteBuf source) {
     ensureNotDisposed();
 
     int inBytes = source.readableBytes();
     byte[] asBytes = ByteBufUtil.getBytes(source);
 
-    int outputSize = cipher.getOutputSize(inBytes);
-    byte[] outBuf = new byte[outputSize];
-    cipher.update(asBytes, 0, inBytes, outBuf);
-    destination.writeBytes(outBuf);
-  }
-
-  @Override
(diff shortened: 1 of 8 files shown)


native/src/test/java/com/velocitypowered/natives/encryption/VelocityCipherTest.java
```diff
@@ -56,20 +56,18 @@ private void check(VelocityCipherFactory factory, Supplier<ByteBuf> bufSupplier)
     VelocityCipher encrypt = factory.forEncryption(new SecretKeySpec(AES_KEY, "AES"));
 
     ByteBuf source = bufSupplier.get();
-    ByteBuf dest = bufSupplier.get();
-    ByteBuf decryptionBuf = bufSupplier.get();
 
     source.writeBytes(TEST_DATA);
 
+    ByteBuf working_buf = source.copy();
```
astei: Checkstyle error. Perhaps `tmpBuf` would be better here.

native/src/main/java/com/velocitypowered/natives/encryption/JavaVelocityCipher.java
```diff
@@ -35,50 +35,24 @@ private JavaVelocityCipher(boolean encrypt, SecretKey key) throws GeneralSecurit
...
-      throw e;
-    } finally {
-      asHeapBuf.release();
+      cipher.update(asBytes, 0, inBytes, asBytes);
+    } catch (ShortBufferException ex) {
+      /* This _really_ shouldn't happen - AES CFB8 will work in place.
+         If you run into this, that means that for whatever reason the Java Runtime has determined
+         that the output buffer needs more bytes than the input buffer. When we are working with
+         AES-CFB8, the output size is equal to the input size. See the problem? */
+      throw new RuntimeException("Cipher update did not operate in place and requested a larger "
```
astei: I'd probably make this an `AssertionError`

native/src/test/java/com/velocitypowered/natives/encryption/VelocityCipherTest.java
Ichbinjoe: workingBuf is more descriptive.

PR #255, opened 2019-10-20
Integrate BungeeQuack functionality into Velocity

### LINKED ISSUE #247: Integrate BungeeCord plugin messaging channel support (BungeeQuack) into Velocity directly
This effectively closes #68.

In the absence of work on a native Velocity plugin communication API, I think we should strongly consider integrating [BungeeQuack](https://github.com/VelocityPowered/BungeeQuack) directly into the Velocity core.

The main advantage of doing this is that users get support with the majority of BungeeCord-Bukkit plugins out there without requiring more drastic work on our end (to, say, implement a compatibility layer for the BungeeCord plugin API).

BungeeQuack is already one of the most popular Velocity plugins out there and provides an important asset to help users move from BungeeCord to Velocity. It makes sense for us to ease the transition as much as possible.

astei:
Fixes #247.

This code needs to be cleaned up, so I'm not merging it right away.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/backend/BungeeCordMessageResponder.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/backend/BungeeCordMessageResponder.java
@@ -53,7 +53,7 @@ class BungeeCordMessageResponder {
     out.writeUTF("IP");
     out.writeUTF(player.getRemoteAddress().getHostString());
     out.writeInt(player.getRemoteAddress().getPort());
-    sendResponse(buf);
+    sendResponseOnConnection(buf);
   }
 
   private void processPlayerCount(ByteBufDataInput in) {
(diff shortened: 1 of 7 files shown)


proxy/src/main/java/com/velocitypowered/proxy/connection/backend/BungeeCordMessageResponder.java
```diff
@@ -0,0 +1,325 @@
...
+
+  public boolean process(PluginMessage message) {
+    if (!MODERN_CHANNEL.getId().equals(message.getChannel()) && !LEGACY_CHANNEL.getId()
+        .equals(message.getChannel())) {
+      return false;
+    }
+
+    ByteBufDataInput in = new ByteBufDataInput(message.content());
+    String subChannel = in.readUTF();
+    if (subChannel.equals("ForwardToPlayer")) {
```
hugmanrique: A `switch` would be better here

astei: The subchannel handling is getting split up anyway, so this is effectively redundant.

PR #256, opened 2019-10-28
Add previous server to ServerConnectedEvent

Gabik21:


api/src/main/java/com/velocitypowered/api/event/player/ServerConnectedEvent.java
```diff
@@ -26,6 +31,10 @@ public RegisteredServer getServer() {
     return server;
   }
 
+  public Optional<RegisteredServer> getPreviousServer() {
+    return Optional.ofNullable(previousServer);
+  }
+
   @Override
   public String toString() {
```
md678685: I believe `toString` should include `previousServer`.

api/src/main/java/com/velocitypowered/api/event/player/ServerConnectedEvent.java
```diff
@@ -12,10 +14,19 @@
 
   private final Player player;
   private final RegisteredServer server;
+  private final @Nullable RegisteredServer previousServer;
 
-  public ServerConnectedEvent(Player player, RegisteredServer server) {
```
astei: API breakage. Readd this constructor and simply forward to the new one.

Gabik21: Should I mark it as deprecated?

astei: That's fine.

PR #268, opened 2020-01-19
Added initial framework for service API. Resolves #231

### LINKED ISSUE #231: Services
Integrating services into Velocity will afford us platform-independence from specific plugins.

- [ ] Services Manager
- [ ] Service interfaces
  - [ ] Permissions
  - Others?

Crypnotic:
This commit adds a barebones Service API to allow a discussion based solely on the framework itself before any service interfaces are created.

Key notes:

- Currently services can only be provided by 1 plugin at a time. Bukkit allows services to be registered with priority. This may be something we should look into, but for the time being 1:1 is what is implemented.
- This is by no means aimed to be a finished product.
- This is loosely based on what Bukkit has implemented currently such that any interface can be registered by any plugin.
- Currently, services cannot be unregistered by either the proxy or any plugins
- There is very little by way of querying the registered services


STG-Allen:
I'm fond of the ability to create services, But I feel in order for this to be implemented we may need to be able to unregister the services. Along with that maybe some basic services like Permissions, User, and Eco. I can look into making a pr to implement these services if needed.

Crypnotic:
That's planned, but with that comes with trying to figure out where within the proxy we need to unregister which I was not ready to dive into. Since it's not crucial at this point, I PRed without

md678685:
> with that comes with trying to figure out where within the proxy we need to unregister

Out of curiosity, why would the proxy need to be able to unregister a service? In its current form, the PR doesn't allow the proxy to register a service (only plugins can), so it would only be able to unregister services that were registered by plugins.

Crypnotic:
> > with that comes with trying to figure out where within the proxy we need to unregister
> 
> Out of curiosity, why would the proxy need to be able to unregister a service? In its current form, the PR doesn't allow the proxy to register a service (only plugins can), so it would only be able to unregister services that were registered by plugins.

I'm not entirely sure that it would need to, but that discussion needs to be had. Which is why it was left out for now

Crypnotic:
We also need to decide the best way to store a plugin -> services map since there's no way to get that information without stepping through the services map

Crypnotic:
> > plugin -> services map since there's no way to get that information without stepping through the services map
> 
> Does it really matter? I don't really see there being any case where such a thing is going to be checked often or in any context where performance is super relevant?

I don't necessarily think it is a great idea, but I can see someone coming into discord in due time asking how to do so. Really just a toss up between exposing the service map so they can step through it themselves, implementing such a method, or linking plugin -> service

proxy/src/main/java/com/velocitypowered/proxy/service/VelocityServiceProvider.java
```diff
@@ -0,0 +1,35 @@
...
+   * @param serviceClass the provided service
+   * @param provider the service provider
+   */
+  VelocityServiceProvider(Object plugin, Class<S> serviceClass, S provider) {
+    this.plugin = plugin;
+    this.serviceClass = serviceClass;
+    this.provider = provider;
+  }
+
+  public Object getPlugin() {
```
kashike, line 25: `@Override`

proxy/src/main/java/com/velocitypowered/proxy/service/VelocityServiceProvider.java
```diff
@@ -0,0 +1,35 @@
...
+    return plugin;
+  }
+
+  public Class<S> getService() {
```
kashike, line 30: `@Override`

proxy/src/main/java/com/velocitypowered/proxy/service/VelocityServiceProvider.java
```diff
@@ -0,0 +1,35 @@
...
+    return serviceClass;
+  }
+
+  public S getProvider() {
```
kashike, line 35: `@Override`

Crypnotic:
Resolved the overrides. Not sure why the ProtocolVersion 1.15.2 updates are showing now, but will sort out later

MrIvanPlays:
any progress on this?

astei:
Plenty of signs this is still a WIP. What more updates could you want?

MrIvanPlays:
> Plenty of signs this is still a WIP. What more updates could you want?

I mean, haven't been updated for like 7 months now and i think it's a dead wip?

Crypnotic:
> 
> 
> > Plenty of signs this is still a WIP. What more updates could you want?
> 
> I mean, haven't been updated for like 7 months now and i think it's a dead wip?
I left this as WIP for input from anyone that wanted changes made before it was merged, but it appears it's fine in its current form. RTM after review.

proxy/src/main/java/com/velocitypowered/proxy/service/VelocityServiceManager.java
```diff
@@ -0,0 +1,60 @@
...
+
+  @Override
+  public <S> boolean register(Object plugin, Class<S> serviceClass, S provider) {
+    checkNotNull(plugin, "plugin");
+    checkNotNull(serviceClass, "serviceClass");
+    checkNotNull(provider, "provider");
+    checkArgument(pluginManager.fromInstance(plugin).isPresent(), "plugin is not registered");
+    if (!registeredServices.containsKey(serviceClass)) {
+      registeredServices
+              .put(serviceClass, new VelocityServiceProvider<>(plugin, serviceClass, provider));
```
A248: You can use putIfAbsent for an atomic put operation, which will make the return value (very slightly) more reliable

PR #269, opened 2020-01-24
Optimize tablist a bit and remove respawn packet

xxDark:
Removing respawn packet was tested on 1.12.2 vanilla & FML. If any issues occur on newer versions - feel free to comment.


xxDark:
An advantage of removing respawn packet is the speed of server switching.
Due to how Mojang made all minecraft versions before 1.13.2 (I tihnk), every time the client receives JoinGame and Respawn packet, `System.gc()` is invoked, causing JVM to process garbage collection. A big pause happens, client continues. 
```Java
        else
        {
            this.saveLoader.flushCache();
            this.player = null;
        }

        System.gc();
        this.systemTime = 0L;
```
Removing `System.gc()` will give huge world switch boost, but unfortunately, Mojang does not backport fixes.

astei:
> `System.gc()` is invoked, causing JVM to process garbage collection
👀 

I believe I had problems with sending just the JoinGame packet, but I'll make some time to test this.

astei:
Just tried this with vanilla 1.12-1.15. Server switching freezes the client.

MrIvanPlays:
You can't just remove the respawn packet and expect it to work on newest versions. If it works on FML, we can add a exception and don't send it if it is a Forge client /shrug

astei:
Closing due to inactivity

PR #272, opened 2020-02-05
Make PluginContainer injectable

alexstaeding:
This PR makes `PluginContainer` injectable in a plugin.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/plugin/VelocityPluginManager.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/plugin/VelocityPluginManager.java
@@ -4,11 +4,18 @@ import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.google.common.base.Joiner;
+import com.google.inject.AbstractModule;
+import com.google.inject.Module;
+import com.google.inject.name.Names;
+import com.velocitypowered.api.command.CommandManager;
+import com.velocitypowered.api.event.EventManager;
 import com.velocitypowered.api.plugin.PluginContainer;
 import com.velocitypowered.api.plugin.PluginDescription;
 import com.velocitypowered.api.plugin.PluginManager;
 import com.velocitypowered.api.plugin.meta.PluginDependency;
+import com.velocitypowered.api.proxy.ProxyServer;
 import com.velocitypowered.proxy.VelocityServer;
+import com.velocitypowered.proxy.plugin.loader.VelocityPluginContainer;
 import com.velocitypowered.proxy.plugin.loader.java.JavaPluginLoader;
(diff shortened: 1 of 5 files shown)


Crypnotic:
I really only see this being practical if we could adapt this so that plugins could inject instances of other plugins that they depend on. That would allow inter-plugin loading more stream-lined. Other than that, this is information that can be gathered by other means for the few people who actually need it

alexstaeding:
The latest commit 0637f6229dbed0332d8bfe49afa48c4a8d856c95 lets plugins inject the `PluginContainer` from any other loaded plugin by writing:
```java
@Inject
@Named("<pluginid>")
PluginContainer pluginContainer;
```
And if they want to inject their own `PluginContainer`:
```java
@Inject
PluginContainer pluginContainer;
```

proxy/src/main/java/com/velocitypowered/proxy/plugin/loader/PluginLoader.java
```diff
@@ -11,5 +13,30 @@
...
-  PluginContainer createPlugin(PluginDescription plugin) throws Exception;
+  /**
+   * Creates a {@link Module} for the provided {@link PluginContainer}
+   * and verifies that the container's {@link PluginDescription} is correct.
+   *
+   * <p>Does not create an instance of the plugin.</p>
+   *
+   * @param container the plugin container
+   * @return the module containing bindings specific to this plugin
+   * @throws Exception If anything went wrong
```
astei: Can we be more specific here? This bit of JavaDoc adds no information.

proxy/src/main/java/com/velocitypowered/proxy/plugin/VelocityPluginManager.java
```diff
@@ -92,19 +100,44 @@ public void loadPlugins(Path directory) throws IOException {
...
+    // Make a global Guice module that with common bindings for every plugin
+    AbstractModule commonModule = new AbstractModule() {
+      @Override
+      protected void configure() {
+        bind(ProxyServer.class).toInstance(server);
+        bind(PluginManager.class).toInstance(server.getPluginManager());
+        bind(EventManager.class).toInstance(server.getEventManager());
+        bind(CommandManager.class).toInstance(server.getCommandManager());
+        for (PluginContainer container : pluginContainers.keySet()) {
+          bind(PluginContainer.class)
```
astei: If you try to inject a `PluginContainer` for some other plugin but that plugin doesn't exist, Guice will refuse to initialize the object for the plugin. Given that Velocity also supports optional dependencies, this isn't some strange thing that will never happen - it is a possibility.

alexstaeding: Guice supports optional dependencies with
```java
@Inject(optional = true)
```

PR #274, opened 2020-02-09
Clean server name keys before attempting to register

Crypnotic:
Server names that don't match [A-Za-z0-9_-] fail to parse unless wrapped in quotes. The TOML api leaves these quotes in place when reading the key which leads to unintended name behavior.


astei:
This sounds like a bug in the [upstream TOML parser](https://github.com/mwanji/toml4j), and Velocity shouldn't need to work around this bug. Unfortunately, the TOML parser is also not really maintained any more.

Let me think some more about this.

Crypnotic:
Fair point, since the parser for the most part works as-is. Would it be an option to fork, commit, and host under the velocity repo? The only options I see are to switch parsers, fork, or handle any bugs on our end.

StealWonders:
https://github.com/SpongePowered/configurate

mikroskeem:
No TOML support there, or why did you put this link here?

StealWonders:
I thought it did, sorry my bad.

astei:
For now I am going to merge this in.

astei:
A permanent fix for this is now in place for the next version of Velocity. The configuration library has been changed to night-config, which is actively maintained (compared to toml4j).

PR #279, opened 2020-03-13
Remove tasks from lookup map upon completion

hugmanrique:


proxy/src/main/java/com/velocitypowered/proxy/scheduler/VelocityScheduler.java
```diff
@@ -187,6 +187,7 @@ public void run() {
         } catch (Exception e) {
           Log.logger.error("Exception in task {} by plugin {}", runnable, plugin, e);
         } finally {
+          onFinish();
```
Leymooo: i think you should call onFinish() only if task is **not** repeatable

hugmanrique: You're absolute right, I missed it :)

astei:
Good catch, thanks

PR #287, opened 2020-04-06
Fire DisconnectEvent in ConnectedPlayer#disconnect

alex11441:
I noticed that in some cases (see: proxy shutdown), DisconnectEvent will never be called for a player whenever they're disconnected by the proxy.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
@@ -279,6 +279,9 @@ public class ConnectedPlayer implements MinecraftConnectionAssociation, Player {
 
   @Override
   public void disconnect(Component reason) {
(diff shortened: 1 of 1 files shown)


Gabik21:
I don't think your claim is correct. The `DisconnectEvent` is called when the `disconnected` method is invoked by netty, which calls the `teardown` method on the player object.
Also I just tried shutting down my test setup and it did fire the `DisconnectEvent`.
Maybe I am wrong here, do you have any code that reproduces your problem?

astei:
This introduces yet another issue: `DisconnectEvent` can now be fired twice, once from `disconnect()` and once again when the connection actually closes.

I believe the issue was supposed to be fixed in 145dfa8ac6e79 (back in July 2019). Can you please verify with the latest 1.1.0 development version?

alex11441:
appears it works on build 80, my bad oops

alex11441:
I'll do some more testing (who would've figured my 10 seconds worth possibly wasn't enough) and edit this reply

PR #288, opened 2020-04-07
Fix tablist behaving wrong with colors (1.7) (Fixes #282)

### LINKED ISSUE #282: 1.7 tablist issue
Velocity version: Velocity 1.1.0-SNAPSHOT (git-c54ea620-b79)

Minecraft version: 1.7.10

What is wrong: Server tablist is wrong.

Expected result: [here](https://imgur.com/a/aVWI8EW)

Actual result: [here](https://imgur.com/a/uVWidvg)

It works perfectly both in vanilla and bungeecord.
I'm adding fake players to the tablist, after they are added, the player itself is removed from the tablist.
These fake players can be added to teams, which can then have their prefixes and suffixes modified to have animated text in the tablist on 1.7.

It only seems to happen when the names have color codes in them (??????????).

Gabik21:
Fixes #282
Text below is copied from the commit message:

> The 1.7 tablist packet only contains three types of information:
> 
>  - Name of the tablist entry (limited to 16 characters including colors)
>  - Ping of the entry
>  - If this entry needs to be added or removed (client accepts duplicates
>  as 'ping update')
> 
> The previous logic was trying to preserve parity with
> GameProfile#getName returning a stripped down name to have a 'real'
> username.
> That is fundamentally broken, because entries with duplicate content,
> but different colors are very common, especially with custom tablists.
> 
> For packets coming from a native 1.7 server we just won't define the
> displayname anymore, as there is no such thing as a 'displayname',
> because tablist entries are not bound to any player.
> 
> Using the Velocity Tablist API to modify existing entries will work, though
> the backend server will completely loose control over the entry. Custom
> entries added over the Velocity Tablist API will work, but are cut off
> by the 16 character limitation.
> 
> This commit only fixes the bug, where entries are incorrectly handled
> with their stripped name, a lot of the things explained above were
> already implemented correctly.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/tablist/VelocityTabListLegacy.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/tablist/VelocityTabListLegacy.java
@@ -60,29 +60,25 @@ public class VelocityTabListLegacy extends VelocityTabList {
 
     Item item = packet.getItems().get(0); // Only one item per packet in 1.7
 
-    Component displayName = LegacyComponentSerializer.legacy().deserialize(item.getName());
-    String strippedName = PlainComponentSerializer.INSTANCE.serialize(displayName);
-
     switch (packet.getAction()) {
       case PlayerListItem.ADD_PLAYER:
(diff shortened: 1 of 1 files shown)


astei:
Thanks.

PR #290, opened 2020-04-07
Refer to VelocityServer as proxy

### LINKED ISSUE #245: Always refer to the VelocityServer instance as the proxy, not as a server
This is one of the bigger internals-breaking changes slated for Velocity 1.1.0, yet one of the more easily-managed one.

Currently, the `VelocityServer` instance in the proxy is usually referred to as `server`. However, we also allow players to connect to `server`s, which makes for some confusing name shadowing. We should always refer to the Velocity server as the proxy.

Gabik21:
Closes #245
IDE also removed some unused imports automagically.
_Damn this felt productive but changes nothing_


astei:
Closing as this is not a priority for us

PR #291, opened 2020-04-07
Per server IP & profile forwarding configuration

mikroskeem:
I heard that this feature is requested a lot, besides it's a feature that I need myself.

Use cases of this feature:
1) You have servers with different versions in your network - some are as old as 1.7.10, some are 1.15.2. Sadly you'd have to use legacy forwarding then, instead of modern.
2) You use different server software in your network - some servers run Paper, some run 1000$ shady MC-Market fork which doesn't support Velocity forwarding mode. Also forced to use legacy forwarding then.
3) You don't like having huge Handshake packet.

This PR attempts to solve this issue by adding configuration option per server which allows you to configure what IP & profile forwarding should be used.

Will very likely involve changing configuration file structure sadly, but it's for the greater good.

PR is marked draft, code will be pushed as I'll work on implementing this into my network.


mikroskeem:
Will be rebased on top of Velocity 2.x.x because of needed API changes.

mikroskeem:
I'm not interested in continuing it this time. Someone else may pick this up.

PR #299, opened 2020-04-26
Add event for commands

Leymooo:


astei:
This is extremely close to the story I want to write for `CommandExecuteEvent`. I have some concerns about this upon initial glance, I'll write more in a formal review.

api/src/main/java/com/velocitypowered/api/command/CommandManager.java
```diff
@@ -35,10 +35,21 @@
...
+
+  /**
+   * Attempts to execute a command from the specified {@code cmdLine}.
+   *
+   * @param source the command's source
+   * @param cmdLine the command to run
+   * @param callEvent will CommandExecuteEvent called or not
+   * @return true if the command was found and executed, false if it was not
+   */
+  boolean execute(CommandSource source, String cmdLine, boolean callEvent);
```
astei, line 54: This is the biggest problem I have with this. I am not a big fan of exposing a method like this. Could we do a bit better?

On an additional note, I think we should probably deprecate `execute` and friends and just outright replace them with `*async` variants. We'll clean up the API mess later on.

Leymooo: I think its not problem, because execute will be called from event/executor thread pool in 99% cases. I think we should not messy this with async stuff.

Leymooo:
Maybe something like this?

api/src/main/java/com/velocitypowered/api/event/command/CommandExecuteEvent.java
```diff
@@ -0,0 +1,143 @@
...
+ * This event is fired when a player types in a chat message.
+ */
+public final class CommandExecuteEvent implements ResultedEvent<CommandResult> {
+
+  private final CommandSource commandSource;
+  private final String command;
+  private CommandResult result;
+
+  /**
+   * Constructs a PlayerChatEvent.
```
astei, line 21: `CommandExecuteEvent`

proxy/src/main/java/com/velocitypowered/proxy/command/VelocityCommandManager.java
```diff
@@ -47,11 +57,38 @@ public void unregister(final String alias) {
...
   }
 
+  /**
+   * Calls CommandExecuteEvent.
+   * @param source the command's source
+   * @param cmd the command
+   * @return CompletableFuture of event
+   */
+  public CompletableFuture<CommandExecuteEvent> callCommandEvent(CommandSource source, String cmd) {
+    Preconditions.checkNotNull(source, "invoker");
```
astei, line 67: `source`, not `invoker`

(This is a typo in the original code. Whoops!)

api/src/main/java/com/velocitypowered/api/command/CommandManager.java
```diff
@@ -34,11 +36,42 @@
...
   /**
-   * Attempts to execute a command from the specified {@code cmdLine}.
+   * Calls CommandExecuteEvent and attempts to execute a command from the specified {@code cmdLine}
+   * sync.
    *
    * @param source the command's source
    * @param cmdLine the command to run
    * @return true if the command was found and executed, false if it was not
    */
   boolean execute(CommandSource source, String cmdLine);
```
astei, line 46: I'd like to add `@Deprecated` here

api/src/main/java/com/velocitypowered/api/command/CommandManager.java
```diff
@@ -34,11 +36,42 @@
...
+
+  /**
+   * Attempts to execute a command from the specified {@code cmdLine} sync
```
astei, line 49: Velocity has no concept of `sync`. More accurately this could say `Attempts to execute a command using the specified {@code cmdLine} in a blocking fashion.`

astei:
I am very happy with this now

PR #301, opened 2020-04-29
fix build

Leymooo:


api/src/main/java/com/velocitypowered/api/event/command/CommandExecuteEvent.java
```diff
@@ -9,7 +9,7 @@
 import org.checkerframework.checker.nullness.qual.Nullable;
 
 /**
- * This event is fired when someone executing command
+ * This event is fired when someone executing command.
```
mikroskeem, line 12: `...when someone is executing a command.` is grammatically more correct.

astei:
I hope this works

PR #303, opened 2020-05-02
Passed the used alias to the command executor

Wiiimm:
Imagine having a command with an alias called message and another alias called msg
Depending on the alias the user used to execute this command, a usage message could
be customized to show the exact same alias as the user used to execute.

For example:

	/message
	would result in: Usage: /message <Player> <Message>

	/msg
	would result in: Usage: /msg <Player> <Message>

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/command/Command.java
+++ b/api/src/main/java/com/velocitypowered/api/command/Command.java
@@ -14,9 +14,10 @@ public interface Command {
    * Executes the command for the specified {@link CommandSource}.
    *
    * @param source the source of this command
+   * @param alias the alias used to execute this command
    * @param args the arguments for this command
    */
-  void execute(CommandSource source, String @NonNull [] args);
+  void execute(CommandSource source, String alias, String @NonNull [] args);
 
   /**
    * Provides tab complete suggestions for a command for a specified {@link CommandSource}.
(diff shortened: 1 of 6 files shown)


api/src/main/java/com/velocitypowered/api/command/Command.java
```diff
@@ -14,9 +14,10 @@
    * Executes the command for the specified {@link CommandSource}.
    *
    * @param source the source of this command
+   * @param alias the alias used to execute this command
    * @param args the arguments for this command
    */
-  void execute(CommandSource source, String @NonNull [] args);
+  void execute(CommandSource source, String alias, String @NonNull [] args);
```
astei, line 20: This change breaks _every_ existing Velocity plugin that has a command. This is unacceptable.

Wiiimm: Would that be possible in the re-write of the command system in the near future?

astei: There would be a compatibility shim that would be compatible with the current command API, and then dropped in a major release (TBD).

astei:
@Wiiimm Do you intend to carry this work over to 1.1.0 in an API-compatible way?

There's probably a better path for this. Perhaps we need to introduce some sort of interface (`CommandInvocationContext`?) to handle the logic better.

Wiiimm:
Not really, no. I forgot you were going to completely rewrite the command system anyway. A `CommandInvocationContext` would probably be the better path, yes.

Issue #296, opened 2020-04-25
Make prevent-proxy-connections optional


Cybermaxke:
This refers to the `ip` parameter in the `hasJoined` block, which is optionally used in vanilla mc but isn't in velocity. This should also become an option in velocity.

`https://sessionserver.mojang.com/session/minecraft/hasJoined?username=username&serverId=hash&ip=ip`

"The ip field is optional and when present should be the IP address of the connecting player; it is the one that originally initiated the session request. The notchian server includes this only when prevent-proxy-connections is set to true in server.properties."

https://wiki.vg/Protocol_Encryption

Code for reference: https://github.com/VelocityPowered/Velocity/blob/master/proxy/src/main/java/com/velocitypowered/proxy/connection/client/LoginSessionHandler.java#L54

astei:
#307 is now merged in, closing this issue.

PR #307, opened 2020-05-08
Add prevent-proxy-connections option

mikroskeem:
Addresses #296 

This PR is basically ready, but I couldn't come up with decent description for the option as there isn't *that* much information available for this in my opinion.

Perhaps option name could be different too?

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/proxy/config/ProxyConfig.java
+++ b/api/src/main/java/com/velocitypowered/api/proxy/config/ProxyConfig.java
@@ -63,6 +63,15 @@ public interface ProxyConfig {
   boolean isOnlineMode();
 
   /**
+   * If client's ISP/AS sent from this proxy is different from the one from Mojang's
+   * authentication server, the player is kicked. This disallows some VPN and proxy
+   * connections but is a weak form of protection.
+   *
+   * @return whether to prevent client proxy connections by checking the IP with Mojang servers
+   */
+  boolean shouldPreventClientProxyConnections();
+
+  /**
    * Get a Map of all servers registered in <code>velocity.toml</code>. This method does
    * <strong>not</strong> return all the servers currently in memory, although in most cases it
    * does. For a view of all registered servers, see {@link ProxyServer#getAllServers()}.
--- a/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
@@ -55,6 +55,14 @@ public class VelocityConfiguration extends AnnotatedConfig implements ProxyConfi
   private boolean onlineMode = true;
 
   @Comment({
+      "If client's ISP/AS sent from this proxy is different from the one from Mojang's",
+      "authentication server, the player is kicked. This disallows some VPN and proxy",
+      "connections but is a weak form of protection."
+  })
+  @ConfigKey("prevent-client-proxy-connections")
+  private boolean preventClientProxyConnections = false;
+
+  @Comment({
       "Should we forward IP addresses and other data to backend servers?",
       "Available options:",
       "- \"none\":   No forwarding will be done. All players will appear to be connecting from the",
@@ -326,6 +334,11 @@ public class VelocityConfiguration extends AnnotatedConfig implements ProxyConfi
   @Override
   public boolean isOnlineMode() {
     return onlineMode;
+  }
+
+  @Override
+  public boolean shouldPreventClientProxyConnections() {
+    return preventClientProxyConnections;
   }
 
   public PlayerInfoForwarding getPlayerInfoForwardingMode() {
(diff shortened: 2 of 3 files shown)


Cybermaxke:
Maybe a bit more specific? `prevent-client-proxy-connections`?

mikroskeem:
That's better and reduces confusion. Now only thing to figure out is proper description for the option

hugmanrique:
The Minecraft Wiki [says the following](https://minecraft.gamepedia.com/Server.properties):

> If the ISP/AS sent from the server is different from the one from Mojang's authentication server, the player is kicked
true - Enabled. Server prevents users from using vpns or proxies.
false - Disabled. The server doesn't prevent users from using vpns or proxies.

mikroskeem:
Ready for proper reviewing now.

api/src/main/java/com/velocitypowered/api/proxy/config/ProxyConfig.java
```diff
@@ -62,6 +62,13 @@
    */
   boolean isOnlineMode();
 
+  /**
+   * TODO
```
hugmanrique, line 66: Mind adding the same description here?

proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
```diff
@@ -54,6 +54,14 @@
   @ConfigKey("online-mode")
   private boolean onlineMode = true;
 
+  @Comment({
+      "If client's ISP/AS sent from this proxy is different from the one from Mojang's",
+      "authentication server, the player is kicked. Effectively disallows players from",
+      "using proxies or VPNs when enabled."
+  })
+  @ConfigKey("prevent-client-proxy-connections")
+  private boolean preventClientProxyConnections = false;
```
hugmanrique, line 63: Previous code defaulted to passing the IP param, should this be `true` instead?

mikroskeem: Vanilla has this off by default, so I went with false.

api/src/main/java/com/velocitypowered/api/proxy/config/ProxyConfig.java
mikroskeem: Oops

proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
astei: I'm inclined to say we should leave this off, as `prevent-proxy-connections` is a rather weak form of protection that can be easily bypassed.

api/src/main/java/com/velocitypowered/api/proxy/config/ProxyConfig.java
```diff
@@ -62,6 +62,15 @@
    */
   boolean isOnlineMode();
 
+  /**
+   * If client's ISP/AS sent from this proxy is different from the one from Mojang's"
```
astei, line 66: Errant `"`

api/src/main/java/com/velocitypowered/api/proxy/config/ProxyConfig.java
```diff
@@ -62,6 +62,15 @@
...
+   * authentication server, the player is kicked. Effectively disallows players from
+   * using proxies or VPNs when enabled.
+   *
+   * @return prevent client proxy connections
```
astei, line 70: Make this a little bit more explanatory

proxy/src/main/java/com/velocitypowered/proxy/connection/client/LoginSessionHandler.java
```diff
@@ -96,8 +96,11 @@ public boolean handle(EncryptionResponse packet) {
 
       String playerIp = ((InetSocketAddress) mcConnection.getRemoteAddress()).getHostString();
       String url = String.format(MOJANG_HASJOINED_URL,
-          urlFormParameterEscaper().escape(login.getUsername()), serverId,
-          urlFormParameterEscaper().escape(playerIp));
+          urlFormParameterEscaper().escape(login.getUsername()), serverId);
+
+      if (server.getConfiguration().shouldPreventClientProxyConnections()) {
+        url += urlFormParameterEscaper().escape("&ip=" + playerIp);
```
astei, line 102: Not sure if we need to escape the `&ip=` here. We probably should escape the IP but that's it.

api/src/main/java/com/velocitypowered/api/proxy/config/ProxyConfig.java
mikroskeem: Yeah copied from the configuration class, oops.

api/src/main/java/com/velocitypowered/api/proxy/config/ProxyConfig.java
```diff
@@ -62,6 +62,15 @@
    */
   boolean isOnlineMode();
 
+  /**
+   * If client's ISP/AS sent from this proxy is different from the one from Mojang's"
+   * authentication server, the player is kicked. Effectively disallows players from
```
astei, line 67: "Effectively" is too strong. I'd do "This disallows some VPNs and proxy connections but is a weak form of protection."

PR #314, opened 2020-05-28
Added new event LoginFailEvent

Nan1t:
Added new event to listen when the login process has failed.

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/event/connection/LoginFailEvent.java
+++ b/api/src/main/java/com/velocitypowered/api/event/connection/LoginFailEvent.java
@@ -1,0 +1,40 @@
+package com.velocitypowered.api.event.connection;
+
+import com.velocitypowered.api.proxy.InboundConnection;
+
+/**
+ * This event is fired when login process has failed for some reason
+ */
+public final class LoginFailEvent {
+
+    private final InboundConnection connection;
+    private final String username;
+    private final Reason reason;
+
+    public LoginFailEvent(InboundConnection connection, String username, Reason reason) {
+        this.connection = connection;
+        this.username = username;
+        this.reason = reason;
+    }
+
+    public InboundConnection getConnection() {
+        return connection;
+    }
+
+    public String getUsername() {
+        return username;
+    }
+
+    public Reason getReason() {
+        return reason;
+    }
+
+    public enum Reason {
+
+        ONLINE_MODE_ONLY,
+        ALREADY_CONNECTED,
+        MANUALLY,
+        UNDEFINED,
+
+    }
+}
--- a/api/src/main/java/com/velocitypowered/api/event/connection/LoginFailEvent.java
+++ b/api/src/main/java/com/velocitypowered/api/event/connection/LoginFailEvent.java
@@ -33,7 +33,9 @@ public final class LoginFailEvent {
 
         ONLINE_MODE_ONLY,
         ALREADY_CONNECTED,
-        MANUALLY,
+        NO_AVAILABLE_SERVERS,
+        PLAYER_INITIATED,
+        PLUGIN,
         UNDEFINED,
 
     }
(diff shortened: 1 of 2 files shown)


proxy/src/main/java/com/velocitypowered/proxy/connection/client/LoginSessionHandler.java
```diff
@@ -127,16 +128,25 @@ public boolean handle(EncryptionResponse packet) {
           } else if (profileResponse.getStatusCode() == 204) {
             // Apparently an offline-mode user logged onto this online-mode proxy.
             inbound.disconnect(VelocityMessages.ONLINE_MODE_ONLY);
+            server.getEventManager().fireAndForget(new LoginFailEvent(inbound,
```
astei, line 131: Consider a `disconnect()` private function for cases like these

proxy/src/main/java/com/velocitypowered/proxy/connection/client/LoginSessionHandler.java
```diff
@@ -225,6 +237,12 @@ private void initializePlayer(GameProfile profile, boolean onlineMode) {
   }
 
   private void finishLogin(ConnectedPlayer player) {
+    Optional<RegisteredServer> toTry = player.getNextServerToTry();
+    if (!toTry.isPresent()) {
+      player.disconnect(VelocityMessages.NO_AVAILABLE_SERVERS);
```
astei, line 242: You forgot this case

api/src/main/java/com/velocitypowered/api/event/connection/LoginFailEvent.java
```diff
@@ -0,0 +1,40 @@
...
+
+    public Reason getReason() {
+        return reason;
+    }
+
+    public enum Reason {
+
+        ONLINE_MODE_ONLY,
+        ALREADY_CONNECTED,
+        MANUALLY,
```
astei, line 36: This is not a very good name - a more accurate one would be `PLUGIN`

api/src/main/java/com/velocitypowered/api/event/connection/LoginFailEvent.java
```diff
@@ -0,0 +1,40 @@
...
+        UNDEFINED,
+
```
astei, line 38: Consider adding a new case for a player-initiated disconnect

proxy/src/main/java/com/velocitypowered/proxy/connection/client/LoginSessionHandler.java
Leymooo: i think this should not be in this pr. https://github.com/VelocityPowered/Velocity/commit/d2b8271eb4b23578d833058af4e2fb1cbff4ba94#diff-8e51e6774e951b58c22066197a838ad4

proxy/src/main/java/com/velocitypowered/proxy/connection/client/LoginSessionHandler.java
Naanit: Done in the new commit. Thanks for the tip.

proxy/src/main/java/com/velocitypowered/proxy/connection/client/LoginSessionHandler.java
astei: Agreed. This reverts a feature added in 1.1.0 to allow users to configure the proxy to run without servers configured. Please revert this.

proxy/src/main/java/com/velocitypowered/proxy/connection/client/LoginSessionHandler.java
```diff
@@ -289,4 +305,8 @@ public void disconnected() {
       connectedPlayer.teardown();
     }
   }
+
+  private void disconnect(String username, LoginFailEvent.Reason reason){
```
astei, line 309: When I suggested this function name, I also wanted it to take over the responsibility of also disconnecting the user, not just firing the event. Please fix this or change the misleading function name.

astei:
Doing some spring cleaning, there's been no response on this so I'm closing this PR for now. If you are still interested let me know and let's find an alternative way to implement this.

Note that I recently did #324 which might handle at least some of the use cases for `LoginFailEvent` in `DisconnectEvent` (which, arguably, is where the functionality ought to have been).

PR #315, opened 2020-06-02
Implement legacy forwarding BungeeGuard handshake support

lucko:


proxy/src/main/java/com/velocitypowered/proxy/config/PlayerInfoForwarding.java
```diff
@@ -3,5 +3,6 @@
 public enum PlayerInfoForwarding {
   NONE,
   LEGACY,
+  LEGACY_BUNGEEGUARD,
```
astei, line 6: Shorten to `BUNGEEGUARD`

proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
```diff
@@ -61,6 +61,8 @@
       "            proxy and will have offline-mode UUIDs.",
       "- \"legacy\": Forward player IPs and UUIDs in a BungeeCord-compatible format. Use this if",
       "            you run servers using Minecraft 1.12 or lower.",
+      "- \"legacy_bungeeguard\": Forward player IPs and UUIDs in a BungeeCord-compatible format.",
```
astei, line 64: I'd prefer a description closer to this:

```
"bungeeguard": Forward player IPs and UUIDs in a format supported by
               the BungeeGuard plugin. Use this if you run servers
               using Minecraft 1.12 and below, are on a shared
               host, and would like to use BungeeGuard to provide
               additional security.
```

I'd prefer if this was made a bit more concise, though.

proxy/src/main/java/com/velocitypowered/proxy/connection/backend/VelocityServerConnection.java
```diff
@@ -133,6 +137,24 @@ private String createLegacyForwardingAddress() {
     return data.toString();
   }
 
+  private String createBungeeGuardForwardingAddress(byte[] forwardingSecret) {
+    StringBuilder data = new StringBuilder()
```
astei, line 141: Instead of duplicating `createLegacyForwardingAddress()`, perhaps we can make this more generic by refactoring `createLegacyForwardingAddress()` to allow the list of profile properties to vary.

PR #319, opened 2020-06-21
Added way to get the "plugins" folder

iamceph:
Also added getDataFolder for the plugin itself, you can get it from PluginDescription.

This should be essentials for all plugin developers. :)


astei:
What would be the purpose of getting the `plugins` folder? I can sorta see a use case where one can change the plugins folder on proxy startup, but I think few people actually need that functionality.

I can sort of understand `PluginDescription#getDataFolder()`. although I disagree with its placement in `PluginDescription` (and injecting with `@DataDirectory` for your own plugin is documented as the way to do this).

A248:
Can't you just call File#getParentFile on the File derived from the Path injected with @DataDirectory? Or is the assumption that all plugin folders are nested directly under the plugins folder incorrect?

iamceph:
I agree that getParentFolder and getPlugins folder are useless, but getDataFolder for plugins should be usefull. @astei Where else do you want it? :)

PR #321, opened 2020-06-29
adventure API integration into Velocity

astei:
There's a lot going on at a glance, so I want to open this up for review for the community at large.

Velocity has historically used the `text` library for dealing with Minecraft chat components. Now with the release of Minecraft 1.16, the opportunity was taken to evolve text into a more encompassing library, called `adventure`. This PR is an _initial draft_ and not all functionality in the new APIs are supported (as of now, only the functionality that was available in Velocity 1.0.x is available), which will be implemented in the future, as I am only focused on gathering API feedback at this stage.

The changes made to the API are as follows:

* `ProxyServer`, `CommandSource` (and by extension `Player`), and `RegisteredServer` now implement `Audience`
  * `ProxyServer` applies actions to everyone on the proxy, including the console
  * `RegisteredServer` applies to everyone connected to a specific server on the proxy
* Every API that took (or returned) `net.kyori.text.Component` has been deprecated and a replacement API taking the `adventure` equivalent (`net.kyori.adventure.text.Component`) was added
* All Velocity boss bar and title APIs were deprecated - these are replaced by `adventure` equivalents.
* The proxy now works internally with adventure `Component`s

With that being said, I am opening this issue for comment on the changes to the API so far.


api/src/main/java/com/velocitypowered/api/proxy/ProxyAudience.java
```diff
@@ -0,0 +1,71 @@
...
+import net.kyori.adventure.text.Component;
+import net.kyori.adventure.title.Title;
+import org.checkerframework.checker.nullness.qual.NonNull;
+
+/**
+ * Indicates an {@link Audience} that is on the proxy. This interface contains no-op default methods
+ * that are used to bridge compatibility issues with the new adventure API. This interface will go
+ * away in Velocity 2.0.0.
+ */
+public interface ProxyAudience extends Audience {
```
kashike, line 17: Should be deprecated since it is going away.

A248:
This looks nice. Having methods ending in -Component is a little ugly, but I can't think of a better way to handle methods returning Component objects. Though, it did seem like a good use of the opportunity to rename KickedFromServerEvent#getserverKickReason

proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
```diff
@@ -605,11 +606,8 @@ public InetSocketAddress getBoundAddress() {
...
 
   @Override
-  public void sendMessage(net.kyori.adventure.text.@NonNull Component message) {
-    Preconditions.checkNotNull(message, "message");
-    for (ConnectedPlayer player : connectionsByUuid.values()) {
-      player.sendMessage(message);
-    }
+  public @NonNull Iterable<? extends Audience> audiences() {
+    return this.getAllPlayers();
   }
```
Crypnotic: Changes list says `ProxyServer` applies actions to all players including console, does this not just give the player audiences?

astei: Should be fixed now.

api/src/main/java/com/velocitypowered/api/event/ResultedEvent.java
```diff
@@ -92,7 +93,11 @@ public boolean isAllowed() {
       return status;
     }
 
-    public Optional<Component> getReason() {
+    public Optional<net.kyori.text.Component> getReason() {
```
hugmanrique, line 96: This should be deprecated

api/src/main/java/com/velocitypowered/api/event/player/KickedFromServerEvent.java
```diff
@@ -20,7 +21,7 @@
 
   private final Player player;
   private final RegisteredServer server;
-  private final @Nullable Component originalReason;
+  private final net.kyori.adventure.text.Component originalReason;
```
hugmanrique, line 24: Is the `@Nullable` annotation removal both here and in the new constructor intentional?
If so, the constructor is missing the non-null precondition.

api/src/main/java/com/velocitypowered/api/event/player/KickedFromServerEvent.java
```diff
@@ -147,10 +186,10 @@ public static DisconnectPlayer create(Component reason) {
    */
   public static final class RedirectPlayer implements ServerKickResult {
 
-    private final Component message;
+    private final net.kyori.adventure.text.Component message;
     private final RegisteredServer server;
 
-    private RedirectPlayer(RegisteredServer server, @Nullable Component message) {
+    private RedirectPlayer(RegisteredServer server, net.kyori.adventure.text.Component message) {
```
hugmanrique, line 192: Mark message as `@Nullable`

api/src/main/java/com/velocitypowered/api/event/player/KickedFromServerEvent.java
```diff
@@ -191,9 +250,10 @@ public static ServerKickResult create(RegisteredServer server) {
    */
   public static final class Notify implements ServerKickResult {
 
-    private final Component message;
+    private final net.kyori.adventure.text.Component message;
 
-    private Notify(Component message) {
+    @Deprecated
+    private Notify(net.kyori.adventure.text.Component message) {
```
hugmanrique, line 256: This is using the new Component and is private, shouldn't be deprecated

api/src/main/java/com/velocitypowered/api/proxy/config/ProxyConfig.java
```diff
@@ -44,8 +43,17 @@
...
-  Component getMotdComponent();
+  @Deprecated
+  net.kyori.text.Component getMotdComponent();
+
+  /**
+   * Get the MOTD component shown in the tab list.
+   *
+   * @return the motd component
+   */
+  net.kyori.adventure.text.Component getMotd();
```
hugmanrique, line 56: Minor nit: It's a shame this method cannot be named consistently. Open to change in 2.0 once the other gets removed?

proxy/src/main/java/com/velocitypowered/proxy/server/VelocityRegisteredServer.java
```diff
@@ -145,4 +148,9 @@ public boolean sendPluginMessage(ChannelIdentifier identifier, ByteBuf data) {
   public String toString() {
     return "registered server: " + serverInfo;
   }
+
+  @Override
+  public @NonNull Iterable<? extends Audience> audiences() {
```
hugmanrique, line 153: Superfluous `@NonNull` annotation

api/src/main/java/com/velocitypowered/api/proxy/config/ProxyConfig.java
astei: Yes, as with everything else. (`ProxyConfig` was basically created with the "Extract Interface" refactoring in IntelliJ IDEA, and hasn't really evolved past that. It's a leaky abstraction.)

api/src/main/java/com/velocitypowered/api/proxy/ProxyServer.java
```diff
@@ -17,13 +17,14 @@
...
-import net.kyori.text.Component;
+import net.kyori.adventure.audience.MultiAudience;
+import net.kyori.adventure.text.Component;
 import org.checkerframework.checker.nullness.qual.NonNull;
 
 /**
  * Provides an interface to a Minecraft server proxy.
  */
-public interface ProxyServer {
+public interface ProxyServer extends MultiAudience {
```
Electroid, line 27: I would recommend extending the more generic `Audience` in the `api` interfaces, then using `MultiAudience` in the implementation (for `ProxyServer` and co.) That way you don't expose `audiences()` in the spec.

astei:
Boss bar API is delayed by KyoriPowered/adventure#89 as the only other solution is not acceptable to me.

astei:
Updated for recent changes and removed some contradictory content (since we now support boss bars, whereas before we didn't).

PR #329, opened 2020-07-14
Add async suggestions

Proximyst:
Fairly simple.

To use them, the developer should still implement sync ones for plugins which depends on others' suggestions. Whether this includes data fetched from e.g. a database is up to them.

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/command/RawCommand.java
+++ b/api/src/main/java/com/velocitypowered/api/command/RawCommand.java
@@ -2,6 +2,7 @@ package com.velocitypowered.api.command;
 
 import com.google.common.collect.ImmutableList;
 import java.util.List;
+import java.util.concurrent.CompletableFuture;
 import org.checkerframework.checker.nullness.qual.NonNull;
 
 /**
@@ -29,13 +30,19 @@ public interface RawCommand extends Command {
    * @param currentLine the current, partial command line for this command
    * @return tab complete suggestions
    */
-  default List<String> suggest(CommandSource source, String currentLine) {
-    return ImmutableList.of();
+  default CompletableFuture<List<String>> suggest(CommandSource source, String currentLine) {
+    return CompletableFuture.completedFuture(ImmutableList.of());
+  }
+
+  @Override
+  default CompletableFuture<List<String>> suggestAsync(CommandSource source,
+      String @NonNull [] currentArgs) {
+    return suggest(source, String.join(" ", currentArgs));
   }
 
   @Override
   default List<String> suggest(CommandSource source, String @NonNull [] currentArgs) {
-    return suggest(source, String.join(" ", currentArgs));
+    return suggestAsync(source, currentArgs).join();
   }
 
   @Override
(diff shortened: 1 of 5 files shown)


proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientPlaySessionHandler.java
```diff
@@ -449,18 +451,20 @@ public void handleTabCompleteResponse(TabCompleteResponse response) {
...
+              response.getOffers().add(new Offer(offer, null));
+            }
+            response.getOffers().sort(null);
+            player.getConnection().write(response);
+          } catch (Exception e) {
+            logger.error("Unable to provide tab list completions for {} for command '{}'",
+                player.getUsername(),
+                command, e);
+          }
+        });
```
astei, line 467: Should run this completion stage on the event loop

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientPlaySessionHandler.java
```diff
@@ -397,26 +397,28 @@ private boolean handleCommandTabComplete(TabCompleteRequest packet) {
...
+          int startPos = packet.getCommand().lastIndexOf(' ') + 1;
+          if (startPos > 0) {
+            TabCompleteResponse resp = new TabCompleteResponse();
+            resp.setTransactionId(packet.getTransactionId());
+            resp.setStart(startPos);
+            resp.setLength(packet.getCommand().length() - startPos);
+            resp.getOffers().addAll(offers);
+            player.getConnection().write(resp);
+          }
+        });
```
astei, line 420: Should run this completion stage on the event loop

Proximyst:
Amended PR for your requested changes.

api/src/main/java/com/velocitypowered/api/command/RawCommand.java
```diff
@@ -29,15 +30,21 @@ default void execute(CommandSource source, String @NonNull [] args) {
...
 
   @Override
-  default List<String> suggest(CommandSource source, String @NonNull [] currentArgs) {
+  default CompletableFuture<List<String>> suggestAsync(CommandSource source,
+      String @NonNull [] currentArgs) {
     return suggest(source, String.join(" ", currentArgs));
   }
 
+  @Override
+  default List<String> suggest(CommandSource source, String @NonNull [] currentArgs) {
```
hugmanrique, line 44: I would deprecate this method/rename the other one to avoid confusing API clients. How do they know which of the 3 methods should they use?

astei: If I had to be completely honest, it's probably better to deprecate `Command` and split it into `RawCommand` and `RegularCommand`, perhaps also introducing `CommandContext` and `CommandExecutionContext` interface at the same time, with `CommandContext` providing command metadata and `CommandExecutionContext` focusing on the invocation of a command.

My 2 cents.

PR #330, opened 2020-07-16
New command API

hugmanrique:
Introduces a new command framework that supports both legacy `String[] args` commands and 1.13+ Brigadier commands.

As discussed in https://github.com/VelocityPowered/Velocity/pull/329#discussion_r454969425 and https://github.com/VelocityPowered/Velocity/issues/232, the current command framework lacks support for new command features. This PR adds 3 subinterfaces to `Command`: `BrigadierCommand`, `LegacyCommand` and `RawCommand`. All execution-related methods such as `execute` and `hasPermission` now take a single `CommandInvocation` argument that includes the command source and additional execution metadata such as the `String[] args` for legacy commands.

TODO:

- [x] Decide on command execution context name. Candidates so far are `CommandExecutionContext`, `CommandContext` and `CommandInvocation`
- [x] Move built-in commands to Brigadier
- [x] Test all the things!!1!
  - [x] Brigadier aliases as redirects (see https://github.com/Mojang/brigadier/issues/46, do we need to create our own redirects system?)


api/src/main/java/com/velocitypowered/api/newcommand/BrigadierCommand.java
```diff
@@ -0,0 +1,45 @@
+package com.velocitypowered.api.newcommand;
+
+import com.mojang.brigadier.exceptions.CommandSyntaxException;
+import com.mojang.brigadier.suggestion.SuggestionProvider;
+import com.mojang.brigadier.suggestion.Suggestions;
+import com.mojang.brigadier.tree.LiteralCommandNode;
+import com.velocitypowered.api.command.CommandSource;
+
+/**
+ * A command that uses the Brigadier library for parsing.
```
astei, line 10: `A command that uses Brigadier for parsing the command and providing suggestions to the client.`

api/src/main/java/com/velocitypowered/api/newcommand/Command.java
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * @param <C> the type of the execution context
+ */
+public interface Command<C extends CommandExecutionContext> {
+
+    /**
+     * Represents the command type, which determines
+     * its parsing capabilities.
+     */
+    // TODO Introduce CommandMeta and move there?
```
astei, line 18: I think this is a good idea

api/src/main/java/com/velocitypowered/api/newcommand/CommandManager.java
```diff
@@ -0,0 +1,55 @@
...
+     */
+    void unregister(Command<?> command);
+
+    /**
+     * Unregisters the command with the specified alias.
+     *
+     * @param alias the alias of the command
+     */
+    // TODO Previous javadoc is ambiguous, does this unregister all the aliases of the command,
+    // or just the given alias?
```
astei, line 32: The previous Javadoc description should have read `Unregisters the specified alias from the command manager`.

build.gradle
```diff
@@ -30,7 +30,7 @@ allprojects {
         getCurrentShortRevision = {
             new ByteArrayOutputStream().withStream { os ->
                 exec {
-                    executable = "git"
+                    executable = "git.bat"
```
astei, line 33: Is this some kind of local change you made? We don't build on Windows.

hugmanrique: Didn't mean to push this, my bad. Signs of my janky WSL.

api/src/main/java/com/velocitypowered/api/newcommand/Command.java
hugmanrique: I'm leaning more towards `Command` containing the immutable meta and have an associated `CommandExecutor`. This way we can have builders that automatically register commands on construction (similar to the scheduler system). Also, we can get rid of the `Type` enum to let users provide their own `Command` implementations (i.e. get rid of `RawCommand`).

astei: I'm happy to run with this idea (`Command.Type` is kind of a kludge anyway).

hugmanrique:
Alright, here's some usage examples:

```java
import static com.mojang.brigadier.arguments.IntegerArgumentType.getInteger;
import static com.mojang.brigadier.arguments.IntegerArgumentType.integer;
import static com.mojang.brigadier.builder.RequiredArgumentBuilder.argument;
import static com.velocitypowered.api.command.BrigadierCommand.argumentBuilder;

import com.google.inject.Inject;
import com.mojang.brigadier.CommandDispatcher;
import com.velocitypowered.api.command.BrigadierCommand;
import com.velocitypowered.api.command.CommandManager;
import com.velocitypowered.api.command.LegacyCommand;
import com.velocitypowered.api.event.Subscribe;
import com.velocitypowered.api.event.proxy.ProxyInitializeEvent;
import com.velocitypowered.api.plugin.Plugin;
import com.velocitypowered.api.proxy.ProxyServer;
import net.kyori.adventure.text.TextComponent;
import org.slf4j.Logger;

@Plugin(id = "test-plugin")
public class TestPlugin {

    @Inject
    public TestPlugin(final ProxyServer server) {
        this.server = server;
    }

    @Subscribe
    public void onProxyInitialization(final ProxyInitializeEvent event) {
        CommandManager commandManager = server.getCommandManager();

        BrigadierCommand command = commandManager.brigadierBuilder().register(
            argumentBuilder("foo")
                .then(argument("bar", integer()))
                    .executes(c -> {
                        System.out.println("Bar is " + getInteger(c, "bar"));
                        return 1;
                    })
                .executes(c -> {
                    System.out.println("Called foo with no arguments");
                    return 1;
                })
        );

        LegacyCommand legacyCommand = commandManager.legacyBuilder().register(context -> {
            CommandSource source = context.source();
            String[] arguments = context.arguments();

            source.sendMessage(TextComponent.of("foo"));
        });
    }
}
```

api/src/main/java/com/velocitypowered/api/command/CommandManager.java
```diff
@@ -1,87 +1,60 @@
...
    *
-   * @param command the command to register
-   * @param aliases the alias to use
-   *
-   * @deprecated This method requires at least one alias, but this is only enforced at runtime.
-   *             Prefer {@link #register(String, Command, String...)} instead.
+   * @return a legacy command builder
    */
-  @Deprecated
-  void register(Command command, String... aliases);
```
astei, line 20: Question: As this presently stands this would break the API. I assume you just want to put the API in front of users before you add any backwards compatibility?

api/src/main/java/com/velocitypowered/api/command/LegacyCommand.java
```diff
@@ -0,0 +1,31 @@
...
+     */
+    Builder aliases(String... aliases);
+
+    /**
+     * Registers the command with the specified execution handler.
+     *
+     * @param onExecute the execution handler
+     * @return the registered command
+     */
+    LegacyCommand register(Consumer<LegacyCommandExecutionContext> onExecute);
```
hugmanrique, line 29: I'm thinking of replacing this consumer by a `LegacyCommandExecutor` functional interface (I think) users are more familiarized with.

api/src/main/java/com/velocitypowered/api/command/CommandManager.java
hugmanrique: Yeah, this would require a major bump, so it's a long-term PR. Users are expected to not instantiate the `Command` interface/subclass it and instead use `LegacyCommand.Builder` and `BrigadierCommand.Builder` instances obtainable from the command manager.

hugmanrique:
I reached a middle ground so current plugins can still directly register commands while adding a fluent command builder some may prefer (not many? do we even need it?).

astei:
My API compatibility concerns are assuaged for now. In 2.0.0 we'll just outright force the explicit use of `LegacyCommand` or `BrigadierCommand`.

alexstaeding:
`CommandExecutionContext` is a bit too wordy in my opinion. Why not just `CommandContext`?

alexstaeding:
For the sake of conventions and consistency with the rest of the API, methods should start with a verb. I think `CommandSource source();` should rather be `CommandSource getSource();`.

hugmanrique:
It's common for immutable classes with getter accessors. See https://stackoverflow.com/a/1680515/4514467 for a more in depth answer.

api/src/main/java/com/velocitypowered/api/command/BrigadierCommand.java
```diff
@@ -0,0 +1,46 @@
...
+import com.mojang.brigadier.tree.CommandNode;
+
+/**
+ * A command that uses Brigadier for parsing the command and providing
+ * suggestions to the client.
+ */
+public interface BrigadierCommand extends Command<BrigadierCommandInvocation> {
+
+  /**
+   * Returns an {@link ArgumentBuilder} used to specify the structure of
```
astei, line 14: I would like to see some better description on why this is the case. This isn't obvious on first glance.

api/src/main/java/com/velocitypowered/api/command/BrigadierCommandInvocation.java
```diff
@@ -0,0 +1,18 @@
...
+
+import com.mojang.brigadier.ParseResults;
+
+/**
+ * Contains the invocation data for a {@link BrigadierCommand}.
+ */
+public interface BrigadierCommandInvocation extends CommandInvocation {
+
+  /**
+   * Returns the valid parse results.
```
astei, line 11: `Returns the result of parsing the given command.`

api/src/main/java/com/velocitypowered/api/command/Command.java
```diff
@@ -1,59 +1,73 @@
...
 
   /**
-   * Executes the command for the specified {@link CommandSource}.
+   * Executes the command for the specified invocation.
    *
-   * @param source the source of this command
-   * @param args the arguments for this command
+   * @param invocation the invocation context
    */
-  void execute(CommandSource source, String @NonNull [] args);
```
astei, line 20: Still API-breaking here - is there a better solution?

api/src/main/java/com/velocitypowered/api/command/LegacyCommand.java
```diff
@@ -0,0 +1,8 @@
+package com.velocitypowered.api.command;
+
+/**
+ * A legacy 1.12-style command.
```
astei, line 4: `A legacy command, modelled after the convention popularized by Bukkit and BungeeCord. Prefer {@code BrigadierCommand} if possible, which is also backwards-compatible with older clients.`

Assuming (of course) that Brigadier support in Velocity does support older clients. Test this perhaps?

api/src/main/java/com/velocitypowered/api/command/RawCommand.java
```diff
@@ -1,67 +1,10 @@
...
-import com.google.common.collect.ImmutableList;
-import java.util.List;
-import java.util.concurrent.CompletableFuture;
-import org.checkerframework.checker.nullness.qual.NonNull;
-
 /**
- * A specialized sub-interface of {@code Command} which indicates that the proxy should pass a
- * raw command to the command. This is useful for bolting on external command frameworks to
+ * A specialized sub-interface of {@code Command} which indicates that the proxy should pass
+ * a raw command to the command. This is useful for bolting on external command frameworks to
```
astei, line 5: `should pass the command and its arguments directly without further processing`

proxy/src/main/java/com/velocitypowered/proxy/command/VelocityBrigadierCommand.java
```diff
@@ -0,0 +1,124 @@
...
+import com.velocitypowered.api.command.BrigadierCommandInvocation;
+import com.velocitypowered.api.command.CommandSource;
+import java.util.List;
+import java.util.Locale;
+import java.util.concurrent.CompletableFuture;
+
+final class VelocityBrigadierCommand implements BrigadierCommand {
+
+  /**
+   * Returns a node builder with the given alias to the specified destination node.
```
astei, line 20: Should explain why we need to use this method.

proxy/src/main/java/com/velocitypowered/proxy/command/VelocityCommandManager.java
```diff
@@ -20,47 +26,101 @@
...
+  //
+  // By design, the API doesn't provide CommandInvocation implementations.
+  // Commands are not meant to be executed directly. Instead, users should
+  // call CommandManager#execute.
+
+  private final Map<String, Command<?>> commands = new HashMap<>();
+
+  private final CommandInvocationFactoryRegistry invocationFactory =
+          new CommandInvocationFactoryRegistry();
+  private final CommandDispatcher<CommandSource> dispatcher = new CommandDispatcher<>();
```
astei, line 45: I would call this field `brigadierDispatcher`

proxy/src/main/java/com/velocitypowered/proxy/command/VelocityCommandManager.java
```diff
@@ -69,107 +129,111 @@ public void unregister(final String alias) {
...
-  public boolean hasCommand(String command) {
-    return commands.containsKey(command);
-  }
+      return command.suggestAsync(invocation).thenApply(ImmutableList::copyOf);
+    } catch (final Exception e) {
+      if (e.getCause() instanceof CommandSyntaxException) {
+        return CompletableFuture.completedFuture(ImmutableList.of());
+      }
 
-  public Set<String> getAllRegisteredCommands() {
```
astei, line 152: I would like to us to be able to inject proxy commands into tab-complete requests and into the Brigadier command tree if we can.

hugmanrique: I think this is already handled by https://github.com/VelocityPowered/Velocity/blob/21ad6a433f8b06d61de7e35bf1cf0658d5ca0f6e/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientPlaySessionHandler.java#L400, `CommandManager#offerSuggestions` merges all suggestions from all command types, including `BrigadierCommand`s.

astei: But how about merging the proxy's Brigadier tree with that from the remote server?

api/src/main/java/com/velocitypowered/api/command/Command.java
hugmanrique: If a command class now doesn't implement `LegacyCommand`, `BrigadierCommand` or `RawCommand`, the fallback methods will be executed. On 2.0.0 we can open up the possibility for custom `CommandInvocation` factories and `Command` implementations. As it stands, I don't think there's a way to detect a class using the deprecated methods without calling them.

api/src/main/java/com/velocitypowered/api/command/BrigadierCommand.java
hugmanrique: This is a helper method, callers can still use the longer `LiteralArgumentBuilder.<CommandSource>.literal(alias)`. Most examples I could find of Brigadier commands prefer to use the statically imported `literal()`, but it's not possible in our case since it requires to specify the `<CommandSource>` type.

api/src/main/java/com/velocitypowered/api/command/Command.java
```diff
@@ -4,10 +4,35 @@
...
+ *
+ * <li>{@link LegacyCommand}, modelled after the convention popularized by
+ * Bukkit and BungeeCord. Older classes directly implementing {@link Command}
+ * are suggested to migrate to this interface.
+ *
+ * <li>{@link RawCommand}, useful for bolting on external command frameworks
+ * to Velocity.
+ *
+ * </ul>
+ *
```
astei, line 28: `<p>For this reason, the legacy `execute()`, `suggest()`, and `hasPermission()` methods are deprecated and will be removed in Velocity 2.0.0. We suggest implementing one of the more specific subinterfaces instead.`

hugmanrique:
Testing revealed some inconsistencies regarding the arguments array passed to tab completion and permission checks in the current 1.1.0 version and this PR. I'm trying to move everything to Brigadier, including legacy commands. This will also falicitate the merging of the proxy's Brigadier tree with that from the remote server.

If benchmarking shows this alternative is too slow, I will most likely revert the following commits.

astei:
@hugmanrique This is mostly due to the unintuitive behavior of the Java `String.split()` method. I patch it using https://github.com/VelocityPowered/Velocity/blob/dev/1.1.0/proxy/src/main/java/com/velocitypowered/proxy/command/VelocityCommandManager.java#L237 but you could also explore an approach using Guava's `Splitter`.

hugmanrique:
Still need to test it, but I'm 99% sure the node merging implementation a) is incorrect, and b) can be optimized.

astei:
The main optimization that can be made is if Mojang/brigadier#68 gets merged in.

hugmanrique:
Added tests for command registration, execution, and suggestions (includes Brigadier, legacy, raw and deprecated variants).
I don't know if testing the actual protocol is worth the time mocking so much stuff, so will test on 1.8.9 and 1.16.1 clients.

hugmanrique:
Last commit fixes e.g. `/lpv` incorrectly returning `false` on execute. Brigadier's redirect system is 100% broken, we sadly construct "duplicate" literals instead.

hugmanrique:
Alright, tested execution, tab complete and permission checks on 1.12.2, 1.16.1 and the console and everything looks fine. This is now ready for review.

One note that should be added to the docs: Brigadier commands should **not** use `#requires` to perform permission checks. This leads to the command being omitted from the parse tree and `#execute` throwing `dispatcherUnknownCommand`, meaning the execute is redirected to the backend server. Instead, they should use `BrigadierCommand.Builder#permission`. This also has the advantage of passing the entire `CommandContext` instead of just the `CommandSource`.

We now clone the entire children tree on redirect, no matter what. `LiteralArgumentBuilder#redirect` breaks the argument-less command even if it is manually added to the builder. `LiteralCommandNode#createBuilder` performs a shallow clone with the same alias. Instead, we use our own builder and add the destination children nodes to the builder. This avoids expensive copying on build, and the _Available Commands_ packet size will be almost the same (if not better) thanks to compression.

astei:
Just to clarify your comment opening this to further review, the Brigadier `requires` behavior is very similar to the current behavior with `hasPermission`.

api/src/main/java/com/velocitypowered/api/command/BrigadierCommand.java
```diff
@@ -0,0 +1,57 @@
...
+public interface BrigadierCommand extends Command {
+
+  /**
+   * Provides a fluent interface to register a Brigadier command.
+   */
+  interface Builder extends Command.Builder<BrigadierCommand, Builder> {
+
+    /**
+     * Specifies the permission-checking predicate the {@link CommandSource}
+     * must pass to execute the command.
```
astei, line 25: Specify that you should use if you don't want the command to be sent to the backend server at all.

api/src/main/java/com/velocitypowered/api/command/CommandManager.java
```diff
@@ -1,87 +1,68 @@
...
   /**
-   * Unregisters a command.
+   * Unregisters the specified command alias from the manager, if registered.
    *
    * @param alias the command alias to unregister
    */
   void unregister(String alias);
 
   /**
-   * Calls CommandExecuteEvent and attempts to execute a command using the specified {@code cmdLine}
```
astei, line 39: This is still API-breaking...

build.gradle
astei: This is still live. Can you still revert this?

proxy/src/main/java/com/velocitypowered/proxy/command/builtin/ShutdownCommand.java
```diff
@@ -1,12 +1,10 @@
...
 
-import com.velocitypowered.api.command.Command;
-import com.velocitypowered.api.command.CommandSource;
+import com.velocitypowered.api.command.LegacyCommand;
 import com.velocitypowered.proxy.VelocityServer;
 import net.kyori.adventure.text.serializer.legacy.LegacyComponentSerializer;
-import org.checkerframework.checker.nullness.qual.NonNull;
 
-public class ShutdownCommand implements Command {
+public class ShutdownCommand implements LegacyCommand {
```
astei, line 7: I would probably make this a `RawCommand`

api/src/main/java/com/velocitypowered/api/command/BrigadierCommand.java
hugmanrique: My bad, didn't notice this was the previous behavior. Thought `hasPermission` being `false` still returned `true` from `execute` et al. As so, I see this method (and its accompanying implementation) a bit superfluous as the command itself can check the predicate.

hugmanrique: I'm thinking about the use cases for this, and letting Brigadier commands return a special code (as the implementation does right now) would be simpler and easier (the current injection is a mess).

hugmanrique: I added a `BrigadierCommand#FORWARD` so implementations can indicate a command execution should be forwarded to the backend server. This is more general and is not only limited to the previous permission-checking logic. Commands can now use the regular `ArgumentBuilder#requires` method to specify a predicate.

astei:
Without any further comments left to make here, I will merge this in. Thanks for your contribution @hugmanrique, it is greatly appreciated.

WesternPine:
That's huge, congrats @hugmanrique. Looks like a great implementation!

PR #334, opened 2020-07-27
Update VelocityServerConnection.java

robinbraemer:
If the serverAddress in `handshake` is empty, why do we prepend it to the HANDSHAKE_HOSTNAME_TOKEN?
Or should it be
```
handshake.setServerAddress(registeredServer.getServerInfo().getAddress().getHostString() + HANDSHAKE_HOSTNAME_TOKEN);
```


astei:
Fixed with a different, more correct patch. Thanks for catching this.

astei:
Build failure looks totally unrelated...

PR #338, opened 2020-07-29
Make initial read timeout configurable

xxDark:
Made it so initial timeout for connections is configurable.
Timeout is restored back later when user switches to PLAY state.

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/proxy/config/ProxyConfig.java
+++ b/api/src/main/java/com/velocitypowered/api/proxy/config/ProxyConfig.java
@@ -150,4 +150,12 @@ public interface ProxyConfig {
    * @return read timeout (in milliseconds)
    */
   int getReadTimeout();
+
+
+  /**
+   * Get how long this proxy will wait until performing an initial read timeout.
+   *
+   * @return read timeout (in milliseconds)
+   */
+  int getInitialReadTimeout();
 }
--- a/api/src/main/java/com/velocitypowered/api/proxy/config/ProxyConfig.java
+++ b/api/src/main/java/com/velocitypowered/api/proxy/config/ProxyConfig.java
@@ -151,7 +151,6 @@ public interface ProxyConfig {
    */
   int getReadTimeout();
 
-
   /**
    * Get how long this proxy will wait until performing an initial read timeout.
    *
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/client/LoginSessionHandler.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/client/LoginSessionHandler.java
@@ -25,6 +25,7 @@ import com.velocitypowered.proxy.config.PlayerInfoForwarding;
 import com.velocitypowered.proxy.config.VelocityConfiguration;
 import com.velocitypowered.proxy.connection.MinecraftConnection;
 import com.velocitypowered.proxy.connection.MinecraftSessionHandler;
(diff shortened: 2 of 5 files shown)


Gabik21:
In theory an idle tcp connection should not be using any resources beside a small memory footprint. If this yields a benefit in experiments, it feels like the actual issue is elsewhere. This is just a wild guess though.

MrIvanPlays:
> In theory an idle tcp connection should not be using any resources beside a small memory footprint. If this yields a benefit in experiments, it feels like the actual issue is elsewhere. This is just a wild guess though.

sure, 1 or 2 idle connections, no problems, but if so-called "bungeesmasher"s are used that bombard you with say 100 connections per second you need to quickly drop them or else what's gonna happen is that the proxy will crash. this is both present on velocity and bungeecord with slightly different results: a so-called "bungeesmasher" would also exploit the explicit exception usage in bungeecord, fortunately velocity doesn't have that problem.

astei:
> but if so-called "bungeesmasher"s are used that bombard you with say 100 connections per second you need to quickly drop them

That is an attack that really needs to be mitigated at a higher layer. It's much more expensive for the proxy to block (or drop) the connections than have the kernel deal with it (context switching is very expensive). Velocity can (and should) guard against the Layer 7 attacks it can (which has been my primary focus with my recent work), but if someone is spamming the proxy with connections, the proxy is ill-equipped to deal with the problem.

MrIvanPlays:
> > but if so-called "bungeesmasher"s are used that bombard you with say 100 connections per second you need to quickly drop them
> 
> That is an attack that really needs to be mitigated at a higher layer. It's much more expensive for the proxy to block (or drop) the connections than have the kernel deal with it (context switching is very expensive). Velocity can (and should) guard against the Layer 7 attacks it can (which has been my primary focus with my recent work), but if someone is spamming the proxy with connections, the proxy is ill-equipped to deal with the problem.

Fully agree with you, unfortunately they can't stop **all** of the attack, that's why we also need some security layers here.

api/src/main/java/com/velocitypowered/api/proxy/config/ProxyConfig.java
```diff
@@ -150,4 +150,12 @@
    * @return read timeout (in milliseconds)
    */
   int getReadTimeout();
+
+
```
astei, line 154: Extra line here

proxy/src/main/java/com/velocitypowered/proxy/connection/client/LoginSessionHandler.java
```diff
@@ -198,6 +201,12 @@ private EncryptionRequest generateEncryptionRequest() {
   }
 
   private void initializePlayer(GameProfile profile, boolean onlineMode) {
+    // Restore normal timeout for connection.
+    mcConnection.getChannel().pipeline()
```
Mystiflow, line 205: Can you do something like mcConnection.getChannel().pipeline().get(Connections.READ_TIMEOUT)**.setTimeout(server.getConfiguration().getReadTimeout()) so maybe the re-instantiation of ReadTimeoutHandler isn't needed?

MrIvanPlays: I don't think such nanooptimization is useful, the gc would catch the non-used object right away.

astei:
While I'm not a big fan of this solution (it really has to be solved at a layer higher than Velocity), it's better than nothing and does add some additional mitigation to this area.

PR #339, opened 2020-07-29
Actually solve framing issues.

xxDark:
This was initialy reported in https://github.com/netty/netty/issues/10417
In fact, the problem is that other decoders does not know about frame-decoder, so it continues to read. This PR makes it so packet decoder notifies varint decoder about fatal errors, causing it to stop reading.

See https://github.com/SpigotMC/BungeeCord/pull/2908#issuecomment-665840788 for results


xxDark:
Will close this now until I more investigate.

PR #340, opened 2020-07-30
Use channel factories to create new channels

xxDark:


proxy/src/main/java/com/velocitypowered/proxy/network/TransportType.java
```diff
@@ -18,30 +20,57 @@
...
       (name, type) -> new NioEventLoopGroup(0, createThreadFactory(name, type))),
-  EPOLL("epoll", EpollServerSocketChannel.class, EpollSocketChannel.class,
-      EpollDatagramChannel.class,
+  EPOLL("epoll", EpollServerSocketChannel::new, EpollServerSocketChannel.class,
+      EpollSocketChannel::new, EpollSocketChannel.class,
+      EpollDatagramChannel::new, EpollDatagramChannel.class,
       (name, type) -> new EpollEventLoopGroup(0, createThreadFactory(name, type)));
 
   final String name;
+  @Deprecated
```
astei, line 33: As an internal class we don't need to deprecate these

astei: Just remove them in this case.

PR #343, opened 2020-07-31
Avoid lists copying

xxDark:


proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/PlayerListItem.java
```diff
@@ -25,11 +25,16 @@
...
-  private final List<Item> items = new ArrayList<>();
+  private List<Item> items;
 
   public PlayerListItem(int action, List<Item> items) {
     this.action = action;
-    this.items.addAll(items);
+    (this.items = new ArrayList<>(items.size())).addAll(items);
+  }
+
+  public PlayerListItem(Void unused, int action, List<Item> items) {
```
Gabik21, line 35: This looks like a really odd hack to distinct the two constructors. I think it must be settled whether or not we make a defensive copy to prevent side-effects.

proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/PlayerListItem.java
```diff
@@ -25,11 +25,16 @@
   public static final int UPDATE_DISPLAY_NAME = 3;
   public static final int REMOVE_PLAYER = 4;
   private int action;
-  private final List<Item> items = new ArrayList<>();
+  private List<Item> items;
 
   public PlayerListItem(int action, List<Item> items) {
     this.action = action;
-    this.items.addAll(items);
+    (this.items = new ArrayList<>(items.size())).addAll(items);
```
hugmanrique, line 32: Can be simplified to `new ArrayList<>(items)`

proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/PlayerListItem.java
hugmanrique: Adding to this, would remove this constructor and pass an empty list to the previous constructor instead. I don't think these kinds of micro-optimizations yield observable performance gains.

Mystiflow: Why was it done like this instead of just using a boolean 'cloneList' for example?

astei:
I'm simply not convinced these changes will do much to improve efficiency. At best, they are micro-optimizations on non-critical paths.

I could accept a more tailored PR focusing on `PlayerListItem` but that's it.

proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/PlayerListItem.java
lineargraph: I'm not sure, but maybe a static method could be a more descriptive interface

astei:
Closing due to a lack of response

PR #346, opened 2020-08-03
Add PRE_SERVER_JOIN to DisconnectEvent#LoginStatus

STG-Allen:


proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
```diff
@@ -695,8 +695,12 @@ void teardown() {
 
     DisconnectEvent.LoginStatus status;
     if (connectedPlayer.isPresent()) {
-      status = connectedPlayer.get() == this ? LoginStatus.SUCCESSFUL_LOGIN
-          : LoginStatus.CONFLICTING_LOGIN;
+      if (!connectedPlayer.get().getCurrentServer().isPresent()) {
```
MrIvanPlays, line 698: Maybe make connectedPlayer.get() to a variable and use that instead of explicitly calling .get

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
```diff
@@ -695,8 +695,12 @@ void teardown() {
...
+        status = LoginStatus.PRE_SERVER_JOIN;
+      } else {
+        status = connectedPlayer.get() == this ? LoginStatus.SUCCESSFUL_LOGIN
```
MrIvanPlays, line 701: See L698

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
STG-Allen: Fixed in 4adc859cf11c62319f637d090bf0bc8d11f23ae8

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
STG-Allen: Fixed in 4adc859cf11c62319f637d090bf0bc8d11f23ae8

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
```diff
@@ -690,13 +690,17 @@ void teardown() {
...
 
-    Optional<Player> connectedPlayer = server.getPlayer(this.getUniqueId());
     server.unregisterConnection(this);
 
     DisconnectEvent.LoginStatus status;
-    if (connectedPlayer.isPresent()) {
-      status = connectedPlayer.get() == this ? LoginStatus.SUCCESSFUL_LOGIN
-          : LoginStatus.CONFLICTING_LOGIN;
+    if (server.getPlayer(this.getUniqueId()).isPresent()) {
+      Player connectedPlayer = server.getPlayer(this.getUniqueId()).get();
```
hugmanrique, line 697: Arguably this is worse than before (even if player lookups are constant). I disagree with @MrIvanPlays, previous code was fine.

STG-Allen: I'll just revert it then

astei: This revised code actually introduces a bug.

STG-Allen: Yeah, figured such would occur. Will do as i stated previously

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
```diff
@@ -690,13 +690,17 @@ void teardown() {
       connectedServer.disconnect();
     }
 
-    Optional<Player> connectedPlayer = server.getPlayer(this.getUniqueId());
     server.unregisterConnection(this);
 
     DisconnectEvent.LoginStatus status;
+    Optional<Player> connectedPlayer = server.getPlayer(this.getUniqueId());
     if (connectedPlayer.isPresent()) {
```
astei, line 697: This is still incorrect. Now the successful login case is never fired. Your first revision was correct.

To fix this just put this line back where it was - before the unregisterConnection call.

astei: I commented on the wrong line, the offending line is above this one. I was reviewing this from my phone. Whoops!

STG-Allen: Should be fixed now.

PR #350, opened 2020-08-05
Respect nodes' requirements

MrIvanPlays:
Only send the nodes to the player for which he meets the requirement.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/backend/BackendPlaySessionHandler.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/backend/BackendPlaySessionHandler.java
@@ -172,23 +172,19 @@ public class BackendPlaySessionHandler implements MinecraftSessionHandler {
     RootCommandNode<CommandSource> rootNode = commands.getRootNode();
     if (server.getConfiguration().isAnnounceProxyCommands()) {
       // Inject commands from the proxy.
-      Collection<CommandNode<CommandSource>> proxyNodes = server.getCommandManager().getDispatcher()
-              .getRoot().getChildren();
+      RootCommandNode<CommandSource> dispatcherRootNode =
+          (RootCommandNode<CommandSource>)
+              filterNode(
+                  server.getCommandManager().getDispatcher().getRoot(), new IdentityHashMap<>());
+      Collection<CommandNode<CommandSource>> proxyNodes = dispatcherRootNode.getChildren();
       for (CommandNode<CommandSource> node : proxyNodes) {
         rootNode.addChild(node);
       }
     }
 
-    server
-        .getEventManager()
-        .fire(new PlayerAvailableCommandsEvent(serverConn.getPlayer(), rootNode))
-        .thenAcceptAsync(
-            event -> {
-              commands.setRootNode(
-                  (RootCommandNode<CommandSource>) filterNode(rootNode, new IdentityHashMap<>()));
-              playerConnection.write(commands);
-            },
-            playerConnection.eventLoop());
+    server.getEventManager().fire(
+        new PlayerAvailableCommandsEvent(serverConn.getPlayer(), rootNode))
+        .thenAcceptAsync(event -> playerConnection.write(commands), playerConnection.eventLoop());
     return true;
   }
 
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/backend/BackendPlaySessionHandler.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/backend/BackendPlaySessionHandler.java
@@ -30,8 +30,6 @@ import io.netty.buffer.ByteBufUtil;
 import io.netty.buffer.Unpooled;
 import io.netty.handler.timeout.ReadTimeoutException;
 import java.util.Collection;
-import java.util.IdentityHashMap;
-import java.util.Map;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
@@ -174,8 +172,7 @@ public class BackendPlaySessionHandler implements MinecraftSessionHandler {
       // Inject commands from the proxy.
       RootCommandNode<CommandSource> dispatcherRootNode =
           (RootCommandNode<CommandSource>)
-              filterNode(
-                  server.getCommandManager().getDispatcher().getRoot(), new IdentityHashMap<>());
+              filterNode(server.getCommandManager().getDispatcher().getRoot());
       Collection<CommandNode<CommandSource>> proxyNodes = dispatcherRootNode.getChildren();
       for (CommandNode<CommandSource> node : proxyNodes) {
         rootNode.addChild(node);
@@ -193,12 +190,9 @@ public class BackendPlaySessionHandler implements MinecraftSessionHandler {
    * by the player (respecting the requirement of the node).
    *
    * @param source source node
-   * @param nodeMapping mapped nodes
    * @return filtered node
    */
-  private CommandNode<CommandSource> filterNode(
-      CommandNode<CommandSource> source,
-      Map<CommandNode<CommandSource>, CommandNode<CommandSource>> nodeMapping) {
+  private CommandNode<CommandSource> filterNode(CommandNode<CommandSource> source) {
     CommandNode<CommandSource> dest;
     if (source instanceof RootCommandNode) {
       dest = new RootCommandNode<>();
@@ -206,7 +200,6 @@ public class BackendPlaySessionHandler implements MinecraftSessionHandler {
       if (source.getRequirement() != null) {
         try {
           if (!source.getRequirement().test(serverConn.getPlayer())) {
-            nodeMapping.put(source, null);
             return null;
           }
         } catch (Throwable e) {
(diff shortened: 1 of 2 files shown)


proxy/src/main/java/com/velocitypowered/proxy/connection/backend/BackendPlaySessionHandler.java
```diff
@@ -173,12 +179,72 @@ public boolean handle(AvailableCommands commands) {
...
 
+  /**
+   * Creates a deep copy of the provided command node, but removes any node that are not accessible
+   * by the player (respecting the requirement of the node).
+   *
+   * @param source source node
+   * @param nodeMapping mapped nodes
+   * @return filtered node
+   */
+  private CommandNode<CommandSource> filterNode(
```
hugmanrique, line 203: This only works for `BrigadierCommand`s, I couldn't think of a way to implement it for simple and raw commands that expect a Context object (creating the context manually is hacky at best)

hugmanrique: Sorry if I'm missing something, but why do we need to treat redirects as a special case? See `BrigadierUtils` for a deep clone method example.

proxy/src/main/java/com/velocitypowered/proxy/connection/backend/BackendPlaySessionHandler.java
```diff
@@ -173,12 +179,72 @@ public boolean handle(AvailableCommands commands) {
...
 
-    server.getEventManager().fire(
-        new PlayerAvailableCommandsEvent(serverConn.getPlayer(), rootNode))
-        .thenAcceptAsync(event -> playerConnection.write(commands), playerConnection.eventLoop());
+    server
+        .getEventManager()
+        .fire(new PlayerAvailableCommandsEvent(serverConn.getPlayer(), rootNode))
+        .thenAcceptAsync(
+            event -> {
+              commands.setRootNode(
```
hugmanrique, line 187: To avoid unnecessarily processing the backend server commands (which don't have a requires  predicate), I would move this filtering to https://github.com/MrIvanPlays/Velocity/blob/dfce2ea2d39d9eb6350db8705c07c705b22d4a52/proxy/src/main/java/com/velocitypowered/proxy/connection/backend/BackendPlaySessionHandler.java#L175 (just above this, where the proxy commands are injected). This also avoids the cloning if the `announceProxyCommands` config is set to `false`

hugmanrique:
Otherwise, this LGTM 😀

proxy/src/main/java/com/velocitypowered/proxy/connection/backend/BackendPlaySessionHandler.java
MrIvanPlays: I can also implement it on the simple/raw commands, it is just command completion, but this method is only for `BrigadierCommand`s

proxy/src/main/java/com/velocitypowered/proxy/connection/backend/BackendPlaySessionHandler.java
MrIvanPlays: If I move it there, the nodes that the event adds wouldn't be captured and the method wouldn't work as intended. 
Additionally, we can remove this event and enforce `BrigadierCommand` if someone wants to modify the root node.

hugmanrique: Imo, nodes added via `PlayerAvailableCommandsEvent` should always be sent, even if the source doesn't pass the predicate.

proxy/src/main/java/com/velocitypowered/proxy/connection/backend/BackendPlaySessionHandler.java
MrIvanPlays: > why do we need to treat redirects as a special case

We need to filter the redirects too

proxy/src/main/java/com/velocitypowered/proxy/connection/backend/BackendPlaySessionHandler.java
MrIvanPlays: > Imo, nodes added via `PlayerAvailableCommandsEvent` should always be sent, even if the source doesn't pass the predicate.

how come

proxy/src/main/java/com/velocitypowered/proxy/connection/backend/BackendPlaySessionHandler.java
hugmanrique: The problem is simple and raw commands don't use the requires predicate since it doesn't pass the used command and thus the context object cannot be created.

proxy/src/main/java/com/velocitypowered/proxy/connection/backend/BackendPlaySessionHandler.java
hugmanrique: If a plugin forcefully adds a command node to the tree, it should be sent to the client regardless of the `requires` return. The plugin can register a command to use this mechanic. The event is mostly intended to remove or edit nodes.

proxy/src/main/java/com/velocitypowered/proxy/connection/backend/BackendPlaySessionHandler.java
MrIvanPlays: > since it doesn't pass the used command

wdym

hugmanrique:
Also, what happens if a permission is assigned to a player (e.g. using LuckPerms)? As it stands, the player would need to relog to be aware of the new available commands.

proxy/src/main/java/com/velocitypowered/proxy/connection/backend/BackendPlaySessionHandler.java
MrIvanPlays: > If a plugin forcefully adds a command node to the tree, it should be sent to the client regardless of the `requires` return. The plugin can register a command to use this mechanic. The event is mostly intended to remove or edit nodes.

Sure, I will move the filtering then

proxy/src/main/java/com/velocitypowered/proxy/connection/backend/BackendPlaySessionHandler.java
hugmanrique: See https://github.com/VelocityPowered/Velocity/blob/76173e414506f361c5acc60318a16bcede44edbb/proxy/src/main/java/com/velocitypowered/proxy/command/CommandNodeFactory.java#L36

MrIvanPlays:
> Also, what happens if a permission is assigned to a player (e.g. using LuckPerms)? As it stands, the player would need to relog to be aware of the new available commands.

That's the minus of being a proxy. The client requests the commands once, but the packet doesn't transfer data about the requirements. Bukkit handles that thru the command completions, what I can see is that velocity doesn't have any control of the `/<command>` suggestions as it forwards that to brigadier's CommandDispatcher which as I see (may be wrong) the method velocity calls doesn't respect the requirement.

hugmanrique:
Yeah, thinking about it currently the same problem occurs when a permission node is removed from a player.

Regarding the second point, Bukkit commands have a top-level permission node that's used by the predicate to check if the player has permission. As so, it can easily filter commands. Velocity (and Brigadier) is unopinionated, and doesn't force commands to have a single permission node and instead accepts a custom predicate that takes the argument list as an argument (inside the context). The problem is Brigadier's predicate only gives us access to the executing source and ignores the arguments (rightly so, because `requires` works on a node-per-node basis).

proxy/src/main/java/com/velocitypowered/proxy/connection/backend/BackendPlaySessionHandler.java
MrIvanPlays: Can't we make the command to forward only on failure? So that we can utilise .requires for better experience?
(asking u before touching anything of that code, I'm scared 👀)

astei:
We could modify the new non-Brigadier command APIs (`SimpleCommand` / `RawCommand`) instead, such that they work like Brigadier, then introduce some kind of `CommandInvocationAction` on `execute()` - how does this sound?

MrIvanPlays:
> We could modify the new non-Brigadier command APIs (`SimpleCommand` / `RawCommand`) instead, such that they work like Brigadier, then introduce some kind of `CommandInvocationAction` on `execute()` - how does this sound?

Actually we don't need the CommandInvocationAction - we can have a simple boolean return (if we don't have that already) for which if returns false we treat it as a failure and redirect the command to the backend server

MrIvanPlays:
Let me scratch on something and you will tell me your opinion about it okay? 
Or if you want we can hop on discord to have a talk (but I'm banned from urs :eyes:)

hugmanrique:
This is how the backend redirect works currently. If you take a look at the previously mentioned code, they return `BrigadierCommand.FORWARD` to indicate the command should be forwarded to the backend server. This return code is handled by `VelocityCommandManager#executeImmediately`

MrIvanPlays:
> However, that doesn't fix the original issue. `hasPermission` still expects a context object including the arguments, and removing it would be a huge breaking change.

Well then idk what I should do about it. Any other changes you can suggest about this pr? Otherwise it is finished.
I don't want to make a decision that would break the api just to have that functionality. This pr works only for brigadier commands. Honestly idk why did you implement other commands at first place. You could just keep the old stuff in order to not break plugins and add the brigadier commands as the new api.

hugmanrique:
Some plugins prefer the Bukkit-style commands or **cannot use** Brigadier (in particular, Lucko complained about this in discord while reviewing the new command API). This includes plugins using alternative command frameworks such as ACF or CommandBook.

However, as noted in the `Command` javadoc Brigadier commands are now preferred so hopefully adoption will increase (and will benefit from this PR).

I won't have a PC for some days so cannot check the redirect `requires` logic, but still believe the map is unnecessary and a `List` containing the removed nodes and checking redirect nodes have that destination (and if so, remove it as well) would be sufficient.

MrIvanPlays:
> but still believe the map is unnecessary

Tested & removed. Works fine without it, though it could be better to perform more tests with a command that has a redirect, cuz the commands I test with don't use redirects.

MrIvanPlays:
Everything is finished.
@astei your turn

PR #356, opened 2020-08-18
Make hardcoded messages configurable

FrankHeijden:


FrankHeijden:
Travis build failed because https://github.com/VelocityPowered/Velocity/pull/358 needs to be merged first.

proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
```diff
@@ -781,4 +791,62 @@ public boolean isFromConfig() {
...
+    private String onlineModeOnly = "&cThis server only accepts connections from online-mode clients."
+            + "\n\n&7Did you change your username? Sign out of Minecraft, sign back in, and try again.";
+    private String noAvailableServers = "&cThere are no available servers.";
+    private String alreadyConnected = "&cYou are already connected to this proxy!";
+    private String movedToNewServerPrefix = "&cThe server you were on kicked you: ";
+    private String genericConnectionError = "&cAn internal error occurred in your connection.";
+
+    private Messages(CommentedConfig toml) {
+      if (toml != null) {
+        this.kickPrefix = toml.getOrElse("kick-prefix", kickPrefix);
```
Mystiflow, line 807: I'm not sure I like the idea of defaulting by hardcoding it in the source. It feels like a maintenance burding having to edit both this and the config and may cause them getting out of sync in future updates.

astei: The configuration-loading code is quite a mess, it's in dire need of a refactor - at least to use an object-mapping solution with defaults. I'll have to investigate if Night-Config will allow us, or explore writing my own TOML parser/serializer and slipping that into Configurate 4 (since zml told me he's not interested).

FrankHeijden: I right now changed this behaviour by creating a temporary default config copied from the jar file, which is used to retrieve the default values from.

proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
```diff
@@ -384,15 +398,28 @@ public String toString() {
...
-        .defaultResource("/default-velocity.toml")
+        .defaultResource(defaultResource)
         .autosave()
         .preserveInsertionOrder()
         .sync()
         .build();
     config.load();
 
+    // Create temporary default configuration
+    File tmpFile = File.createTempFile(defaultResource, null);
```
astei, line 412: Yikes, this is what I feared. Maybe I need to work on a proper TOML serializer for Configurate!

PR #357, opened 2020-08-19
Skip unnecessary decompression of compressed packet from server/client

CatCoderr:
If compression of connection is enabled, we try to decompress first 5 bytes (max Varint length) in the input packet as its ID. Compressed packet will be written to client/server directly without unnecessary decompression if ID of the packet is not registered in packet registry. For example, large packets such as chunk data and entity metadata will be skipped in most cases. _Tested on Minecraft 1.16.2._

Debug logs:
image

### DIFF
--- a/native/src/main/java/com/velocitypowered/natives/compression/JavaVelocityCompressor.java
+++ b/native/src/main/java/com/velocitypowered/natives/compression/JavaVelocityCompressor.java
@@ -16,12 +16,14 @@ public class JavaVelocityCompressor implements VelocityCompressor {
   private final Deflater deflater;
   private final Inflater inflater;
   private final byte[] buf;
+  private final byte[] varintBuffer;
   private boolean disposed = false;
 
   private JavaVelocityCompressor(int level) {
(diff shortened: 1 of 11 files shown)


hugmanrique:
Some servers prefer to handle all packet decompression at the proxy level (due to its better performance). Perhaps a config option could be added to preserve current behavior?

Imo velocity-natives should stay protocol-agnostic and shouldn't assume the data to decompress is a packet starting with a varint id.

astei:
This PR is currently -1 for me:

* The recommended setup is to have backend servers send uncompressed data to the proxy and let it compress data for you.
* This is targeted to the wrong branch, it should target `dev/1.1.0` as that has new feature development.
* Even if it _were_ targeted to the right branch, it won't work with Velocity 1.1.0 because we switched to libdeflate, which doesn't support streaming (it is entirely block based). I'm not sure supporting this feature is worth having to drop libdeflate and revert to using zlib.
* In addition you seem to have applied a few unrelated style changes.

PR #361, opened 2020-08-24
Handle IPOther

MrIvanPlays:
BungeeCord has recently added IPOther special message. This PR
ports the handling behind it in velocity.


MrIvanPlays:
Could've requested changes and I'd implement them, but sure this works too.

PR #365, opened 2020-09-14
Added lowercase to invalid plugin id message

STG-Allen:
added "be lowercase" to the invalid plugin id message


alexstaeding:
I think you missed a space before “be”

astei:
We can deal with it later. Not a deal breaker.

PR #371, opened 2020-10-13
Handle unknown commands properly

MrIvanPlays:
Previously, the check for unknown commands was done if a CommandSyntaxException with dispatcherUnknownCommand was thrown, which as expected, forwarded the command to the backend. This is a hacky way to handle unknown commands (and CommandSyntaxExceptions). My changes did it so it checks for a child first, so every CommandSyntaxException is sent as a message to the sender and only unknown commands to be forwarded to the server.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/command/VelocityCommandManager.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/command/VelocityCommandManager.java
@@ -155,8 +155,7 @@ public class VelocityCommandManager implements CommandManager {
     Preconditions.checkNotNull(cmdLine, "cmdLine");
 
     String normalizedInput = BrigadierUtils.normalizeInput(cmdLine, true);
-    String[] args = normalizedInput.split(" ");
-    if (dispatcher.getRoot().getChild(args[0]) == null) {
+    if (dispatcher.getRoot().getChild(BrigadierUtils.getCommandName(normalizedInput)) == null) {
       return false;
     }
     ParseResults<CommandSource> results = dispatcher.parse(normalizedInput, source);
@@ -209,8 +208,7 @@ public class VelocityCommandManager implements CommandManager {
     Preconditions.checkNotNull(source, "source");
     Preconditions.checkNotNull(cmdLine, "cmdLine");
 
-    ParseResults<CommandSource> parse =
-        dispatcher.parse(BrigadierUtils.normalizeInput(cmdLine, false), source);
+    ParseResults<CommandSource> parse = dispatcher.parse(cmdLine, source);
     return dispatcher.getCompletionSuggestions(parse)
             .thenApply(suggestions -> Lists.transform(suggestions.getList(), Suggestion::getText));
   }
--- a/proxy/src/main/java/com/velocitypowered/proxy/command/VelocityCommandManager.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/command/VelocityCommandManager.java
@@ -25,7 +25,6 @@ import java.util.List;
 import java.util.Locale;
 import java.util.concurrent.CompletableFuture;
 import net.kyori.adventure.text.Component;
-import net.kyori.adventure.text.TextComponent;
 import net.kyori.adventure.text.format.NamedTextColor;
 
 public class VelocityCommandManager implements CommandManager {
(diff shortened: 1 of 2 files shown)


Proximyst (changes requested):
The command forwarding makes sense. This should check if the command *wants* to be forwarded (i.e. add new API).

MrIvanPlays:
> The command forwarding makes sense. This should check if the command _wants_ to be forwarded (i.e. add new API).

you can just return BrigadierCommand.FORWARD if you want to forward it

hugmanrique:
The explanation that follows is similar to that on https://github.com/VelocityPowered/Velocity/issues/369#issuecomment-703304786. A paradigm shift that came from the introduction of Brigadier is that command labels are no longer more important than the arguments. They are (and imo should be) handled equally by Brigadier (and therefore Velocity). `BrigadierCommand`s are now the preferred choice when writing commands, and if a required argument isn't recognized, then the command is forwarded to the backend server. This is the same as how command labels are and were handled prior to the introduction of Brigadier. If you want to handle arbitrary arguments, add a `StringArgumentType` to the tree, or use `SimpleCommand` or `RawCommand`.

Additionally, this is a breaking change and should be submitted to Velocity 2.0.

proxy/src/main/java/com/velocitypowered/proxy/command/VelocityCommandManager.java
```diff
@@ -211,17 +209,12 @@ public boolean executeImmediately(final CommandSource source, final String cmdLi
     Preconditions.checkNotNull(source, "source");
     Preconditions.checkNotNull(cmdLine, "cmdLine");
 
-    ParseResults<CommandSource> parse = parse(cmdLine, source, false);
+    ParseResults<CommandSource> parse =
```
hugmanrique, line 211: The command line is intentionally not normalized for tab completion (spaces matter).

proxy/src/main/java/com/velocitypowered/proxy/command/VelocityCommandManager.java
```diff
@@ -155,19 +154,18 @@ public boolean executeImmediately(final CommandSource source, final String cmdLi
...
-      if (isSyntaxError) {
-        source.sendMessage(Component.text(e.getMessage(), NamedTextColor.RED));
-        // This is, of course, a lie, but the API will need to change...
-        return true;
-      } else {
-        return false;
-      }
+      source.sendMessage(Component.text(e.getMessage(), NamedTextColor.RED));
+      // This is, of course, a lie, but the API will need to change...
+      return true;
```
hugmanrique, line 167: I agree returning `true` on a syntax exception is reasonable since the API docs say "@return `true` if the command was found and executed". Indeed, the command wasn't executed but the impl uses the return as an indication to forward the command. Maybe update the API javadoc instead?

proxy/src/main/java/com/velocitypowered/proxy/command/VelocityCommandManager.java
```diff
@@ -155,19 +154,18 @@ public boolean executeImmediately(final CommandSource source, final String cmdLi
     Preconditions.checkNotNull(source, "source");
     Preconditions.checkNotNull(cmdLine, "cmdLine");
 
-    ParseResults<CommandSource> results = parse(cmdLine, source, true);
+    String normalizedInput = BrigadierUtils.normalizeInput(cmdLine, true);
+    String[] args = normalizedInput.split(" ");
```
hugmanrique: Instead of splitting, we can do the complement of https://github.com/VelocityPowered/Velocity/blob/e8abdba4a6f54d1fa0a4d793d994adebdf48957c/proxy/src/main/java/com/velocitypowered/proxy/util/BrigadierUtils.java#L85 to avoid the array allocation + entire string lookup.

hugmanrique (changes requested):

MrIvanPlays:
@hugmanrique idk but seems like you didnt understand the point of this pull request. 

What my target is here is for unknown to velocity commands, so instead of doing unnecessary work and load, skip all handling if it is an unknown command, which forwards it to the backend server. this won't solve proximyst's original interest, which (in my eyes) is to have a special case command handling for CommandSyntaxExceptions. 

this is changing the behavior _only_ for unknown literals or arguments (argument types) of known commands, it now instead of forwarding it to the server simply sends the message from the exception.

proxy/src/main/java/com/velocitypowered/proxy/command/VelocityCommandManager.java
MrIvanPlays: I'm scared of changing any of the docs so im not touching anything there

hugmanrique:
Oh yeah, my bad, read `dispatcherUnknownArgument` instead of `dispatcherUnknownCommand`, I'm tired.

Still, this assumes Brigadier uses `RootCommandNode#getChild` to look up commands, which isn't the case https://github.com/Mojang/brigadier/blob/master/src/main/java/com/mojang/brigadier/CommandDispatcher.java#L356, and even if it were, it's bad to rely on implementation details. Imo it's better to let Brigadier do the parsing and let Velocity do the exception handling.

proxy/src/main/java/com/velocitypowered/proxy/util/BrigadierUtils.java
```diff
@@ -91,6 +91,20 @@ public static String getRawArguments(final CommandContext<CommandSource> context
...
+  /**
+   * Returns the command name from the specified input.
+   *
+   * @param input command input
+   * @return command name from the specified input
+   */
+  public static String getCommandName(final String input) {
+    int firstSpace = input.indexOf(' ');
+    if (firstSpace == -1) {
+      return input;
```
hugmanrique: Missing `.toLowerCase(Locale.ENGLISH)`, command aliases are case insensitive (Brigadier isn't).

proxy/src/main/java/com/velocitypowered/proxy/util/BrigadierUtils.java
```diff
@@ -91,6 +91,20 @@ public static String getRawArguments(final CommandContext<CommandSource> context
...
+    }
+    return input.substring(0, firstSpace);
```
hugmanrique: Same as above

hugmanrique (changes requested):

proxy/src/main/java/com/velocitypowered/proxy/command/VelocityCommandManager.java
```diff
@@ -211,17 +208,11 @@ public boolean executeImmediately(final CommandSource source, final String cmdLi
     Preconditions.checkNotNull(source, "source");
     Preconditions.checkNotNull(cmdLine, "cmdLine");
 
-    ParseResults<CommandSource> parse = parse(cmdLine, source, false);
+    ParseResults<CommandSource> parse = dispatcher.parse(cmdLine, source);
```
hugmanrique: I'm really tired, sorry for wasting your time, previous code was fine.

MrIvanPlays:
I find checking the CommandSyntaxException a hacky way instead of doing it that way. Plus, we're kinda saving resources this way.

hugmanrique:
While talking on discord, we figured the command will be found via `getChild` even if the player has no permission to run it. Parsing it with Brigadier ensures the syntax exception is thrown if the player has no permission so that the command is forwarded.

MrIvanPlays:
After some researching we've found a big issue with that change; we're closing it

Issue #374, opened 2020-10-16
CompletableFuture potentially swallowing exceptions


A248:
A link Crypnotic posted to LoginSessionHandler caught my attention. From what I can tell, the use of CompletableFuture in LoginSessionHandler has the potential to swallow exceptions.

https://github.com/VelocityPowered/Velocity/blob/fd6394517595e291e95bca5f60cb1a68bc18c8de/proxy/src/main/java/com/velocitypowered/proxy/connection/client/LoginSessionHandler.java#L255

thenAccept(Async), thenRun(Async), and all other transformation methods on CompletableFuture not related to exception handling store any thrown unchecked exceptions in the returned future wrapped in CompletionException, but if the using code discards this future, the stored exception is lost.

astei:
That is not really an issue, since the fire method catches all the exceptions and logs them, never throwing an exception. The issue he is having is completely different.

A248:
No, I'm suggesting it is proxy code, not event listeners, which may throw an unchecked exception, inside the thenAcceptAsync callback.

astei:
I will need proof that a problem actually exists.

A248:
```java
	@Test
	public void testSwallowExceptions() {
		java.util.concurrent.CompletableFuture.runAsync(() -> {
			System.out.println("Pre throw");
			// Can also throw here and achieve the same result
		}).thenRunAsync(() -> {
			System.out.println("Throwing...");
			throw new RuntimeException();
		});
	}
```

I'm not saying there are any exceptions being swallowed at the moment, merely that if an unchecked exception were to be thrown, it would not be detected.

The key thing to note is these are futures which are neither awaited nor have any dependent actions.

Gabik21:
Took me an eternity to debug one of these. It is very frustrating that CompletableFutures behave that way. It looks so ugly to glue that block to every one of these:
```java
        .exceptionally(throwable -> {
          throwable.printStackTrace();
          return null;
        });
```

PR #376, opened 2020-10-26
Handle exceptions relating to CompletableFuture operations

A248:
Solves #374 by adding `exceptionally` where CompletableFutures are not otherwise chained or returned. The last future in a chain, if it is not awaited, must have some sort of exception handler.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
@@ -425,8 +425,11 @@ public class VelocityServer implements ProxyServer, ForwardingAudience {
                   .toArray((IntFunction<CompletableFuture<Void>[]>) CompletableFuture[]::new));
 
           playersTeardownFuture.get(10, TimeUnit.SECONDS);
-        } catch (TimeoutException | ExecutionException e) {
+        } catch (TimeoutException e) {
           timedOut = true;
+        } catch (ExecutionException e) {
+          timedOut = true;
+          logger.error("Exception while tearing down player connections", e);
         }
 
         eventManager.fireShutdownEvent();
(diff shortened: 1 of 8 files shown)


astei (changes requested):
You have failed to convince me there exists a real problem to be addressed here, something I  brought up in #374. You've convinced me the pitfall _exists_ but you've failed to convince me that this quirk of `CompletableFuture` is an actual problem as it is used in Velocity, particularly when paired with `fire()` (which, I have checked, and unless there is some sort of bizarro `Throwable` getting thrown, there is no exception thrown). This, in my opinion, pollutes all async code with a condition to deal with a concern that, in my eyes, is currently _purely theoretical_.

Come back to me with a _real_ problem that this patch fixes and I will definitely merge this in.

A248:
Not following best practices is a real problem, one which is as real as the bugs it leads to.

Consider:

```java
try {
  // some operation
} catch (Throwable ignored) {}
```

You don't write synchronous code this way. It would be ludicrous. Even if you knew your current code did not throw any exceptions, it makes no sense to swallow them.

What happens when you change the body of the try block, and it starts to throw? You'll receive little indication something is wrong aside from the fact that bits and pieces of your code are no longer running, because it abruptly and silently terminated due to an exception. Or, what happens if your code doesn't throw an exception most of the time, but there are those edge cases and race conditions in which it does? Either way, hiding all Throwables is a terrible way to go.

The equivalent of the empty catch block in asynchronous code is failing to attach an exception handler to the last CompletableFuture in a chain:

```java
CompletableFuture.runAsync(() -> {
  // some operation
});
```

This is just as deadly and just as potent as a synchronous empty catch block. I would add just as hard to debug, except that it is *harder* to debug considering the nature of debugging asynchronous interactions.

It's a little bit of extra code on the tail end of your CompletableFutures to prevent a whole host of silent future problems.

astei (approved):
I'll merge this in but I still question the need for it.

astei:
You were right, I was wrong. Some legitimate bugs were uncovered by this. Thank you so much @A248.

Issue #378, opened 2020-10-28
Rework the event API


astei:
The Velocity event API, as currently designed, has some major flaws in its design that we should aim to correct for Velocity Polymer.

## Handling of potentially blocking events

**Status**: Implemented with #384, thanks @Cybermaxke! It works essentially as I described, though the terminology is a bit different.

The original design choice for dealing with events that might potentially take too long in Velocity 1.x.x was to offload all event execution to a separate thread pool and execute the event handlers there. Unfortunately it is quite possible for this thread pool to be easily exhausted, Switching from the current fixed-size thread pool to a cached thread pool may help a little here, but it still poses issues. It also introduces context switching which may not be ideal for some users.

My current proposed solution is:

* All events will now fire on the event loop initially. This means relatively lightweight listeners can run on the Netty event loop and reduce latency.
* If there is a possibility for an event listener to do a task that can take some time, the event handler can choose to pause bubbling the event, perform the logic, and then resume bubbling the event.

## Concrete implementations of events

**Status**: zml needs to do some refactoring for this, but [event-impl-gen](https://github.com/SpongePowered/event-impl-gen) is what we plan to use to generate event implementations.

This concept was borrowed from Bukkit and as I have come to learn, is a very bad idea. This to some degree significantly complicated the transition from `text` to adventure, for instance.

For Velocity 2.0.0 we will take a page from Sponge and rework all Velocity events to use interfaces, then hide the actual event implementation in the proxy. So as to prevent duplication of code, we will try to auto-generate as much event-related code as possible.

## Reconsider the event result system

**Status**: The result system will not be changed. There will be some reorganization, but most developers seem to need help with using it. Velocity 2.0.0 will add convenience methods that will set results for the user.

The event result system was introduced to combine the action to take as a result of an event and any associated actions to go with it. This has resulted in much confusion from plugin developers. I do not have a concrete strategy to handle this yet.

## ~~Consider adding a cause system ala Sponge~~

**Status**: This system isn't likely to add value since most events in Velocity already have a clear, fixed cause-and-effect relationship. We may reconsider this later if there is developer demand for it.

MrIvanPlays:
> For Velocity 2.0.0 we will take a page from Sponge and rework all Velocity events to use interfaces, then hide the actual event implementation in the proxy. 

Don't rly understand how can this make a transition as breaking as the transition from `text` to `adventure` less complicated.

PR #384, opened 2020-11-05
Implement async event tasks.

Cybermaxke:
This implements the "Handling of potentially blocking events" portion of the event rework issue #378. 

The implementation no longer uses the kyori event bus, but a custom implementation for finer control. The [Lmbda](https://github.com/LanternPowered/Lmbda) library is used to generate annotated event handler implementations.

This new API also opens up the possibility of using Kotlin Coroutines in your event handlers, the following gist gives an example how to do that: [Velocity Event Coroutines](https://gist.github.com/Cybermaxke/50188a54130a29e9bd3696f06427cb33)


astei:
I'm very excited to see this! I'll review this soon (likely this weekend).

api/src/main/java/com/velocitypowered/api/event/AsyncTask.java
```diff
@@ -0,0 +1,78 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Consumer;
+
+@FunctionalInterface
+public interface AsyncTask {
+
+  /**
+   * Runs the task.
+   *
+   * @deprecated Should only be implemented as a lambda and called internally.
```
kashike: Why only as a lambda?

api/src/main/java/com/velocitypowered/api/event/Async.java
```diff
@@ -0,0 +1,17 @@
...
+import java.lang.annotation.Target;
+
+/**
+ * Marks an event handler method as async. By default are events handled on the
+ * network task executor, using this annotations performs them on an asynchronous
+ * task executor.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.METHOD)
+public @interface Async {
```
kashike: `AsyncSubscriber` or something, maybe?

api/src/main/java/com/velocitypowered/api/event/AsyncTask.java
```diff
@@ -0,0 +1,78 @@
+package com.velocitypowered.api.event;
+
+import static java.util.Objects.requireNonNull;
+
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Consumer;
+
+@FunctionalInterface
+public interface AsyncTask {
```
astei: I'm not a huge fan of the name `AsyncTask`. I don't have another good name for this - perhaps `EventTask` or even `EventContinuation` would be better (and then rename `AsyncTask.WithContinuation` to `EventContinuation.Deferred`).

Cybermaxke: I'm also not a big fan of the current naming, but I haven't found a good name to actually represent what is does. `DeferredEventTask` (`DeferredTask`) could be another option?

A248:
Since the implementation uses `MethodHandlesExtensions.privateLookupIn` which in turn uses MethodHandles#privateLookupIn, it should ideally be documented that anyone using modules needs to use `opens` in the module declaration, so that the event manager can perform its private lookups.

astei:
I have yet to add JPMS support to Velocity. We're going to have to see where it takes us. I anticipate a lot of challenges with adding it and allowing a degree of backwards compatibility.

A248:
There doesn't need to be any explicit support - it is still possible someone has modular code which depends on the current Velocity API and might need to change their module declaration. But really this is very minor.

Cybermaxke:
Velocity has control over how plugin jars can be loaded into the module system, so velocity can make sure it has the right amount of access it needs to handle events, without plugins having to do anything. But anyway, that's something for proper module support.

api/src/main/java/com/velocitypowered/api/event/AsyncTask.java
astei: It needs to include the word `Event` as otherwise it could be confused with the scheduler. I think `DeferredEventTask` is fine.

api/src/main/java/com/velocitypowered/api/event/Async.java
astei: In the code prototype he showed off on Discord, `@Async` is meant to be combined with `@Subscribe`.

kashike: Sure, but `Async` is a bit vague here, especially when working with certain other libraries that have classes named the same.

Cybermaxke: I don't think `@Async` will be a big problem, you're more likely to confuse the guava `Subscribe` with the velocity one with the current libraries, I don't see another `Async` class or annotation

Proximyst: Perhaps then Velocity's annotation should just get a new `async` boolean property?

MrIvanPlays:
ah yes, a feature I need. tahnks <3

api/src/main/java/com/velocitypowered/api/event/Async.java
Cybermaxke: Personally I prefer `@Async`, but an `async` property is also an option. @astei what do you prefer?

kashike: `async` property gets my vote.

MrIvanPlays: `async` property

astei:
@Cybermaxke can you make sure all tests pass and fix the merge conflict? I'd like to merge this in soon.

Cybermaxke:
Should be fine now, the only test that fails is `VelocityCipherTest.nativeIntegrityCheck()`.

astei:
I noticed that test failure and need to investigate it more deeply. Merging this for now.

astei:
FYI: The native tests do pass on my local system (Fedora 33) and on our Jenkins (Ubuntu 18.04).

PR #405, opened 2020-12-19
build: Fix search when generating Javadoc with broken java versions

zml2008:
This will fix Javadoc generated in the future. Existing Javadoc can be fixed manually, just replace:

```
var useModuleDirectories = true;
```
in each file's header with
```
var useModuleDirectories = false;
```

This bug was fixed for JDK 12 and the option was removed, so the application of the option is version-gated.


astei:
Javadocs are fully rebuilt with each push so I'll just merge this in directly.

PR #411, opened 2021-01-06
Expose proxy protocol setting in ProxyConfig

Camotoy:
Turns out people see `use-proxy-protocol` in our Geyser configuration and think it's required for proxy usage. This way we can auto-toggle our proxy protocol setting based on Velocity's.


astei:
We should probably rename it to `use-haproxy-protocol` instead. I'm not fully comfortable adding this to the API.

Camotoy:
> We should probably rename it to `use-haproxy-protocol` instead.

On Geyser's end or Velocity's? If Geyser, I'm not a fan of config options that break backwards compatibility. Totally understandable on the API front, though. Do what you want with this PR.

astei:
Both ends. You can add a check to fall back to `use-proxy-protocol` if needed.

Camotoy:
Fair enough.

PR #412, opened 2021-01-17
Add option for a SSL/TLS connection to backend servers

McMackety:
Due to the nature of the Minecraft protocol, encryption can only be done between the client and Velocity proxy, leaving an unencrypted proxy to backend server connection. In local area networks, this is not a problem, however, if a proxy connects to a backend server over the internet, all player data is exposed to the routers that the packets hop through.

To fix this, I believe that encapsulating the proxy to backend connection in an SSL/TLS connection would be the easiest and fastest method. Obviously, the backend servers would still need a plugin to inject a SslHandler into the server's Netty pipeline,  however, ViaVersion already provides a great example for injecting into Netty so it _shouldn't_ be a huge issue (_knock on wood_).

Here is a quick roadmap to making this feature a reality:
- [X] Add an option to enable SSL/TLS when connecting to the backend in the configuration.
- [X] Initialize an SSL/TLS context in the VelocityServer.
- [X] Inject an SSL/TLS handler into the backend channel pipeline.
- [ ] Add option to use certificate files and a certificate authority certificate to prevent MITM.
- [ ] Might want to find a way to automate certificate and CA generation to prevent MITM, as the user experience of manually generating it with OpenSSL or keytool is... well... yeahhhh.

I'm still working on this, but of course I'm looking for feedback on this, and if there are any glaring issues with my methodology feel free to point them out.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/scheduler/VelocityScheduler.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/scheduler/VelocityScheduler.java
@@ -7,6 +7,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Multimap;
 import com.google.common.collect.Multimaps;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
+import com.velocitypowered.api.plugin.PluginContainer;
 import com.velocitypowered.api.plugin.PluginManager;
 import com.velocitypowered.api.scheduler.ScheduledTask;
 import com.velocitypowered.api.scheduler.Scheduler;
@@ -183,8 +184,18 @@ public class VelocityScheduler implements Scheduler {
         currentTaskThread = Thread.currentThread();
         try {
           runnable.run();
-        } catch (Exception e) {
-          Log.logger.error("Exception in task {} by plugin {}", runnable, plugin, e);
+        } catch (Throwable e) {
+          //noinspection ConstantConditions
+          if (e instanceof InterruptedException) {
+            Thread.currentThread().interrupt();
+          } else {
+            String friendlyPluginName = pluginManager.fromInstance(plugin)
+                .map(container -> container.getDescription().getName()
+                      .orElse(container.getDescription().getId()))
+                .orElse("UNKNOWN");
+            Log.logger.error("Exception in task {} by plugin {}", runnable, friendlyPluginName,
+                e);
+          }
         } finally {
           if (repeat == 0) {
             onFinish();


astei:
Hi there, thanks for your initial contribution. It's still a WIP but I would like to provide a few comments:

The introduction of TLS to the Minecraft protocol is a noble endeavour. In a perfect world, it would provide proof that the server is communicating securely with the proxy and stop snooping attacks. With client certificates, it would also be able to prevent spoofing as well, since the spoofer probably doesn't have the required certificates.

However, there is a bit of an issue with this. TLS is _much_ more difficult to set up correctly in the context of Minecraft. There is no native TLS support in the Minecraft server, let alone the client. If you propose using a TLS tunnel for the server side to work around that, then you're assuming a level of control that effectively means that you might as well set up a TLS client tunnel on each proxy to forward.

Because you have such a high level of control if you're powerful enough to use a TLS tunnel on each server, why not use a VPN? They are protocol-agnostic, so they can protect data going over to other services that would otherwise transmit them in the clear. The only case where TLS makes the most sense is if you're in an environment that's locked down enough that a VPN isn't an option but TLS is (probably shared hosting, obscure Docker-based setups, or Kubernetes over the internet). I view this as very rare.

McMackety:
To clarify what I mean by the proxy connecting to each backend server, I don't mean a TLS tunnel as in a separate application that proxies traffic to the Minecraft server. I mean injecting TLS into Minecraft's Netty pipeline in the same way ViaVersion injects itself into the Netty pipeline, via a Bukkit/Sponge/Fabric plugin/mod that does such. (That would provide native TLS support as Netty is both included and used for all networking in the Minecraft Server.)

However, yes, it would be a complete pain to use TLS on the Minecraft server, because certificates really do suck. As I'm writing this I had another idea instead of TLS though. With Velocity modern forwarding there is a secret that needs to be on all backend servers in order for the proxy to connect to them. This secret could be used to seed a cipher (Probably ChaCha20 as it's built into Java 11) without performing any Diffie-Hellman style handshakes in the process. It would provide both verifications of the connection and encryption. Though if the private key is ever shared all communications would be instantly compromised.

Also, the reason why I'm doing this is that I've hit that rare case in which encryption is necessary for data security. Though I can see it be useful as a default for modern forwarding as most people setting up Velocity likely aren't networking engineers and do leave their networks vulnerable.

Let me know if that cipher seeded by the secret key may be a better option. I'm going to try testing that out right now.

astei:
I wouldn't recommend using the forwarding secret for security - people paste it publicly all the time despite us telling them not to (this is why we added `/velocity dump`). Also note that Velocity 1.1.x still needs to work on Java 8 so ChaCha20 isn't an option.

I still believe that a VPN is a better solution than extending the protocol arbitrarily. If you need encryption for data security over the wire you might as well apply it wholesale, and what better than a VPN for that?

McMackety:
Whoops, forgot to add into that previous post why I closed it.

Decided that it would be better to use a plugin for my specific use case, as my use case truly is ultra-niche and is going to be integrated with Hashicorp Vault. Thanks for looking at this though, will take a look at some of your issues and see if there's anything I can help out on.

PR #419, opened 2021-01-23
Use keepalive queue instead of just the last keepalive.

0-x-2-2:
This solves some issues with some Anti Cheats and custom server software that would work normally with a vanilla client and bungeecord.


astei (approved):

astei:
Thanks!

PR #421, opened 2021-01-26
Update bStats and migrate to new config file

Bastian:
This PR simplifies the Metrics class by using the new base module from bStats. It also migrates to the new bStats config file that will be used by plugins that integrate bStats. If a user disabled bStats in the old config file, this setting will be copied to the new config file.

This PR is WIP because it still depends on a snapshot version of the bStats base metrics class. I will soon release `2.1.1` and then update this PR to not use a snapshot version.

I've also removed the following code
```java
if (!metricsConfig.isFromConfig()) {	
  // Log an informational message.	
  logger.info("Velocity collects metrics and sends them to bStats (https://bstats.org).");	
  logger.info("bStats collects some basic information like how many people use Velocity and");	
  logger.info("their player count. This has no impact on performance and this data does not");	
  logger.info("identify your server in any way. However, you may opt-out by editing your");	
  logger.info("velocity.toml and setting enabled = false in the [metrics] section.");	
}
```
because I was unsure what the `#isFromConfig()` method does and how it might be affected by the change. If you still want to log this info message, I have to re-add the log messages somewhere.

### DIFF
--- a/proxy/build.gradle
+++ b/proxy/build.gradle
@@ -76,6 +76,8 @@ dependencies {
 
     implementation 'com.electronwill.night-config:toml:3.6.3'
 
+    implementation 'org.bstats:bstats-base:2.2.0'
+
     compileOnly 'com.github.spotbugs:spotbugs-annotations:4.1.2'
 
     testImplementation "org.junit.jupiter:junit-jupiter-api:${junitVersion}"
@@ -126,6 +128,8 @@ shadowJar {
     exclude 'it/unimi/dsi/fastutil/objects/*Reference*'
     exclude 'it/unimi/dsi/fastutil/shorts/**'
     exclude 'org/checkerframework/checker/**'
+
+    relocate 'org.bstats', 'com.velocitypowered.proxy.bstats'
 }
 
 artifacts {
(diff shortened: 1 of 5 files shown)


Bastian:
I'm not quite sure why the tests are failing.
Locally, `gradlew test` and `gradlew check` run successfully. As the previous commits in the `dev/1.1.0` branch also failed in Travis, I'll just assume that it's unrelated to my PR. Please tell me, if there's something that I'm missing and that needs to be fixed.

astei:
I believe it is a Travis issue (this only started in November if I'm correct). I will switch over to GitHub Actions.

astei:
> because I was unsure what the #isFromConfig() method does and how it might be affected by the change. If you still want to log this info message, I have to re-add the log messages somewhere.

This is intended to notify users on the first time the proxy starts that we will send statistics to bStats. This is intended to provide the user with the opportunity to opt out of bStats if they don't want to use it.

Bastian:
I see. I can add a similar message when the new config file gets generated for the first time.

Bastian:
I've just published version 2.2.0 of the bStats Metrics classes to Maven Central. Thus, this PR is now ready for review and for being merged.

astei (approved):
LGTM

PR #422, opened 2021-01-31
WIP: Unix domain socket binding support

hugmanrique:
A niche feature, but useful for upper-level proxies running on the same host, e.g. HAProxy (which Velocity already supports). I followed the Redis configuration convention, where the binding address and Unix domain socket address are different options. Mixing both types of sockets under the `bind` config would complicate determining if the given address is a hostname or a filename.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
@@ -50,8 +50,10 @@ import io.netty.bootstrap.Bootstrap;
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelInitializer;
 import io.netty.channel.EventLoopGroup;
+import io.netty.channel.unix.DomainSocketAddress;
 import java.io.IOException;
 import java.net.InetSocketAddress;
+import java.net.SocketAddress;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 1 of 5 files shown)


astei:
I actually already did this in Velocity 2.0.0: https://github.com/VelocityPowered/Velocity/commit/b00389029f4c11ad34898b5298625f4fcfc03de5

I'd be fine with you backporting this.

hugmanrique:
Ughh, I hate myself for not looking on the 2.0 branch. API changes like https://github.com/VelocityPowered/Velocity/commit/b00389029f4c11ad34898b5298625f4fcfc03de5#diff-9045d615f487de57c8df1f7e008486d03740e5051247b1b9007e8d15f42f2eccL18 are not possible in minor versions, so will have to wait.

astei:
Fair enough. I will need to find some time to forward-port all the recent Velocity 1.1.x changes to the 2.0.0 branch. At the same time I'm going to try and rein in all the changes planned for 2.0.0.

8MiYile:
I hope your UNIX Socket method is used for the following configuration
Example:
```toml
[servers]
# Configure your servers here. Each key represents the server's name, and the value
# represents the IP address of the server to connect to.
lobby01 = "tcp:127.0.0.1:30066"
lobby02 = "unix:/tmp/mcs-lobby.sock"
```

In addition, Paper server needs this support or other servers need to load additional plugins and support disabling the original game port

PR #430, opened 2021-02-16
Suggestion Permissions

davidmayr:
When using Velocity on my Network my players could see all the backend servers. They could not connect, but they could just tab the /server command and get them suggested. Back then this hasn't really bothered me... But now I have private servers no normal player should know about.

I added a basic check for permissions when autocompleting commands. They command will show up, but they won't be able to autocomplete. I tried to remove the command too, but that would require some API changes since I cannot get a CommandContext in the predicate that will be provided in the `ArgumentBuilder#requires(Predicate<S>)` method. Without the CommandContext I cannot create an Invocation which is required by the InvocableCommand to check for permissions.


astei:
Looks like this reimplements missing functionality from the Velocity command manager rewrite.

However, I'm going to need you to write a unit test for this. Perhaps try to suggest a command and check for a permission explicitly.

PR #433, opened 2021-02-22
Command system implementation cleanup

hugmanrique:
**Arguments are parsed by Brigadier**. This avoids needlessly parsing the arguments first as
a string and then as the specific type required by each `Command` implementation. This required
adding a `StringArrayArgumentType` for `SimpleCommand` and legacy commands.

**Introducing `ArgumentBuilder#requiresWithContext`**. This allows plugins using `BrigadierCommand`s
to get information about the command being parsed via `CommandContextBuilder` and `StringReader`
objects passed to the `BiPredicate`. These are used internally and are not available in the [upstream
Brigadier repo](https://github.com/Mojang/brigadier), so beware if you want cross-platform compatibility.
The state of the context and reader received by the `BiPredicate` remains undocumented and might
change in the future.

**Hints can now provide suggestions to all clients**. Previously they were limited to client-side 1.13+
suggestions. This means your own `SuggestionProvider`s on `ArgumentCommandNode` hinting nodes are
considered (and merged with regular suggestions) when all clients tab-complete. These are also
returned by `CommandManager#offerSuggestions`. Don't worry, we still check for permissions in both
cases.

**Suggestions for `InvocableCommand`s only check for permissions once**. Before, plugins could receive
up to 4 `#hasPermission` calls (where 2 of them received exactly the same arguments). We have
implemented some alias-specific parsing logic for handling suggestions (that still mainly relies
on Brigadier) that doesn't rely on recursion and doesn't allocate as much memory as the previous
method. `BrigadierCommand`s will still use the previous logic for backwards compatibility, but
will now check for permissions (https://github.com/VelocityPowered/brigadier/pull/6). This means
nodes may still receive duplicate `requires` calls if the given input is included in the
`ParseResults` (first in `parseNodes` and then in `getCompletionSuggestions`).

**More correct execution via hinting nodes**. These are now ignored when parsing. We attach a
`source -> false` requirement predicate to every node recursively, instead of _effectively_ cloning
the arguments node.

**Other optimizations and bug fixes**.
- Micro-optimize literal node parsing (https://github.com/VelocityPowered/brigadier/pull/3).
- `SuggestionsProvider` doesn't have the issue described on https://github.com/Mojang/brigadier/pull/81.
- Redirects where the target has no children now work as expected (https://github.com/VelocityPowered/brigadier/pull/2).

--- 
See https://github.com/VelocityPowered/Velocity/issues/432

TODO:
- [x] Add hinting support back (some tests are failing)
- [x] Introduce new tests
- [ ] Run new tests against dev/1.1.0 branch to check for regressions
- [x] Remove BrigadierUtils


astei:
I've gone ahead and merged in VelocityPowered/brigadier#1.

hugmanrique:
Apart from missing tests for legacy commands, Brigadier commands, and RawCommands, this is ready for review.
CI is failing because we use methods implemented on https://github.com/VelocityPowered/brigadier/pull/6.

hugmanrique:
It turns out some API changes are needed to make the command framework thread-safe. These are prohibiting mutation of the command tree (e.g. registering and unregistering commands) in the `Command#hasPermission` and `Command#suggest` methods. If you have some use cases for this please share them here. I'm probably gonna target this PR to the 2.0 branch once I have more free time.

astei:
> It turns out some API changes are needed to make the command framework thread-safe. These are prohibiting mutation of the command tree (e.g. registering and unregistering commands) in the `Command#hasPermission` and `Command#suggest` methods. If you have some use cases for this please share them here. I'm probably gonna target this PR to the 2.0 branch once I have more free time.

We've never officially said that mutating the command tree in `Command#hasPermission` or `Command#suggest` was permissible. The expectation is that they don't modify state, but rather look up data for display to the user. I think we _should_ codify this for 2.0.0 though.

hugmanrique:
Yes, of course. I meant to say I was going to create another PR targeting the 2.0 branch adding these requirements to the javadoc, and making this PR target 2.0 as well (albeit I will try to fix the tab complete issue with less code modifications in 1.1)

hugmanrique:
Superseded by #520.

PR #434, opened 2021-02-26
Extend ConnectionHandshakeEvent

paulzhng:
## Summary
This PR extends the `ConnectionHandshakeEvent` for it to include methods to set the hostname of the handshake and to set the remote address (the IP). It also allows denying the event.

## Motivation
Currently, there's no way to access this information without at least some kind of Reflection. This is not only hard-to-write and hard-to-read code but also code which is not future-proof at all. Internal changes in Velocity might affect the functionality of some plugins which **depend** on modifying internal fields.

There are several use-cases for this kind of extension:
* IP bans / blocks in an earlier state of connection
* DDoS mitigation solutions like TCPShield 
* Other services which inject data into the `hostname` field of the packet (e.g. Geyser although they've changed their procedure for Floodgate 2.0 so this PR probably wouldn't affect them)

## Prior art
From a proxy side of things, there is currently no proxy software supporting an equal use-case. From a Minecraft server side of things though, there is Paper's [`PlayerHandshakeEvent`](https://papermc.io/javadocs/paper/1.16/com/destroystokyo/paper/event/player/PlayerHandshakeEvent.html) which includes everything added by this PR.

## Drawbacks
Previously, `ConnectionHandshakeEvent` would not block the connection flow. After this change, `connection#setSessionHandler()` will only be invoked **after** the event is handled. This could lead, depending on the listeners, to unexpected behavior. 

## Alternatives
1. keep on using Reflection
2. using a more specialized event for this kind of usage (e.g. `ExtendedConnectionHandshakeEvent`) so we don't change the current behavior

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/event/connection/ConnectionHandshakeEvent.java
+++ b/api/src/main/java/com/velocitypowered/api/event/connection/ConnectionHandshakeEvent.java
@@ -3,28 +3,40 @@ package com.velocitypowered.api.event.connection;
 import com.google.common.base.Preconditions;
 import com.velocitypowered.api.event.ResultedEvent;
 import com.velocitypowered.api.proxy.InboundConnection;
+import java.util.Optional;
+import javax.annotation.Nullable;
 import net.kyori.adventure.text.Component;
 import net.kyori.adventure.text.serializer.plain.PlainComponentSerializer;
 import org.checkerframework.checker.nullness.qual.NonNull;
 
-import javax.annotation.Nullable;
-import java.util.Optional;
-
 /**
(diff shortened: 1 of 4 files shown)


astei:
For various reasons, this is probably better suited for Velocity 2.0.0.

astei:
@paulzhng please take a look at #498 and see if you like it.

paulzhng:
Hi, I'm no longer involved with the service I did the PR for but as far as I am / was concerned, your implementation looks like exactly what I had needed back then. Thanks for taking your time on this.

astei:
PR #498 merged.

PR #436, opened 2021-03-03
PermissionCheckEvent

MrLolok:
Add an event in velocity's API that is triggered when the proxy check if someone got permission to do something (like [PermissionCheckEvent](https://papermc.io/javadocs/waterfall/index.html?net/md_5/bungee/api/event/PermissionCheckEvent.html))


astei:
Why did you open a PR, trying to merge something entirely unrelated? I don't intend to revisit the permissions system right now, there are more fundamental issues that need attention.

A248:
How this PR was opened aside, I'm not so sure an event for every permission check is desirable. That seems like it would cause annoying overhead on permission checks.

If you are trying to implement a permissions plugin, you need to use the PermissionsSetupEvent

PR #437, opened 2021-03-06
Add 1.16.5

Euphillya:
This is frankly not a major update, and is even unnecessary, but at least showing server status through minecraft or whatever, showing that 1.16.5 is supported won't hurt. Especially since not everyone knows that 1.16.4 and 1.16.5 are identical in terms of the connection protocol.


Euphillya:
image

Xernium (changes requested):
Doing this has the adverse side-effect of duplicating the packet state registry for these versions. However your point is taken. I will see what can be done about this minor inconvenience.

Xernium:
https://github.com/VelocityPowered/Velocity/commit/92e8aa0a374458d196de914aaaeb9a0fd6811480 for a first attempt. I want @astei to look over this

astei:
92e8aa0 is the most probable path for adding this. Thanks for PRing anyway.

PR #443, opened 2021-03-14
Remove circular dependencies in VelocityServer where sensible

A248:


proxy/src/main/java/com/velocitypowered/proxy/Velocity.java
```diff
@@ -54,14 +60,23 @@ public static void main(String... args) {
 
     long startTime = System.currentTimeMillis();
 
-    VelocityServer server = new VelocityServer(options);
+    VelocityPluginManager pluginManager;
+    VelocityEventManager eventManager;
+    VelocityConsoleCommandSource console;
+    VelocityServer server = new VelocityServer(
+            (pluginManager = new VelocityPluginManager()),
```
kashike: Can you move these assignments to the variables directly, and reference them normally here instead?

kashike (changes requested):

proxy/src/main/java/com/velocitypowered/proxy/Velocity.java
A248: It will be less beautiful then. I will do it, though

A248:
I'm not sure why the tests are failing. They succeed locally. The failure is related to the CommandManager tests.

astei:
I think @hugmanrique needs to look into that.

hugmanrique:
The failing test is incorrect and the actual expected behavior was introduced by the recent Brigadier merges. It has to do with considering both the hint suggestions and the regular suggestions when a command has both. I'm not at the computer right now, but the test can be `@Disabled` for now.

Sorry for the breakage 😓

PR #452, opened 2021-03-28
Make the logging of server connections configurable

roccodev:
Makes the logging of server connections (`[server connection] Player -> SERVER has (dis)connected`) configurable.

I think `MinecraftConnection` can only handle `VelocityServerConnection`s, but I added an instance check anyway.


A248:
Any and all log messages in Velocity are already configurable via log4j2. I'd suggest you refer to such documentation on logging. It's quite powerful.

roccodev:
While that's a good thing, log messages for command invocations are already configurable, and I wouldn't expect most server owners to be familiar with log4j.
If this ends up being added I would move all the logging-related settings to a `logging` section in the config.

4drian3d:
Since pull request #673 was merged, I think this pull request can be closed

4drian3d:
This change was already integrated in pull request #673 (2 years ago)

PR #453, opened 2021-03-28
Add a display name attribute for Player

elliotnash:
This pr adds a display name attribute to the Player interface, much like what spigot and bungeecord have. Although not used from within velocity, this could be useful to standardize name customization across plugins. For example, a chat plugin might have a nickname feature, and a tablist plugin should respect that nickname.
Currently I have the display name as a Component, with no functions for legacy text serialization. My understanding is that legacy text is being phased out in favour of Components, so that shouldn't be a problem.

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/proxy/Player.java
+++ b/api/src/main/java/com/velocitypowered/api/proxy/Player.java
@@ -31,6 +31,19 @@ public interface Player extends CommandSource, Identified, InboundConnection,
    */
   String getUsername();
 
+  /**
+   * Sets the player's display name.
+   *
+   * @param displayName the display name as a {@link Component}
+   */
+  void setDisplayNameComponent(Component displayName);
+
+  /**
+   * Returns the player's current display name as a {@link Component}.
+   *
+   * @return the display name
+   */
+  Component getDisplayNameComponent();
 
   /**
    * Returns the player's UUID.
--- a/api/src/main/java/com/velocitypowered/api/proxy/Player.java
+++ b/api/src/main/java/com/velocitypowered/api/proxy/Player.java
@@ -36,14 +36,14 @@ public interface Player extends CommandSource, Identified, InboundConnection,
    *
    * @param displayName the display name as a {@link Component}
    */
-  void setDisplayNameComponent(Component displayName);
+  void setDisplayName(Component displayName);
 
   /**
    * Returns the player's current display name as a {@link Component}.
    *
    * @return the display name
    */
-  Component getDisplayNameComponent();
+  Component getDisplayName();
 
   /**
    * Returns the player's UUID.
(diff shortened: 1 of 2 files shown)


api/src/main/java/com/velocitypowered/api/proxy/Player.java
```diff
@@ -31,6 +31,19 @@
    */
   String getUsername();
 
+  /**
+   * Sets the player's display name.
+   *
+   * @param displayName the display name as a {@link Component}
+   */
+  void setDisplayNameComponent(Component displayName);
```
kashike:
```suggestion
  void setDisplayName(Component displayName);
```

api/src/main/java/com/velocitypowered/api/proxy/Player.java
```diff
@@ -31,6 +31,19 @@
...
+
+  /**
+   * Returns the player's current display name as a {@link Component}.
+   *
+   * @return the display name
+   */
+  Component getDisplayNameComponent();
```
kashike:
```suggestion
  Component getDisplayName();
```

kashike (changes requested):

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
```diff
@@ -405,6 +417,7 @@ public ConnectionRequestBuilder createConnectionRequest(RegisteredServer server)
   @Override
   public void setGameProfileProperties(List<GameProfile.Property> properties) {
     this.profile = profile.withProperties(Preconditions.checkNotNull(properties));
+    this.displayName = Component.text(profile.getName());
```
A248: Why are you setting the display name back to the original profile name here? By resetting the display name in a place where the name is not concerned, you're creating some minorly surprising behavior.

elliotnash: I thought that setting the gameprofile here could change the username, in which case I thought it might make sense to reset the display name, but I'm not sure that's the case

mikroskeem:
In my opinion this does not belong in Velocity. Shouldn't bloat API with anything not related / helpful to the vanilla client/server, this remains quite useless without a plugin in place actually utilizing display names.

elliotnash:
What would be the prefered solution to setting a proxy wide nickname/display name then? I think having this built into the proxy is much more convenient than having to rely on the API's of other plugins to set plugin-wide nicknames

mikroskeem:
#268 & get your interface popular enough for plugins to use and/or implement.

astei:
This is superseded by the APIs in KyoriPowered/adventure#352 which, unlike this PR, would be cross-platform. The approach taken in Adventure is IMO more palatable to me.

PR #454, opened 2021-03-31
Added ListenerBoundEvent and ListenerCloseEvent.

### LINKED ISSUE #444: Socket bound event
Add an event for when the proxy is bound to an address. Eg. `ProxyBoundEvent`. This will be useful for dynamically managing the proxy.

Cubxity:
Resolves #444.

## Description
This introduces events that allow plugins to dynamically manage the proxy. 

## Events
ListenerBoundEvent - fires asynchronously after a socket is bound
ListenerCloseEvent - fires synchronously before a socket is closed


api/src/main/java/com/velocitypowered/api/event/proxy/ProxyBoundEvent.java
```diff
@@ -0,0 +1,25 @@
+package com.velocitypowered.api.event.proxy;
+
+import com.google.common.base.Preconditions;
+import java.net.InetSocketAddress;
+
+/**
+ * This event is fired by the proxy after the proxy starts accepting connections.
```
astei: As presently used, this will in fact fire after _any_ bind (this includes the query protocol and proxy reloads). Please update the documentation - personally, `ListenerBoundEvent` or similar would be a better name.

astei (changes requested):

api/src/main/java/com/velocitypowered/api/event/proxy/ProxyBoundEvent.java
Cubxity: I believe it only fires when the Minecraft port is bind, not query. But I'll update the documentation and the name.

Cubxity: Should I also rename ProxyCloseEvent to something like `ListenerCloseEvent`, which fires after each listener is closed?

astei: Yeah, my bad - it won't fire for query. However, it probably should fire for query too. I do think a `ListenerCloseEvent` is better than a `ProxyCloseEvent`, which means reloads could be accommodated (say for a hypothetical dynamic DNS plugin updating a port).

Cubxity: Would separate events be better (e.g. `ListenerBoundEvent` and `QueryListenerBoundEvent`) or some sort of field indicating whether the listener is for query?

api/src/main/java/com/velocitypowered/api/event/proxy/ListenerBoundEvent.java
```diff
@@ -0,0 +1,38 @@
...
+import com.google.common.base.Preconditions;
+import java.net.InetSocketAddress;
+
+/**
+ * This event is fired by the proxy after a listener starts accepting connections.
+ */
+public final class ListenerBoundEvent {
+
+  private final InetSocketAddress address;
+  private final boolean isQuery;
```
astei: Enum called `ListenerType` instead?

api/src/main/java/com/velocitypowered/api/event/proxy/ListenerCloseEvent.java
```diff
@@ -0,0 +1,32 @@
...
+
+import com.google.common.base.Preconditions;
+import java.net.InetSocketAddress;
+
+/**
+ * This event is fired by the proxy before the proxy stops accepting connections.
+ */
+public final class ListenerCloseEvent {
+
+  private final InetSocketAddress address;
```
astei, line 19: Include the type here too?

astei (changes requested):

api/src/main/java/com/velocitypowered/api/event/proxy/ListenerCloseEvent.java
Cubxity: The close() method in ConnectionManager does not include type, so I'm not quite sure what to do with that

astei: Then add it.

Cubxity: Oh, my bad - I thought it was a public api.

api/src/main/java/com/velocitypowered/api/network/ListenerType.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+package com.velocitypowered.api.network;
+
+/**
+ * Represents each listener type.
+ */
+public enum ListenerType {
+  MINECRAFT("Minecraft"),
+  GS4("GS4");
```
astei: Call it just query here.

astei (changes requested):

astei (approved):

PR #447, opened 2021-03-21
Updated guice version to support Java 15

QuickGlare:
I've updated the version of Guice to support Java 15, I've done some tests and in my environment works fine.

PR #459, opened 2021-04-14
Bump guice to 5.0.1 to prevent issues with java 16

derklaro:
As java released the latest version 16 there was an issue introduced into velocity because of using an older version of guice the proxy was unable to create plugin instances failing with:
```
Caused by: com.google.inject.internal.cglib.core.$CodeGenerationException: java.lang.reflect.InaccessibleObjectException-->Unable to make protected final java.lang.Class java.lang.ClassLoader.defineClass(java.lang.String,byte[],int,int,java.security.ProtectionDomain) throws java.lang.ClassFormatError accessible: module java.base does not "opens java.lang" to unnamed module @73a1e9a9
```

Updating guice to the latest release 5.0.1 solves the issue.

### DIFF
--- a/api/build.gradle
+++ b/api/build.gradle
@@ -47,7 +47,7 @@ dependencies {
     }
 
     api "org.slf4j:slf4j-api:${slf4jVersion}"
-    api 'com.google.inject:guice:4.2.3'
+    api 'com.google.inject:guice:5.0.1'
     api "org.checkerframework:checker-qual:${checkerFrameworkVersion}"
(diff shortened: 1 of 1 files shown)


Xernium:
Not happening. See https://github.com/VelocityPowered/Velocity/pull/447

astei:
Please see #447 for why we're not merging this. 01569309 is the provided workaround which will mitigate the issue and does not break backwards compatibility.

PR #481, opened 2021-04-24
Made shutdown message configurable

nicokempe:
I imagined that some users would want to customize the shutdown message.


astei:
Comments that are effectively +1s are not welcome.

I am not likely to merge this, since the next major revision of Velocity goes much further and allows every user-facing message in the proxy to be changed.

astei:
Closing since this feature did land in Velocity 3.0.0

PR #482, opened 2021-04-25
Ignore empty server packets

Etrayed:
A fix similar to [this](https://github.com/PaperMC/Waterfall/commit/c19c4771deb96ad88e01781674d7d716dd10d3c0). 1.12 players are sometimes receiving empty packets from the server, these should be discarded, but empty packets from the client will still throw an exception.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/protocol/netty/MinecraftDecoder.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/protocol/netty/MinecraftDecoder.java
@@ -19,6 +19,8 @@ package com.velocitypowered.proxy.protocol.netty;
 
 import com.google.common.base.Preconditions;
 import com.velocitypowered.api.network.ProtocolVersion;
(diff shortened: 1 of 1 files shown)


Etrayed:
The changes done by the commit linked above have already been changed by [this commit](https://github.com/PaperMC/Waterfall/commit/d43773ad9c55a1e28463d9d6654cefaa072ef052), just to clarify.

Leymooo:
Does this problem exist in velocity? Looks like velocity already discard these packets https://github.com/VelocityPowered/Velocity/blob/dev/1.1.0/proxy/src/main/java/com/velocitypowered/proxy/protocol/netty/MinecraftVarintFrameDecoder.java#L58

Etrayed:
Yes it does still exist, we had it several times. The server sends like "empty" but still encrypted packets, so this check has to be on MinecraftDecoder level.

astei:
Worked around on our end. but please fix your server.

PR #485, opened 2021-04-27
Add color support for multi-line kick messages

cderszteler:
Add formtated kick reason to add color support over multiple lines


astei:
This PR makes no sense and even then it won't compile because you bungled the code style. Velocity supports multi-line disconnect messages just fine - your server doesn't, so fix your server.

cderszteler:
Ah yeah, you're right, I'm sorry. I serialized my messages with a paragraph and not a ampersand. My fault, sorry

PR #487, opened 2021-04-29
support arguments for ap

keyute:


zml2008:
Why make this special case when you could have a [simple source template](https://github.com/zml2008/gradle-plugins#templating), which is more flexible.

Either way, if this goes in, it should use constants rather than repeating magic string literals.

keyute:
From what I understand, this would require placing the plugin file at a different location, which can make referencing static variables defined in the class confusing.

zml2008:
wat

keyute:
If I were to add a static method in the Plugin class I've placed in the template folder, I would have to run the generate task every time the class is changed in order for IDEs to pick up those methods right?

keyute:
My impression of using templates for classes is that these class should not change much since they are source code. Putting the class which Velocity will load and is bound to have lots of internal changes (init code) is probably a bad idea.

zml2008:
don't do that then -- have a separate package-private `Constants` class that is the templated one, and just reference that from the `@Plugin`-annotated class.

api/src/ap/java/com/velocitypowered/api/plugin/ap/PluginAnnotationProcessor.java
```diff
@@ -99,4 +109,8 @@ public synchronized boolean process(Set<? extends TypeElement> annotations,
 
     return false;
   }
+
+  private String getOrElse(Map<String, String> map, String key, String defaultValue) {
+    return map.get(key) == null ? defaultValue : map.get(key);
```
tomcraft, line 114: Map#getOrDefault ?

keyute: https://docs.oracle.com/javase/7/docs/api/javax/annotation/processing/ProcessingEnvironment.html

An option with no value returns null which getOrDefault does not cover

astei:
I'm extremely unwilling to do this just so that you can satisfy your desire to, quote, "break my 100% Kotlin streak" because you want to do a source replacement using only Kotlin code. I'm going to need another good reason to consider this PR.

keyute:
many proposed solutions are simply too complicated or require additional dependencies just to get the version defined in gradle replaced. A filter() in ProcessResources is enough for Bukkit/Spigot projects while a huge workaround is needed for Velocity.

astei:
[Here's a solution in a mere 4 lines of code](https://github.com/NuVotifier/NuVotifier/blob/master/velocity/build.gradle#L1).

astei:
I remain unconvinced this is something we should add. I am willing to make reasonable concessions to support other JVM languages, but you refuse to consider other solutions that don't require changes to Velocity.

zml2008:
I've done Kotlin source templating just fine -- there's no reason you can't too.

keyute:
closing PR in favour for a self implemented solution

PR #491, opened 2021-05-06
Avoid unnecessary data copy when packet should not be compressed

Leymooo:


Leymooo:
Closed it prior to astei's better solution that will be added later

PR #500, opened 2021-05-16
Lessen restriction that EventHandler's event type extend Event

A248:
In Velocity 1.x, events only had to be `Object`s. I presume that the reason `Event` was introduced is to increase type-safety of the event manager, which is reasonable. For that reason, `fire(E)` still uses `Event`.

For `EventHandler`s, the restriction that the event type be an `Event` has been lifted. This generally allows more flexibility for users defining platform-independent APIs, restoring a feature of EventManager which was possible in 1.x.

Here's one example where this is useful: a platform-independent API defines an event interface which can be listened to through platform-specific means, including `EventHandler`. The platform-independent event interface does *not* need to extend Event.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/event/VelocityEventManager.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/event/VelocityEventManager.java
@@ -73,13 +73,13 @@ public class VelocityEventManager implements EventManager {
   private static final LambdaType<VoidHandler> untargetedVoidHandlerType =
       LambdaType.of(VoidHandler.class);
 
-  private static final Comparator<HandlerRegistration> handlerComparator =
+  private static final Comparator<HandlerRegistration<?>> handlerComparator =
       Comparator.comparingInt(o -> o.order);
 
   private final ExecutorService asyncExecutor;
   private final PluginManager pluginManager;
 
-  private final Multimap<Class<?>, HandlerRegistration> handlersByType = HashMultimap.create();
+  private final Multimap<Class<?>, HandlerRegistration<?>> handlersByType = HashMultimap.create();
   private final LoadingCache<Class<?>, HandlersCache> handlersCache =
       Caffeine.newBuilder().build(this::bakeHandlers);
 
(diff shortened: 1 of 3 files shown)


proxy/src/main/java/com/velocitypowered/proxy/event/VelocityEventManager.java
```diff
@@ -103,12 +103,12 @@ public VelocityEventManager(final PluginManager pluginManager) {
...
    * Represents the registration of a single {@link EventHandler}.
    */
-  static final class HandlerRegistration {
+  static final class HandlerRegistration<E> {
 
     final PluginContainer plugin;
     final short order;
     final Class<?> eventType;
-    final EventHandler<Event> handler;
+    final EventHandler<E> handler;
```
Cybermaxke, line 111: I prefer this changed back to `EventHandler<Object>` then and remove the generic, no need to pollute the impl with unnecessary generics.

proxy/src/main/java/com/velocitypowered/proxy/event/VelocityEventManager.java
```diff
@@ -426,12 +425,14 @@ public void fireAndForget(final Event event) {
 
   private <E extends Event> void fire(final @Nullable CompletableFuture<E> future,
```
Cybermaxke, line 426: `<E>` here as well

A248: Since *firing events* does have the restriction that the event fired extend `Event`, using `<E extends Event>` here is accurate. `fire(E)` has the same generic requirement.

If you'd like to remove `Event` entirely, that's possible too, although not the goal of this PR.

Cybermaxke:
I would personally either get rid of `Event` itself (so to before it was added), or leave it as it is now. Not something in between, this feels inconsistent. 
Between which platforms do you want to be able to share code? Bukkit/Sponge/BungeeCord also require some extend/implement of an `Event` interface/class.

A248:
Some of the limitations of the `Event` interface were discussed on discord. It's more about general flexibility with regards to other event systems, not necessarily any specific platforms such as Bukkit. KyoriPowered/event, for example, can provide an event bus with no such restriction.

I agree that the result of this PR would be a somewhat inconsistent event API, so I've opened a new one to remove `Event` entirely: https://github.com/VelocityPowered/Velocity/pull/502

PR #502, opened 2021-05-16
Remove the restriction that events extend Event

A248:
This change allows for greater flexibility with respect to other event systems. For example, it allows one to fire the same event on multiple event buses, so long as those other event buses also impose no restrictions on the fired event.

While the concept of the `Event` interface is not uncommon, it inhibits the flexibility of APIs which wish to use platform-independent event interfaces. It's also a little bit tedious to add to every event.

This feature was in Velocity 1.x, and hopefully it can stay for Velocity 2.x. I feel that adding `Event` is a regression in terms of API progress.


proxy/src/main/java/com/velocitypowered/proxy/event/VelocityEventManager.java
```diff
@@ -108,7 +107,7 @@ public VelocityEventManager(final PluginManager pluginManager) {
     final PluginContainer plugin;
     final short order;
     final Class<?> eventType;
-    final EventHandler<Event> handler;
+    final EventHandler<?> handler;
```
Cybermaxke, line 110: Might as well change this back to `EventHandler<Object>`

A248:
The original idea behind this PR was to allow plugins to define platform-independent events but listen to them via platform-specific event buses. As Cableguy pointed out, this is not a very sound strategy. So the benefits of this PR are probably very small.

In contrast, this sacrifices a lot of type-safety, so I'm no longer convinced that this change is justified. If someone else has a compelling idea or use-case for this change, feel free to revive this PR. I personally just don't see much more value to it as things are.

PR #504, opened 2021-05-17
Simplify EventTask. Add Continuation parameter support.

Cybermaxke:
- Simplify EventTask. No more subtypes and is now an interface. No default methods though to prevent it being implemented as a lambda. Renamed `run` -> `execute` to be consistent with `EventHandler`.  Deleted `EventTask.asyncWithContinuation` and `EventTask.of`.
- Add support for `Continuation` as a second subscribe method parameter, instead of having to use `EventTask` as return type.
```java
@Subscribe
public void onEvent(MyEvent event, Continuation continuation) { 
  // do things
  continuation.resume();
}
```
- Preparations to support `suspend fun` for kotlin event handling.

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/event/EventTask.java
+++ b/api/src/main/java/com/velocitypowered/api/event/EventTask.java
@@ -17,13 +17,10 @@ import java.util.function.Consumer;
  * be suspended and resumed at a later time, and executing event handlers completely or partially
  * asynchronously.
  *
- * <p>By default will all event handlers be executed on the thread the event was posted, using
- * event tasks this behavior can be altered.</p>
+ * <p>By default will all event handlers be executed on the thread the event was posted, this
+ * behavior can be altered by using event tasks.</p>
  */
-public abstract class EventTask {
-
-  EventTask() {
-  }
+public interface EventTask {
 
   /**
    * Whether this {@link EventTask} is required to be called asynchronously.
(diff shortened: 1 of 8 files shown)


Cybermaxke:
Hold a bit off before merging this in, the continuation adapter approach may need to be handled differently because of how Kotlin handles suspending functions, I need to investigate a bit first.

astei:
@Cybermaxke I've marked this PR as a draft for you. When you're ready, you can just mark it ready for review.

Cybermaxke:
Should be good for review, I made the adapter system more flexible so it should work with this. I have the kotlin adapter ready locally, but I'll wait for my formatting PR to create a new PR for that.

astei:
Got it!

api/src/main/java/com/velocitypowered/api/event/EventTask.java
```diff
@@ -101,13 +55,14 @@ public boolean requiresAsync() {
...
+  static EventTask async(final Runnable task) {
     requireNonNull(task, "task");
-    return new Basic() {
+    return new EventTask() {
 
       @Override
-      public void run() {
+      public void execute(Continuation continuation) {
         task.run();
+        continuation.resume();
```
astei, line 65:
```suggestion
        try {
          task.run();
        } catch (Throwable throwable) {
          continuation.resumeWithException(throwable);
          return;
        }
        continuation.resume();
```

astei (changes requested):

Cybermaxke:
Something else came up for the adapter, so it will require even more changes.

api/src/main/java/com/velocitypowered/api/event/EventTask.java
Cybermaxke: This isn't necessary, if an exception is thrown during an event task execution, it will be resumed automatically.

Cybermaxke:
It's alive! Should work now.

astei (approved):

PR #508, opened 2021-05-24
Guarantee the presence of snakeyaml

A248:
In Velocity 1.x, snakeyaml is a visible transitive API dependency. With Configurate 4, it becomes an "implementation" dependency.

SnakeYaml is essentially a cornerstone of the plugin ecosystem, with many things - not just Configurate - relying on it. Velocity should guarantee its presence just as other platforms do.

This will also ease source code migration from 1.x to 3.x for codebases directly using snakeyaml (though I don't recommend such things)

### DIFF
--- a/api/build.gradle
+++ b/api/build.gradle
@@ -45,6 +45,7 @@ dependencies {
     api "org.spongepowered:configurate-hocon:${configurateVersion}"
     api "org.spongepowered:configurate-yaml:${configurateVersion}"
     api "org.spongepowered:configurate-gson:${configurateVersion}"
+    api "org.yaml:snakeyaml:1.28"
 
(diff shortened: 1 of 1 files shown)


astei:
This is redundant since Velocity 3.0.0 doesn't bump Configurate.

zml2008:
Longer-term, I am against this. The author of SnakeYaml encourages using SnakeYaml Engine, a major redesign of the library for Yaml 1.2 and Java 8. I intend to migrate Configurate over to that long-term, and I would encourage all other projects to do the same. It would be silly for velocity to ship two separate YAML libraries.

A248:
I'm going to close this for now (as it's targeted to dev/3.0.0). However, I've opened #509 for Polymer-related discussion on the matter

PR #515, opened 2021-06-05
[3.0] Document some restrictions in the command API

hugmanrique:
Cherry-picked from `dev/2.0.0`.

Among the changes, there are minor precondition additions that I suspect no plugins rely on (I cannot be 100% sure, though). However, plugins relying on the absence of the preconditions result in undefined behavior in `1.1`. Other changes include removing the reference to the version where the new API was introduced, and typo fixes.

This is needed in the `3.0` release so that the command implementation refactor can also target this major. Otherwise, we will have to wait for `4.0` (which could be a long way off).


astei:
This is good as any time to do it.

PR #516, opened 2021-06-07
Bump to Caffeine 3

A248:
The latest version of Caffeine has completely removed usage of `sun.misc.Unsafe`. Unsafe is still in the JDK, but for future compatibility, which Velocity 3 aims at, it is better to less rely on it (netty still does, AFAIK).

There are no breaking changes in Caffeine 3 which affect Velocity's usage of it in EventManager -- all breaking changes in [release 3.0](https://github.com/ben-manes/caffeine/releases/tag/v3.0.0) cause compile-time errors.


astei:
Given that our usage of Caffeine is internal to us alone, this is fine.

PR #517, opened 2021-06-07
Update to 1.17-rc2

James76931:
This PR allows users to connect with the new 1.17-rc2 version.


astei:
Hi there - since 1.17 is going to be released tomorrow (and we're releasing Velocity 1.1.6 today), there's no point in updating to -rc2 (it just has a couple of bug fixes). Thanks for the contribution!

PR #520, opened 2021-06-10
[3.0] Command system implementation refactor

hugmanrique:
Supersedes #433, targets the 3.0 branch.

Same notes from that PR apply, except that suggestions for Brigadier commands now fallback to Brigadier (to avoid inconsistencies). Also, `CommandManager` is now thread-safe. This PR contains _even_ more tests (106 at the time of writing); we now check the shape of the command graph sent through the `AvailableCommands` packet after injection for inconsistencies, and the suggestions system in general.

All tests pass on my machine, using a Brigadier build including https://github.com/VelocityPowered/brigadier/pull/9 and https://github.com/VelocityPowered/brigadier/pull/10. In addition, I've tested the behavior of suggestions in-game and in the console.


astei:
GitHub Actions didn't like it (and the build even hung for a nice 4+ hours for some reason). Can you try pushing another commit to get it to rebuild?

proxy/src/main/java/com/velocitypowered/proxy/command/CommandGraphInjector.java
```diff
@@ -0,0 +1,142 @@
...
+   * <p>Prior to adding a literal with the same name as one previously contained
+   * in the destination node, the old node is removed from the destination node.
+   *
+   * @param dest the root node to add the permissible nodes to
+   * @param source the command source to inject the nodes for
+   */
+  public void inject(final RootCommandNode<S> dest, final S source) {
+    lock.lock();
+    try {
+      final RootCommandNode<S> origin = this.dispatcher.getRoot();
```
astei, line 70: I generally am uneasy with running "alien code" we don't trust in critical sections like this. Can this be modified to clone the `RootCommandNode` in its entirety instead?

proxy/src/main/java/com/velocitypowered/proxy/command/SuggestionsProvider.java
```diff
@@ -0,0 +1,367 @@
...
+   *
+   * @param fullInput the command input
+   * @param futures the futures that complete with the suggestions
+   * @return the future that completes with the merged suggestions
+   */
+  @SafeVarargs
+  private CompletableFuture<Suggestions> merge(
+          final String fullInput, final CompletableFuture<Suggestions>... futures) {
+    // https://github.com/Mojang/brigadier/pull/81
+    return CompletableFuture.allOf(futures).handle((unused, throwable) -> {
```
astei, line 358: Consider using [Spotify `completable-futures`](https://github.com/spotify/completable-futures) which Velocity already includes. in particular, `CompletableFutures.combineFutures()` would probably lead to a cleaner implementation of this.

astei (changes requested):

proxy/src/main/java/com/velocitypowered/proxy/command/CommandGraphInjector.java
astei: (Likewise, the `SuggestionsProvider` does something similar. I'm not sure about the best way to handle that.)

astei: Now that I think about it, this is why you wanted the precondition in `InvocableCommand` for #515 - because as designed, this will cause a deadlock. This neatly explains why.

proxy/src/main/java/com/velocitypowered/proxy/command/SuggestionsProvider.java
```diff
@@ -0,0 +1,367 @@
...
+      final List<Suggestions> suggestions = new ArrayList<>(futures.length);
+      for (final CompletableFuture<Suggestions> future : futures) {
+        if (!future.isCompletedExceptionally()) {
```
A248: Would it be possible to at least log exceptionally-completed futures? Swallowing exceptions in whole makes for difficult debugging.

proxy/src/main/java/com/velocitypowered/proxy/command/VelocityCommandManager.java
```diff
@@ -190,30 +217,52 @@ private boolean executeImmediately0(final CommandSource source, final String cmd
...
-    return dispatcher.getCompletionSuggestions(parse)
-        .thenApply(suggestions -> Lists.transform(suggestions.getList(), Suggestion::getText));
+    final String normalizedInput = VelocityCommands.normalizeInput(cmdLine, false);
+    try {
+      return suggestionsProvider.provideSuggestions(normalizedInput, source)
+              .thenApply(suggestions ->
+                      Lists.transform(suggestions.getList(), Suggestion::getText));
+    } catch (final Throwable e) {
+      // Again, plugins are naughty
+      return CompletableFutures.exceptionallyCompletedFuture(
```
A248:
```suggestion
      return CompletableFuture.failedFuture(
```

A248 (commented):
Just what I noticed with respect to exceptions and futures.

proxy/src/main/java/com/velocitypowered/proxy/command/SuggestionsProvider.java
hugmanrique: Unfortunately `completable-futures` only provides overloads of [up to 6 `CompletableFuture`s](https://javadoc.io/doc/com.spotify/completable-futures/latest/index.html) with no varargs variant. However, e.g. `provideAliasSuggestions` needs to merge an arbitrary number of futures. In addition, the function uses [`thenCompose`](https://github.com/spotify/completable-futures/blob/bfd73772ebbf7ee60f742ebb4f176830ab1b9345/src/main/java/com/spotify/futures/CompletableFutures.java#L574), which is only called if all the futures complete normally. We also want to merge the suggestions (skipping the ones that completed exceptionally) on exceptional completion of the `allOf` future.

I'm not an expert on `CompletableFuture`(s), so I might have missed some method that handles this precise case.

proxy/src/main/java/com/velocitypowered/proxy/command/SuggestionsProvider.java
```diff
@@ -357,7 +358,10 @@ private static boolean shouldConsider(final String name, final String input) {
     return CompletableFuture.allOf(futures).handle((unused, throwable) -> {
       final List<Suggestions> suggestions = new ArrayList<>(futures.length);
       for (final CompletableFuture<Suggestions> future : futures) {
-        if (!future.isCompletedExceptionally()) {
+        if (future.isCompletedExceptionally()) {
+          final Throwable exception = CompletableFutures.getException(future);
```
A248, line 362: By the way, sometimes this exception, which is the cause of the CompletionException, can be null, but it's quite bad practice, requiring a subclass of CompletionException which adds no cause. I don't think such case is worth guarding against.

hugmanrique: As far as I can see, log4j will [handle the null `Throwable`](https://github.com/apache/logging-log4j2/blob/e395e13c513e17cb91781808f83b20ee559bf534/log4j-api/src/main/java/org/apache/logging/log4j/simple/SimpleLogger.java#L232-L244) just fine (I assume other `Logger` implementations use the same check, didn't bother looking).

astei:
I've merged in the appropriate Brigadier PRs BTW

astei:
The test suite passes (I was able to re-run it myself). This PR otherwise looks good to me, so I'm merging.

astei (approved):

PR #522, opened 2021-06-12
Expand Adventure Support

qixils:
- Adds Player#asComponent which mirrors vanilla behavior
- Adds DISPLAY_NAME pointer for ^
- Adds Player#key (yes it is a little useless)
- Adds Player#asHoverEvent
- Uses the new(?) Pointers class to allow #getOrDefault (and #getOrDefaultFrom) to work
- Moves the PermissionChecker pointer to the CommandSource class
- Moves other pointers to Player (not sure about this one-- it uses exclusively API methods so it makes sense to me that it should be on Player instead of ConnectedPlayer, but I'm open to criticism on that front, and everything else here for that matter :P)

first time making a PR like this, hope it's alright 😅

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/permission/PermissionSubject.java
+++ b/api/src/main/java/com/velocitypowered/api/permission/PermissionSubject.java
@@ -7,6 +7,9 @@
 
 package com.velocitypowered.api.permission;
 
+import net.kyori.adventure.permission.PermissionChecker;
+import net.kyori.adventure.util.TriState;
+
 /**
  * Represents a object that has a set of queryable permissions.
  */
@@ -29,4 +32,24 @@ public interface PermissionSubject {
    * @return the value the permission is set to
    */
   Tristate getPermissionValue(String permission);
+
+  /**
+   * Gets the permission checker for the subject.
+   *
+   * @return subject's permission checker
+   */
+  default PermissionChecker getPermissionChecker() {
+    return permission -> {
+      final Tristate state = getPermissionValue(permission);
+      if (state == Tristate.TRUE) {
+        return TriState.TRUE;
+      } else if (state == Tristate.UNDEFINED) {
+        return TriState.NOT_SET;
+      } else if (state == Tristate.FALSE) {
+        return TriState.FALSE;
+      } else {
+        throw new IllegalArgumentException();
+      }
+    };
+  }
 }
--- a/api/src/main/java/com/velocitypowered/api/permission/PermissionSubject.java
+++ b/api/src/main/java/com/velocitypowered/api/permission/PermissionSubject.java
@@ -39,17 +39,6 @@ public interface PermissionSubject {
    * @return subject's permission checker
    */
   default PermissionChecker getPermissionChecker() {
-    return permission -> {
-      final Tristate state = getPermissionValue(permission);
-      if (state == Tristate.TRUE) {
-        return TriState.TRUE;
-      } else if (state == Tristate.UNDEFINED) {
-        return TriState.NOT_SET;
-      } else if (state == Tristate.FALSE) {
-        return TriState.FALSE;
-      } else {
-        throw new IllegalArgumentException();
-      }
-    };
+    return permission -> getPermissionValue(permission).toAdventureTriState();
   }
 }
--- a/api/src/main/java/com/velocitypowered/api/proxy/Player.java
+++ b/api/src/main/java/com/velocitypowered/api/proxy/Player.java
@@ -23,16 +23,25 @@ import com.velocitypowered.api.util.title.Title;
 import java.util.List;
 import java.util.Optional;
 import java.util.UUID;
+import java.util.function.UnaryOperator;
 import net.kyori.adventure.identity.Identified;
 import net.kyori.adventure.identity.Identity;
+import net.kyori.adventure.key.Key;
+import net.kyori.adventure.key.Keyed;
+import net.kyori.adventure.pointer.Pointers;
 import net.kyori.adventure.text.Component;
+import net.kyori.adventure.text.ComponentLike;
+import net.kyori.adventure.text.event.HoverEvent;
+import net.kyori.adventure.text.event.HoverEventSource;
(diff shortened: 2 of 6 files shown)


api/src/main/java/com/velocitypowered/api/proxy/Player.java
```diff
@@ -23,16 +23,26 @@
...
 import org.checkerframework.checker.nullness.qual.Nullable;
+import org.jetbrains.annotations.NotNull;
 
 /**
  * Represents a player who is connected to the proxy.
  */
 public interface Player extends CommandSource, Identified, InboundConnection,
-    ChannelMessageSource, ChannelMessageSink {
+    ChannelMessageSource, ChannelMessageSink, HoverEventSource<HoverEvent.ShowEntity>,
+    ComponentLike, Keyed {
```
A248: A player is a ComponentLike? I'm confused why that would be the case. A player does not logically seem like a component to me.

(Also, that slightly weakens type-safety by making Player something API users might accidentally use in a place where ComponentLike is accepted)

api/src/main/java/com/velocitypowered/api/proxy/Player.java
```diff
@@ -23,16 +23,26 @@
...
+import net.kyori.adventure.text.event.HoverEventSource;
 import org.checkerframework.checker.nullness.qual.Nullable;
+import org.jetbrains.annotations.NotNull;
 
 /**
  * Represents a player who is connected to the proxy.
  */
 public interface Player extends CommandSource, Identified, InboundConnection,
-    ChannelMessageSource, ChannelMessageSink {
+    ChannelMessageSource, ChannelMessageSink, HoverEventSource<HoverEvent.ShowEntity>,
```
A248: Same thing for HoverEventSource. If I saw `component.hoverEvent(player)` in source code, I'd be somewhat puzzled.

A248 (commented):
Seems nice, I'm just not sure about some things

(edit: just so you know, I'm not a maintainer)

api/src/main/java/com/velocitypowered/api/proxy/Player.java
qixils: that's actually the intended use of HoverEventSource-- at least, that's how it's used in Paper. it indicates that a class is the Source of a HoverEvent, and makes a lot of sense in my eyes

api/src/main/java/com/velocitypowered/api/proxy/Player.java
qixils: you're definitely right, i'm just lazy and like avoiding ".asComponent()" in my code :P but for actual APIs people use you're right, i'll remove it

api/src/main/java/com/velocitypowered/api/proxy/Player.java
```diff
@@ -275,4 +283,36 @@ default void sendMessage(net.kyori.text.Component component) {
...
   boolean sendPluginMessage(ChannelIdentifier identifier, byte[] data);
+
+  @Override
+  default @NotNull Pointers pointers() {
+    return CommandSource.super.pointers().toBuilder()
+            .withDynamic(Identity.UUID, this::getUniqueId)
+            .withDynamic(Identity.NAME, this::getUsername)
+            .withDynamic(Identity.DISPLAY_NAME, this::getDisplayName)
+            .build();
+  }
```
kashike: This should be cached in the implementation class instead.

kashike (changes requested):

api/src/main/java/com/velocitypowered/api/proxy/Player.java
qixils: permissions as well?

kashike: yes

qixils: thank you, done :)

api/src/main/java/com/velocitypowered/api/proxy/Player.java
```diff
@@ -275,4 +282,27 @@ default void sendMessage(net.kyori.text.Component component) {
...
+  default @NotNull Key key() {
+    return Key.key("player");
+  }
+
+  /**
+   * Gets a {@link Component} that renders a player name similarly to vanilla.
+   *
+   * @return a Component representing this player
+   */
+  default @NotNull Component getDisplayName() {
```
astei: There have been prior discussions on including a display name component API in Velocity, but this isn't the way to go about it. In some cases, this is even entirely inappropriate because `/tell` might not exist or the user doesn't want to provide an insertion. The only sane implementation of this is `Component.text(getUsername())` and have this API hidden in the pointer, unless your plan is to add in the display name component API (which would have to be done for a future version of Velocity).

astei (changes requested):

api/src/main/java/com/velocitypowered/api/proxy/Player.java
kashike: I'd prefer we keep display name stuff out of this pull request, personally.

qixils: display name stuff has been removed 👍🏻

api/src/main/java/com/velocitypowered/api/permission/PermissionSubject.java
```diff
@@ -29,4 +32,24 @@ default boolean hasPermission(String permission) {
...
+
+  /**
+   * Gets the permission checker for the subject.
+   *
+   * @return subject's permission checker
+   */
+  default PermissionChecker getPermissionChecker() {
+    return permission -> {
+      final Tristate state = getPermissionValue(permission);
+      if (state == Tristate.TRUE) {
```
astei: Nit: Move this `if` block to the Velocity `Tristate` class, call it `toAdventureTriState()`.

astei (changes requested):
One minor change and this looks good to me.

qixils:
Should be good now :)

astei (approved):

PR #525, opened 2021-06-13
Expand PlayerChatEvent

qixils:
Adds additional options to the `PlayerChatEvent.ChatResult`, similarly to https://github.com/PaperMC/Paper/commit/ecb0d32. The notable exception is a `Set<Audience> recipients` as I was unable to find a good way to implement this in a way that could safely replace the `Destination` enum, though I'm open to suggestions ^^ for now i settled for allowing renderers to return `Component.empty()` to skip sending the message.

the unfortunate part of this PR is how many methods will prevent proxied servers from detecting the chat messages as real messages in their event listeners and whatnot. i don't know if all the javadoc warnings are unnecessary or if it would be considered too inappropriate for the proxy to break proxied servers' event handlers (only if a plugin uses it, of course) or if this is fine, so i figured it wouldn't hurt to try a PR :P


astei:
I'm not really convinced this is something that's within scope for Velocity. Velocity's main role in processing chat messages from the client are, principally speaking:

* Executing commands on the proxy if the player invokes them (and forwarding to the server when they don't exist or upon request).
* Allowing plugins to inspect and potentially modify chat messages sent to the server.

This uncomfortably crosses the territory between "basic essential functionality" and "bells and whistles better implemented in a plugin."

qixils:
Yeah, that makes a lot of sense and I definitely understand. Figured it wouldn't hurt to try 😅

PR #537, opened 2021-07-09
Custom localization for each player

davidmayr:
I have a network where the player can choose their language for the network independently of their client language.
Velocity won't cooperate with this, so I have added a method to switch their language with a method. 

If not set by a plugin, it will return the default language of the player in the settings object or null if the settings are null(On player login). The result of the getLocale method in the settings object will be unchanged by the set value.

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/proxy/Player.java
+++ b/api/src/main/java/com/velocitypowered/api/proxy/Player.java
@@ -47,9 +47,12 @@ public interface Player extends CommandSource, Identified, InboundConnection,
   String getUsername();
 
   /**
+   * The locale the velocity proxy is translating its messages to. By default {@link PlayerSettings#getLocale()}
    *
-   * @return the locale the velocity proxy is translating its messages to. By default {@link PlayerSettings#getLocale()}
-   * Can be null when the settings have not been initialized yet.
+   * <p>Can be null when the settings have not been initialized yet.</p>
+   *
+   * @return the locale.
+   *
    */
   @Nullable Locale getProxyLocale();
 
--- a/api/src/main/java/com/velocitypowered/api/proxy/Player.java
+++ b/api/src/main/java/com/velocitypowered/api/proxy/Player.java
@@ -49,14 +49,14 @@ public interface Player extends CommandSource, Identified, InboundConnection,
   /**
    * The locale the velocity proxy is translating its messages to. By default {@link PlayerSettings#getLocale()}
    *
-   * <p>Can be {@code null} when the settings have not been initialized yet.</p>
+   * <p>This can be {@code null} when the client has not yet connected to any server.</p>
    *
    * @return the locale.
    */
   @Nullable Locale getEffectiveLocale();
 
   /**
-   * Change the locale the proxy is translating its messages to.
+   * Change the locale the proxy will be translating its messages to.
    *
    * @param locale the locale to translate to
    */
--- a/api/src/main/java/com/velocitypowered/api/proxy/Player.java
+++ b/api/src/main/java/com/velocitypowered/api/proxy/Player.java
@@ -49,17 +49,18 @@ public interface Player extends CommandSource, Identified, InboundConnection,
   /**
    * The locale the velocity proxy is translating its messages to. By default {@link PlayerSettings#getLocale()}
    *
-   * <p>Can be null when the settings have not been initialized yet.</p>
+   * <p>Can be {@code null} when the settings have not been initialized yet.</p>
    *
    * @return the locale.
-   *
    */
-  @Nullable Locale getProxyLocale();
+  @Nullable Locale getEffectiveLocale();
 
-  /*
-  Change the locale the proxy is translating its messages to.
+  /**
+   * Change the locale the proxy is translating its messages to.
+   *
+   * @param locale the locale to translate to
    */
-  void setProxyLocale(Locale locale);
+  void setEffectiveLocale(Locale locale);
 
   /**
    * Returns the player's UUID.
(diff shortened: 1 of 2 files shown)


api/src/main/java/com/velocitypowered/api/proxy/Player.java
```diff
@@ -45,6 +46,20 @@
    */
   String getUsername();
 
+  /**
+   * The locale the velocity proxy is translating its messages to. By default {@link PlayerSettings#getLocale()}
+   *
+   * <p>Can be null when the settings have not been initialized yet.</p>
+   *
+   * @return the locale.
+   *
```
astei: Code style: don't have an empty line here.

api/src/main/java/com/velocitypowered/api/proxy/Player.java
```diff
@@ -45,6 +46,20 @@
...
+   */
+  @Nullable Locale getProxyLocale();
+
+  /*
+  Change the locale the proxy is translating its messages to.
```
astei: Not a complete Javadoc

api/src/main/java/com/velocitypowered/api/proxy/Player.java
```diff
@@ -45,6 +46,20 @@
    */
   String getUsername();
 
+  /**
+   * The locale the velocity proxy is translating its messages to. By default {@link PlayerSettings#getLocale()}
+   *
+   * <p>Can be null when the settings have not been initialized yet.</p>
```
astei:
```suggestion
   * <p>This can be {@code null} when the client has not yet connected to any server.</p>
```

astei (changes requested):

api/src/main/java/com/velocitypowered/api/proxy/Player.java
```diff
@@ -45,6 +46,20 @@
...
+   *
+   * @return the locale.
+   *
+   */
+  @Nullable Locale getProxyLocale();
```
astei: I don't like this name - perhaps `getEffectiveLocale()`?

astei (changes requested):

api/src/main/java/com/velocitypowered/api/proxy/Player.java
davidmayr: Sounds like a good name. Changed.

api/src/main/java/com/velocitypowered/api/proxy/Player.java
davidmayr: Fixed.

api/src/main/java/com/velocitypowered/api/proxy/Player.java
davidmayr: Should be complete now.

api/src/main/java/com/velocitypowered/api/proxy/Player.java
davidmayr: The line was removed.

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
```diff
@@ -178,6 +175,22 @@ public String getUsername() {
...
   }
 
+  @Override
+  public Locale getEffectiveLocale() {
+    if (effectiveLocale == null) {
+      if (settings == null) {
+        return null;
+      }
+      return settings.getLocale();
+    }
```
astei:
```suggestion
    if (effectiveLocale == null && settings != null) {
      return settings.getLocale();
    }
```

api/src/main/java/com/velocitypowered/api/proxy/Player.java
```diff
@@ -45,6 +46,21 @@
    */
   String getUsername();
 
+  /**
+   * The locale the velocity proxy is translating its messages to. By default {@link PlayerSettings#getLocale()}
```
astei: Still not satisfied with this Javadoc. The description is very imprecise - we are not necessarily _translating_ messages at this level.

api/src/main/java/com/velocitypowered/api/proxy/Player.java
```diff
@@ -45,6 +46,21 @@
...
+   *
+   * <p>Can be {@code null} when the settings have not been initialized yet.</p>
```
astei: This is still imprecise - the wording I originally suggested is better here.

astei (changes requested):

api/src/main/java/com/velocitypowered/api/proxy/Player.java
astei: This is not fixed

api/src/main/java/com/velocitypowered/api/proxy/Player.java
davidmayr: Oh, I overlooked that you did other changes than the code null thing. Is it good now?

api/src/main/java/com/velocitypowered/api/proxy/Player.java
davidmayr: Is it better now?

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
davidmayr: Changed.

api/src/main/java/com/velocitypowered/api/proxy/Player.java
astei: No, it's not. Perhaps `Returns the locale the proxy will use to send messages translated via the Adventure global translator. By default, the value of {@link PlayerSettings#getLocale()} is used.`.

davidmayr: Is it good now?

astei (approved):

PR #543, opened 2021-07-13
fix Cipher Block Chaining IV vulnerability

EzxD:
this seems to be a vulnerability
https://owasp.org/www-project-top-ten/2017/A6_2017-Security_Misconfiguration.html


astei:
Closing this issue, as this is a design flaw in the Minecraft protocol rather than a security issue in Velocity. I'd suggest you contact Mojang to report this issue. (Even then, I doubt they will consider this an actual security flaw in the protocol, and I do not consider Minecraft server connection encryption to be secure.)

PR #544, opened 2021-07-15
Expose CommandMeta in CommandManager and add a ref to the plugin instance

FrankHeijden:
Instead of having a reference to the plugin instance, a string with the plugin id might also be a nice idea -- but not sure which one would be preferred.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/command/VelocityCommandManager.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/command/VelocityCommandManager.java
@@ -20,6 +20,7 @@ package com.velocitypowered.proxy.command;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
 import com.mojang.brigadier.CommandDispatcher;
 import com.mojang.brigadier.ParseResults;
 import com.mojang.brigadier.exceptions.CommandSyntaxException;
@@ -40,6 +41,8 @@ import com.velocitypowered.proxy.command.registrar.SimpleCommandRegistrar;
 import com.velocitypowered.proxy.event.VelocityEventManager;
 import java.util.List;
 import java.util.Locale;
+import java.util.Map;
+import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
@@ -58,6 +61,7 @@ public class VelocityCommandManager implements CommandManager {
   private final List<CommandRegistrar<?>> registrars;
   private final SuggestionsProvider<CommandSource> suggestionsProvider;
   private final CommandGraphInjector<CommandSource> injector;
+  private final Map<String, CommandMeta> commandMeta;
 
   /**
    * Constructs a command manager.
@@ -75,6 +79,7 @@ public class VelocityCommandManager implements CommandManager {
             new RawCommandRegistrar(root, this.lock.writeLock()));
     this.suggestionsProvider = new SuggestionsProvider<>(this.dispatcher, this.lock.readLock());
     this.injector = new CommandGraphInjector<>(this.dispatcher, this.lock.readLock());
+    this.commandMeta = Maps.newConcurrentMap();
   }
 
   public void setAnnounceProxyCommands(boolean announceProxyCommands) {
@@ -134,6 +139,7 @@ public class VelocityCommandManager implements CommandManager {
       return false;
     }
     registrar.register(meta, superInterface.cast(command));
+    commandMeta.put(meta.getAliases().iterator().next(), meta);
     return true;
   }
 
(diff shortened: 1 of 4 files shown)


proxy/src/main/java/com/velocitypowered/proxy/command/VelocityCommandManager.java
```diff
@@ -75,6 +79,7 @@ public VelocityCommandManager(final VelocityEventManager eventManager) {
             new RawCommandRegistrar(root, this.lock.writeLock()));
     this.suggestionsProvider = new SuggestionsProvider<>(this.dispatcher, this.lock.readLock());
     this.injector = new CommandGraphInjector<>(this.dispatcher, this.lock.readLock());
+    this.commandMeta = Maps.newConcurrentMap();
```
astei: `new ConcurrentHashMap<>()`

astei (changes requested):
Sorry for leaving this too dry for so long. I've been busy.

hugmanrique (commented):
I'd prefer the plugin return types being `@Nullable Object` over optionals, Also, why do we only register the first alias? I would add all them to the map (to follow the javadoc). The worst thing is that the main alias of a returned meta object might not match the one given, but I don't see this as an issue. Everything else LGTM.

proxy/src/main/java/com/velocitypowered/proxy/command/VelocityCommandManager.java
```diff
@@ -142,7 +150,21 @@ public void unregister(final String alias) {
...
     // The literals of secondary aliases will preserve the children of
     // the removed literal in the graph.
-    dispatcher.getRoot().removeChildByName(alias.toLowerCase(Locale.ENGLISH));
+    String aliasLowercase = alias.toLowerCase(Locale.ENGLISH);
+    dispatcher.getRoot().removeChildByName(aliasLowercase);
+
+    CommandMeta meta = commandMetas.get(alias);
+    if (meta != null) {
+      for (String metaAlias : meta.getAliases()) {
+        commandMetas.remove(metaAlias);
```
astei: Nit: perhaps this should check if the meta is the same as the primary alias with `Map.remove(key, onlyIfValue)`.

proxy/src/main/java/com/velocitypowered/proxy/command/VelocityCommandMeta.java
```diff
@@ -30,19 +30,22 @@
...
 import java.util.stream.Stream;
+import org.checkerframework.checker.nullness.qual.Nullable;
 
 public final class VelocityCommandMeta implements CommandMeta {
 
   static final class Builder implements CommandMeta.Builder {
 
     private final ImmutableSet.Builder<String> aliases;
     private final ImmutableList.Builder<CommandNode<CommandSource>> hints;
+    private Object plugin;
```
astei:
```suggestion
    private @MonotonicNonNull Object plugin;
```

proxy/src/main/java/com/velocitypowered/proxy/command/VelocityCommandManager.java
```diff
@@ -148,6 +156,13 @@ public void unregister(final String alias) {
       // The literals of secondary aliases will preserve the children of
       // the removed literal in the graph.
       dispatcher.getRoot().removeChildByName(alias.toLowerCase(Locale.ENGLISH));
+
+      CommandMeta meta = commandMetas.get(alias);
+      if (meta != null) {
+        for (String metaAlias : meta.getAliases()) {
+          commandMetas.remove(metaAlias);
```
astei, line 163: Needs to check if the meta saved in the alias is equal to the original.

astei (changes requested):

astei (approved):

hugmanrique:
This should unconditionally remove the single mapping since we create a mapping from command aliases (not commands) to metas. As it stands, one may register the same command with two aliases, unregister one alias, and `getCommandMeta("otherAlias")` will incorrectly return `null` (doesn't follow the javadoc). https://github.com/VelocityPowered/Velocity/pull/544/files#diff-00c49e74f0fbdfecca0759479a72ed71560803e34e2ef3e0b4b6f686d69a3cfcR160

astei:
> This should unconditionally remove the single mapping since we create a mapping from command aliases (not commands) to metas. As it stands, one may register the same command with two aliases, unregister one alias, and `getCommandMeta("otherAlias")` will incorrectly return `null` (doesn't follow the javadoc). https://github.com/VelocityPowered/Velocity/pull/544/files#diff-00c49e74f0fbdfecca0759479a72ed71560803e34e2ef3e0b4b6f686d69a3cfcR160

This is also incorrect - `unregister(String)` unregisters a _single_ alias. Perhaps we should provide an `unregister(CommandMeta)` as well to unregister every alias pointing to a given `CommandMeta`.

hugmanrique:
That's my point, right now `unregister("a")` will unregister the meta mappings for any other aliases for the same command (maybe I misunderstood your message, sorry if so 😛). To be clear, the diff above links to this PR, it's not my proposed changes.

On `unregister(CommandMeta)`, that's a good idea.

astei:
Fixed that in 0b0c36d, thanks!

Issue #547, opened 2021-07-19
Minor breakages in event listener registration


A248:
With the rewrite of the EventManager implementation in Velocity 3, behavior has changed.

The `unregisterListener(Object, Object)` no longer allows unregistering annotated listeners, that is, listener objects which use annotated methods to declare handlers. For example:

```java
Object annotatedListener = new AnnotatedListener();
eventManager.register(PLUGIN_A, annotatedListener);
eventManager.unregisterListener(PLUGIN_A, annotatedListener);
```

In Velocity 1.x, the listener will be unregistered. In 3.x, the listener remains registered.

If this is an intended behavioral change, it would be worthwhile to mention it on the [migration docs](https://github.com/VelocityPowered/site/blob/master/docs/developers/porting-from-velocity-1.md).

astei:
This sounds like a regression to me.

PR #548, opened 2021-07-19
Demonstrate Two Bugs in 3.0 VelocityEventManager

A248:
The first issue is that annotated listener objects cannot be unregistered via `unregisterListener(Object, Object)`, which results in 4 test failures here. See also #547 . The cause relates to that an annotated listener will not be equal in instance to its generated EventHandler, so the following condition cannot possibly be satisfied:

https://github.com/VelocityPowered/Velocity/blob/f79736b8b759b4d7d0de36c87bbdebb579ada69d/proxy/src/main/java/com/velocitypowered/proxy/event/VelocityEventManager.java#L409-L410

The second problem, which was not mentioned in #547 , is that unregistration does not properly invalidate the handlers cache when the fired event is a subclass of the event listened to. This is demonstrated by the `simpleRegisterAndUnregister` test, which fails two executions. Changing the following source lines to a blanket cache invalidation, i.e. `handlersCache.invalidateAll()`, fixes the relevant failures:

https://github.com/VelocityPowered/Velocity/blob/f79736b8b759b4d7d0de36c87bbdebb579ada69d/proxy/src/main/java/com/velocitypowered/proxy/event/VelocityEventManager.java#L434-L438

-----------------------------------------

Applying these tests to Velocity 1.1.x, they all pass.

Note that this PR probably does not pass checkstyle. It is primarily intended as a bug demonstration.

### DIFF
--- a/proxy/src/test/java/com/velocitypowered/proxy/event/VelocityEventManagerTest.java
+++ b/proxy/src/test/java/com/velocitypowered/proxy/event/VelocityEventManagerTest.java
@@ -1,0 +1,197 @@
+/*
+ * Copyright (C) 2018 Velocity Contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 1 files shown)


astei:
Fixed in e017949abf775cace3f0c63b6e7f9f576d77ff93, incorporating these tests. Thanks!

PR #556, opened 2021-08-05
added a simple team api

booky10:
I had created a similar team api for my personal fork of velocity, but thought that it could be in velocity itself. I have tested this api with a simple plugin ([booky10/VelocityTeamTest](https://github.com/booky10/VelocityTeamTest)), which worked as expected:

Screenshot of tablist

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/proxy/ProxyServer.java
+++ b/api/src/main/java/com/velocitypowered/api/proxy/ProxyServer.java
@@ -17,6 +17,7 @@ import com.velocitypowered.api.proxy.player.ResourcePackInfo;
 import com.velocitypowered.api.proxy.server.RegisteredServer;
 import com.velocitypowered.api.proxy.server.ServerInfo;
 import com.velocitypowered.api.scheduler.Scheduler;
+import com.velocitypowered.api.team.TeamManager;
 import com.velocitypowered.api.util.ProxyVersion;
 import java.net.InetSocketAddress;
 import java.util.Collection;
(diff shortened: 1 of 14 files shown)


astei:
I know @Xernium was working on a fully-featured scoreboard API for Velocity - hopefully he can throw up a draft PR of what he has now.

I'm not sure what the best approach is to take here, since the team API at this point would be highly specific to Velocity. In general, I would prefer the API be fleshed out in Adventure for the benefit of all Adventure users. Yet in past discussions with the Adventure developers, they are not willing to add much more than basic manipulation APIs.

For now, while we sort out these issues, I'm going to place this PR on hold.

booky10:
Ok, feel free to close this PR if you think this change wouldn't fit velocity.

NEZNAMY:
I may have overlooked something, but it looks like this will disconnect <1.13 players due to display name, prefix and suffix being plain text with a limit of 16 characters, not components.

booky10:
Should be fixed now, tested with [my testing plugin](https://github.com/booky10/VelocityTeamTest) and minecraft client/server version 1.8.8, 1.9.4, 1.12.2 and 1.17.1.

NEZNAMY:
Does it account for 16 character limit as well? Since with the way API is currently made you can't cut it only for legacy clients.

booky10:
Since Xernium has recently continued his scoreboard api (and his approach to the package structure is much better than mine), I don't think this pull request should remain open much longer.

PR #565, opened 2021-08-24
Fix brigadier command redirect not work

extclp:


astei:
Fixed in a more concise matter, thanks for the patch.

PR #567, opened 2021-08-28
Introduce `PlayerClientBrandEvent`

drunderscore:
Fired when the player sends the `minecraft:brand` (or equivalent) plugin message.  If the result is
`DENIED`, the player brand is not set (`setClientBrand` is not called),
and it is not forwarded to the backend.

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/event/player/PlayerClientBrandEvent.java
+++ b/api/src/main/java/com/velocitypowered/api/event/player/PlayerClientBrandEvent.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright (C) 2018 Velocity Contributors
+ *
+ * The Velocity API is licensed under the terms of the MIT License. For more details,
+ * reference the LICENSE file in the api top-level directory.
+ */
+
+package com.velocitypowered.api.event.player;
+
+import com.google.common.base.Preconditions;
+import com.velocitypowered.api.proxy.Player;
+
+/**
+ * Fired when a {@link Player} sends the <code>minecraft:brand</code> plugin message.
+ */
+public final class PlayerClientBrandEvent {
+  private final Player player;
+  private final String brand;
+
+  /**
+   * Creates a new instance.
+   *
+   * @param player the {@link Player} of the sent client brand
+   * @param brand the sent client brand
+   */
+  public PlayerClientBrandEvent(Player player, String brand) {
+    this.player = Preconditions.checkNotNull(player);
+    this.brand = Preconditions.checkNotNull(brand);
+  }
+
+  public Player getPlayer() {
+    return player;
+  }
+
+  public String getBrand() {
+    return brand;
+  }
+
+  @Override
+  public String toString() {
+    return "PlayerClientBrandEvent{"
+      + "player=" + player
+      + ", brand='" + brand + '\''
+      + '}';
+  }
+}
+
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientPlaySessionHandler.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientPlaySessionHandler.java
@@ -27,6 +27,7 @@ import com.velocitypowered.api.event.command.CommandExecuteEvent.CommandResult;
 import com.velocitypowered.api.event.connection.PluginMessageEvent;
 import com.velocitypowered.api.event.player.PlayerChannelRegisterEvent;
 import com.velocitypowered.api.event.player.PlayerChatEvent;
+import com.velocitypowered.api.event.player.PlayerClientBrandEvent;
 import com.velocitypowered.api.event.player.PlayerResourcePackStatusEvent;
 import com.velocitypowered.api.event.player.TabCompleteEvent;
 import com.velocitypowered.api.network.ProtocolVersion;
@@ -229,7 +230,9 @@ public class ClientPlaySessionHandler implements MinecraftSessionHandler {
         player.getKnownChannels().removeAll(PluginMessageUtil.getChannels(packet));
         backendConn.write(packet.retain());
       } else if (PluginMessageUtil.isMcBrand(packet)) {
-        player.setClientBrand(PluginMessageUtil.readBrandMessage(packet.content()));
+        String brand = PluginMessageUtil.readBrandMessage(packet.content());
+        server.getEventManager().fireAndForget(new PlayerClientBrandEvent(player, brand));
+        player.setClientBrand(brand);
         backendConn.write(PluginMessageUtil
             .rewriteMinecraftBrand(packet, server.getVersion(), player.getProtocolVersion()));
       } else if (BungeeCordMessageResponder.isBungeeCordMessage(packet)) {


drunderscore:
Fix the wrong year in the copyright header

api/src/main/java/com/velocitypowered/api/event/player/PlayerClientBrandEvent.java
```diff
@@ -0,0 +1,62 @@
...
+
+import com.google.common.base.Preconditions;
+import com.velocitypowered.api.event.ResultedEvent;
+import com.velocitypowered.api.event.connection.PluginMessageEvent;
+import com.velocitypowered.api.proxy.Player;
+
+/**
+ * Fired when a {@link Player} sends the <code>minecraft:brand</code> plugin message.
+ */
+public class PlayerClientBrandEvent implements ResultedEvent<PluginMessageEvent.ForwardResult> {
```
astei: I don't see any purpose to allow this event to be cancelled. Why would you want to do that?

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientPlaySessionHandler.java
```diff
@@ -229,9 +230,14 @@ public boolean handle(PluginMessage packet) {
         player.getKnownChannels().removeAll(PluginMessageUtil.getChannels(packet));
         backendConn.write(packet.retain());
       } else if (PluginMessageUtil.isMcBrand(packet)) {
-        player.setClientBrand(PluginMessageUtil.readBrandMessage(packet.content()));
-        backendConn.write(PluginMessageUtil
-            .rewriteMinecraftBrand(packet, server.getVersion(), player.getProtocolVersion()));
+        var brand = PluginMessageUtil.readBrandMessage(packet.content());
```
astei: Inconsistent with the rest of the code here, which does not use `var`.

astei (changes requested):

api/src/main/java/com/velocitypowered/api/event/player/PlayerClientBrandEvent.java
drunderscore: No clue! Honestly I just copied what most events did which was that they were cancelable. Can't think of a reason you'd want to cancel it :^)

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientPlaySessionHandler.java
drunderscore: no `var` :(
No problem, consistency over everything for sure.

drunderscore:
Event is no longer cancelable, and the commit is now consistent with the reset of the codebase's lack of `var`.

Only issue is that the copyright year in `PlayerClientBrandEvent.java` is wrong (2018). However the licenser doesn't accept any other year, as the year is baked into `HEADER.txt` as 2018.

api/src/main/java/com/velocitypowered/api/event/player/PlayerClientBrandEvent.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+package com.velocitypowered.api.event.player;
+
+import com.velocitypowered.api.proxy.Player;
+
+/**
+ * Fired when a {@link Player} sends the <code>minecraft:brand</code> plugin message.
+ */
+public class PlayerClientBrandEvent {
```
A248: In tandem with preconditions (see other comment), a final class prevents sub-classes from overriding methods and breaking contracts. Other events do this as well.
```suggestion
public final class PlayerClientBrandEvent {
```

api/src/main/java/com/velocitypowered/api/event/player/PlayerClientBrandEvent.java
```diff
@@ -0,0 +1,46 @@
...
+
+  /**
+   * Creates a new instance.
+   *
+   * @param player the {@link Player} of the sent client brand
+   * @param brand the sent client brand
+   */
+  public PlayerClientBrandEvent(Player player, String brand) {
+    this.player = player;
+    this.brand = brand;
```
A248: I believe there should be preconditions here, i.e. Preconditions.checkNotNull as the other events in the API use.

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientPlaySessionHandler.java
```diff
@@ -229,9 +230,8 @@ public boolean handle(PluginMessage packet) {
         player.getKnownChannels().removeAll(PluginMessageUtil.getChannels(packet));
         backendConn.write(packet.retain());
       } else if (PluginMessageUtil.isMcBrand(packet)) {
-        player.setClientBrand(PluginMessageUtil.readBrandMessage(packet.content()));
-        backendConn.write(PluginMessageUtil
-            .rewriteMinecraftBrand(packet, server.getVersion(), player.getProtocolVersion()));
+        String brand = PluginMessageUtil.readBrandMessage(packet.content());
+        server.getEventManager().fire(new PlayerClientBrandEvent(player, brand));
```
A248: After the event is fired, player.setClientBrand and backendConn.write should still be called.

drunderscore: whoops those accidentally disappeared in me making it non-cancelable!

drunderscore:
Changes made based on @A248 feedback, commit message updated to not reference the cancelability of `PlayerClientBrandEvent` as it does not exist.

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientPlaySessionHandler.java
```diff
@@ -229,9 +230,11 @@ public boolean handle(PluginMessage packet) {
         player.getKnownChannels().removeAll(PluginMessageUtil.getChannels(packet));
         backendConn.write(packet.retain());
       } else if (PluginMessageUtil.isMcBrand(packet)) {
+        String brand = PluginMessageUtil.readBrandMessage(packet.content());
+        server.getEventManager().fire(new PlayerClientBrandEvent(player, brand));
         player.setClientBrand(PluginMessageUtil.readBrandMessage(packet.content()));
```
A248:
```suggestion
        player.setClientBrand(brand);
```
Strings are immutable.

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientPlaySessionHandler.java
```diff
@@ -229,9 +230,11 @@ public boolean handle(PluginMessage packet) {
...
         backendConn.write(PluginMessageUtil
-            .rewriteMinecraftBrand(packet, server.getVersion(), player.getProtocolVersion()));
+                .rewriteMinecraftBrand(packet, server.getVersion(), player.getProtocolVersion()));
```
A248: Unnecessary formatting change

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientPlaySessionHandler.java
```diff
@@ -229,7 +230,9 @@ public boolean handle(PluginMessage packet) {
         player.getKnownChannels().removeAll(PluginMessageUtil.getChannels(packet));
         backendConn.write(packet.retain());
       } else if (PluginMessageUtil.isMcBrand(packet)) {
-        player.setClientBrand(PluginMessageUtil.readBrandMessage(packet.content()));
+        String brand = PluginMessageUtil.readBrandMessage(packet.content());
+        server.getEventManager().fire(new PlayerClientBrandEvent(player, brand));
```
A248:
```suggestion
        server.getEventManager().fireAndForget(new PlayerClientBrandEvent(player, brand));
```
I forgot about this, it seems.

A248 (approved):
Besides fire versus fireAndForget, I approve. (disclaimer: I am not a core Velocity contributor)

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientPlaySessionHandler.java
drunderscore: Makes sense to me -- it's done above for `PlayerChannelRegisterEvent` as well.

astei (approved):

PR #571, opened 2021-09-01
Fix legacy handshake with paper 1.16.5 and higher, update jenkins link

mdxd44:
If you have Minecraft servers on different versions (1.12.2 and 1.16.5 in our case), you can only use "bungeeguard" or "legacy" forwarding type. If you use the "bungeeguard" type, then everything will be ok, but if you use the "legacy" type, then players, who have an IP with a hostname (like 127.0.0.1.your-provider.net), won't be able to connect to the Paper 1.16.5+ Minecraft server. Players will see an error message "`If you wish to use IP forwarding, please enable it in your BungeeCord config as well!`"

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Velocity
 
-[![Build Status](https://img.shields.io/jenkins/s/https/ci.velocitypowered.com/job/velocity.svg)](https://ci.velocitypowered.com/job/velocity/)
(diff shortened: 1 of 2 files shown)


astei:
I'm genuinely confused what this is supposed to fix. The code path you've modified is invoked when player info forwarding is not in use.

mdxd44:
> 
> 
> I'm genuinely confused what this is supposed to fix. The code path you've modified is invoked when player info forwarding is not in use.

I'm sorry, I changed the wrong method, check [link](https://github.com/PaperMC/Waterfall/pull/679/commits/fdc3f3b8ff5a0459d8d77b3c86b0e1ebbb1e3ad2),

PR #578, opened 2021-09-29
Add ProxyExceptionEvent to allow for outsourcing exception logging

FrankHeijden:
Adds ProxyExceptionEvent in a similar fashion to [Waterfall's implementation](https://papermc.io/javadocs/waterfall/1.17/io/github/waterfallmc/waterfall/event/ProxyExceptionEvent.html).

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/event/proxy/ProxyExceptionEvent.java
+++ b/api/src/main/java/com/velocitypowered/api/event/proxy/ProxyExceptionEvent.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright (C) 2018 Velocity Contributors
+ *
+ * The Velocity API is licensed under the terms of the MIT License. For more details,
+ * reference the LICENSE file in the api top-level directory.
+ */
+
+package com.velocitypowered.api.event.proxy;
+
+import com.google.common.base.Preconditions;
+import com.velocitypowered.api.proxy.exception.ProxyException;
+
+/**
+ * This event is fired by the proxy when an exception is thrown
+ * in a recoverable section of the server.
+ */
+public class ProxyExceptionEvent {
+
+  private final ProxyException exception;
+
+  public ProxyExceptionEvent(ProxyException exception) {
+    this.exception = Preconditions.checkNotNull(exception, "exception");
+  }
+
+  /**
+   * Gets the wrapped exception that was thrown.
+   *
+   * @return Thrown exception
+   */
+  public ProxyException getException() {
+    return exception;
+  }
+
+  @Override
+  public String toString() {
+    return "ProxyExceptionEvent{"
+        + "exception=" + exception
+        + '}';
+  }
+}
--- a/api/src/main/java/com/velocitypowered/api/proxy/exception/ProxyCommandException.java
+++ b/api/src/main/java/com/velocitypowered/api/proxy/exception/ProxyCommandException.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright (C) 2018 Velocity Contributors
+ *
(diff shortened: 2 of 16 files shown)


A248:
This is quite a lot of boilerplate and a non-negligible maintenance burden for such a feature, but I'm not to be the judge of that considering I'm not a core maintainer.

Since an event like this one is most commonly used for server administration purposes, why not create a Log4j appender? With an appender you'd have access to all exceptions, not just those from velocity. Of course, you wouldn't have access to context like a Player or command; however, I don't understand why you want to use these objects. Could you elaborate?

hugmanrique:
In addition, the composition over inheritance of these exception classes, apart of not being the norm in Java, can hurt the GC. Would be interesting to benchmark the impact with bots force-causing disconnect exceptions.

FrankHeijden:
Initially, when I thought of this addition was to provide an event which would just expose the `Throwable` through a getter. However when looking at the waterfall implementation I saw they used wrappers around exceptions to expose more metadata around the exceptions, which I thought could serve more use cases of this event.

PaulBGD:
I implemented something like this recently to add Sentry support, using log4j is fairly straightforward. Send me a DM and I can share some code.

TheMeinerLP:
I solve that problem with following tech stack and config.
I host an internal Loki server for my logs and use promtail to collect them.
Besides, I made a custom config for log4j that made the logging format for the file to logfmt.
log4j2.xml:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<!--
    Copyright (C) 2018 Velocity Contributors

    This program is free software: you can redistribute it and/or modify
    it under the terms of the GNU General Public License as published by
    the Free Software Foundation, either version 3 of the License, or
    (at your option) any later version.

    This program is distributed in the hope that it will be useful,
    but WITHOUT ANY WARRANTY; without even the implied warranty of
    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
    GNU General Public License for more details.

    You should have received a copy of the GNU General Public License
    along with this program.  If not, see <https://www.gnu.org/licenses/>.
-->

<!-- Disable shutdown hook, because we have our own -->
<Configuration status="warn" shutdownHook="disable">
  <Appenders>
    <TerminalConsole name="TerminalConsole">
      <PatternLayout>
        <LoggerNamePatternSelector
          defaultPattern="%highlightError{[%d{HH:mm:ss} %level] [%logger]: %msg%n%xEx}">
          <!-- Velocity doesn't need a prefix -->
          <PatternMatch key="com.velocitypowered."
            pattern="%highlightError{[%d{HH:mm:ss} %level]: %msg%n%xEx}"/>
        </LoggerNamePatternSelector>
      </PatternLayout>
    </TerminalConsole>
    <RollingRandomAccessFile name="File" fileName="logs/latest.log"
      filePattern="logs/%d{yyyy-MM-dd}-%i.log.gz"
      immediateFlush="false">
      <PatternLayout
        pattern="time=&quot;%d{HH:mm:ss}&quot; thread=&quot;%t&quot; level=&quot;%level&quot; logger=&quot;%logger&quot; message=&quot;%stripAnsi{%msg}&quot;%n"/>
      <Policies>
        <TimeBasedTriggeringPolicy/>
        <OnStartupTriggeringPolicy/>
      </Policies>
    </RollingRandomAccessFile>
  </Appenders>

  <Loggers>
    <Root level="info">
      <AppenderRef ref="TerminalConsole"/>
      <AppenderRef ref="File"/>
    </Root>
  </Loggers>
</Configuration>
```

Place that config inside your velocity root folder and use at the start of the proxy following parameter before `-jar`: `-Dlog4j2.configurationFile=./log4j2.xml`

PR #580, opened 2021-10-02
Don't allow illegal characters

Foorcee:
The Minecraft server will not accept a chat packet from a player that contains an invalid character. 
If the chat packet is handled on Velocity, such as a command, it doesn't check if the message is valid.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientPlaySessionHandler.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientPlaySessionHandler.java
@@ -640,6 +640,8 @@ public class ClientPlaySessionHandler implements MinecraftSessionHandler {
   }
 
   private boolean isAllowedCharacter(char c) {
+    // 167 = §, 127 = DEL
+    // https://minecraft.fandom.com/wiki/Multiplayer#Chat
     return c != 167 && c >= ' ' && c != 127;
   }
 }
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientPlaySessionHandler.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientPlaySessionHandler.java
@@ -153,6 +153,15 @@ public class ClientPlaySessionHandler implements MinecraftSessionHandler {
     }
 
     String msg = packet.getMessage();
+    for (int i = 0; i < msg.length(); i++) {
+      if (!isAllowedCharacter(msg.charAt(i))) {
+        player.disconnect(Component.translatable("velocity.error.illegal-chat-characters",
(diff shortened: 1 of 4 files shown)


proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientPlaySessionHandler.java
```diff
@@ -629,4 +638,8 @@ public void flushQueuedMessages() {
       }
     }
   }
+
+  private boolean isAllowedCharacter(char c) {
```
hugmanrique: Nice catch! Could you document this function noting the source (apart from the client source) like wiki.vg for the allowed characters?

hugmanrique:
I'm not 100% sure on how the translations currently work, but I think you also need to add the corresponding entry to https://github.com/VelocityPowered/Velocity/blob/dev/3.0.0/proxy/src/main/resources/com/velocitypowered/proxy/l10n/messages.properties

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientPlaySessionHandler.java
```diff
@@ -153,6 +153,15 @@ public boolean handle(Chat packet) {
     }
 
     String msg = packet.getMessage();
+    for (int i = 0; i < msg.length(); i++) {
```
astei: Please make this a unit test, ideally by extracting this out into a static method that just does the character validity check.

Foorcee:
I have added the translation. I am not sure how the translation system works.

astei (changes requested):

Foorcee:
@astei What should I do with the new translation?

astei:
Don't worry about it, I'll handle it.

proxy/src/test/java/com/velocitypowered/proxy/util/CharacterUtilTest.java
```diff
@@ -0,0 +1,50 @@
...
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import org.junit.jupiter.api.Test;
+
+class CharacterUtilTest {
+
+  private static final String CHARACTERS = "!\\\"#$%&'()*+,-./0123456789:;<=>?"
+      + "@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\\\]^_'abcdefghijklmnopqrstuvwxyz¡«»";
+
```
astei, line 30: This test is insufficient. We should also make sure the check does not reject strings with non-ASCII characters in them, such as CJK (Chinese/Japanese/Korean).

astei (changes requested):

PR #581, opened 2021-10-03
Redirect player without message

### LINKED ISSUE #392: Server kick redirection message enhancements
Hello, 
It would be useful if you could also specify a message in the player's redirection if it was successful.

It would also be convenient if you don't want to display a message.

Currently there is already a message, it seems to be for when the connection failed. 

https://github.com/VelocityPowered/Velocity/blob/9825f5891bf2ec19facddc829726255540d92000/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java#L634

The message attribute is also missing in the java docs.

https://github.com/VelocityPowered/Velocity/blob/dev/1.1.0/api/src/main/java/com/velocitypowered/api/event/player/KickedFromServerEvent.java#L235

Foorcee:
Fixes #392

I don't know if this is the best solution, but I tried not to change the existing behavior as much as possible. 
I also improved/corrected the javadocs.

Overall it would be best if the message could be set based on the connection state. Since that would require more redesigning, so this could be best added to the polymer.

Feedback is welcome

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/event/player/KickedFromServerEvent.java
+++ b/api/src/main/java/com/velocitypowered/api/event/player/KickedFromServerEvent.java
@@ -136,13 +136,11 @@ public final class KickedFromServerEvent implements
 
     private final net.kyori.adventure.text.Component message;
     private final RegisteredServer server;
-    private final boolean sendNotification;
 
     private RedirectPlayer(RegisteredServer server,
-        net.kyori.adventure.text.@Nullable Component message, boolean sendNotification) {
+        net.kyori.adventure.text.@Nullable Component message) {
       this.server = Preconditions.checkNotNull(server, "server");
       this.message = message;
-      this.sendNotification = sendNotification;
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


astei:
Instead of adding another parameter to the `RedirectPlayer`, couldn't we just do a check to see if the message is `Component.empty()` and just don't print a message in this case?

Foorcee:
That was basically my first idea as well.  I just thought it would be inconsistent. If the message is null, the kick reason is sent as a message. If empty, no message is sent and otherwise a message is sent again. But this is the simplest solution, if it fits so I would change it again.

api/src/main/java/com/velocitypowered/api/event/player/KickedFromServerEvent.java
```diff
@@ -159,15 +158,36 @@ public RegisteredServer getServer() {
...
     }
 
+    /**
+     * Creates a new redirect result to forward the player to the specified {@code server}.
+     * The kick reason will be displayed to the player
+     *
+     * @param server the server to send the player to
+     * @return the redirect result
+     */
+    @Deprecated
```
astei: Why is this deprecated?

api/src/main/java/com/velocitypowered/api/event/player/KickedFromServerEvent.java
```diff
@@ -159,15 +158,36 @@ public RegisteredServer getServer() {
 
     /**
      * Creates a new redirect result to forward the player to the specified {@code server}.
+     * No message will be sent from the proxy.
      *
      * @param server the server to send the player to
      * @return the redirect result
      */
+    public static RedirectPlayer connect(RegisteredServer server) {
```
astei: Probably not necessary, a note on the two-parameter constructor would be sufficient, such as:

```
* Use {@code Component.empty()} to skip sending any messages to the player.
```

astei (changes requested):

astei (approved):

PR #582, opened 2021-10-03
feat: added VELOCITY_FORWARDING_SECRET environment variable

Cubxity:
This PR allows Velocity forwarding secret to be set at runtime, instead of being defined in the config.
Therefore, it is perfect for containerized environments, such as: Kubernetes and Docker.

Related issue(s):
https://github.com/PaperMC/Paper/issues/6332

*PS: Please mark this PR as `hacktoberfest-accepted` once accepted.*

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
@@ -426,15 +426,12 @@ public class VelocityConfiguration implements ProxyConfig {
 
     // Retrieve the forwarding secret. First, from environment variable, then from config.
     byte[] forwardingSecret;
-    String forwardingSecretString = System.getenv("VELOCITY_FORWARDING_SECRET");
+    String forwardingSecretString = System.getenv()
+        .getOrDefault("VELOCITY_FORWARDING_SECRET", config.get("forwarding-secret"));
     if (forwardingSecretString == null || forwardingSecretString.isEmpty()) {
-      forwardingSecretString = config.get("forwarding-secret");
-
-      if (forwardingSecretString == null || forwardingSecretString.isEmpty()) {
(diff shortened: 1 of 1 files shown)


proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
```diff
@@ -424,16 +424,21 @@ public static VelocityConfiguration read(Path path) throws IOException {
     CommentedFileConfig defaultConfig = CommentedFileConfig.of(tmpFile, TomlFormat.instance());
     defaultConfig.load();
 
-    // Handle any cases where the config needs to be saved again
+    // Retrieve the forwarding secret. First, from environment variable, then from config.
     byte[] forwardingSecret;
-    String forwardingSecretString = config.get("forwarding-secret");
+    String forwardingSecretString = System.getenv("VELOCITY_FORWARDING_SECRET");
```
NoahvdAa:
```suggestion
    String forwardingSecretString = System.getenv("VELOCITY_FORWARDING_SECRET", config.get("forwarding-secret"));
```
This would allow you to remove the the first if block (+ the config fetch in line 431 ofc)

Cubxity: `System#getenv` does not provide a "default" parameter.

NoahvdAa: Whoops, my bad, confused it with system properties. Here's what I should've suggested :D
```suggestion
    String forwardingSecretString = System.getenv().getOrDefault("VELOCITY_FORWARDING_SECRET", config.get("forwarding-secret")
```

astei (approved):

misuzu:
It's weird that it's possible to set a velocity secret but it's not possible to actually enable velocity this way...

PR #587, opened 2021-10-24
Add support for sending and receiving login plugin messages from players and servers

### LINKED ISSUE #570: Can't send Login custom payloads to client while using Velocity
I've recently been messing around with sending packets from paper to modded clients and needed to send some data during the `LOGIN` phase to prepare for the `PLAY` phase. This would be the ClientboundCustomQueryPacket (mojmap) or PacketLoginInCustomPayload (spigot). It's the same packet that's used to add velocity support.

after poking around on wiki.vg and velocity the equivalent would be 
https://github.com/VelocityPowered/Velocity/blob/dev/3.0.0/proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/LoginPluginMessage.java

I look and see where it's being handled
https://github.com/VelocityPowered/Velocity/blob/dev/3.0.0/proxy/src/main/java/com/velocitypowered/proxy/connection/backend/LoginSessionHandler.java#L86

and it seems anything that isn't the velocity info packet is just discarded or rather sent back to the server empty. 

I'm wondering is this an intentional thing? is it some limitation of going through a proxy that CustomQueryPackets can't be sent to the client? Normally i'd just spend a few days messing around with the code and see if I could get it to work, but in this instance I think I'd rather ask.

astei:
Fixes #570.

This support will be useful mostly for plugins aiming to enable support for certain client mods.

### DIFF
--- a/proxy/build.gradle
+++ b/proxy/build.gradle
@@ -68,6 +68,7 @@ dependencies {
 
     implementation(platform("net.kyori:adventure-bom:${adventureVersion}"))
     implementation("net.kyori:adventure-nbt")
+    implementation("net.kyori:adventure-text-minimessage:4.1.0-SNAPSHOT")
 
     implementation 'org.asynchttpclient:async-http-client:2.12.3'
 
@@ -79,6 +80,9 @@ dependencies {
     implementation 'org.lanternpowered:lmbda:2.0.0'
 
     implementation 'com.github.ben-manes.caffeine:caffeine:3.0.3'
+
+    implementation 'space.vectrix.flare:flare:2.0.0'
+    implementation 'space.vectrix.flare:flare-fastutil:2.0.0'
 
     compileOnly 'com.github.spotbugs:spotbugs-annotations:4.4.0'
 
--- a/proxy/build.gradle
+++ b/proxy/build.gradle
@@ -68,7 +68,6 @@ dependencies {
 
     implementation(platform("net.kyori:adventure-bom:${adventureVersion}"))
     implementation("net.kyori:adventure-nbt")
-    implementation("net.kyori:adventure-text-minimessage:4.1.0-SNAPSHOT")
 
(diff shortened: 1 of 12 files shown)


proxy/build.gradle
```diff
@@ -68,6 +68,7 @@ dependencies {
 
     implementation(platform("net.kyori:adventure-bom:${adventureVersion}"))
     implementation("net.kyori:adventure-nbt")
+    implementation("net.kyori:adventure-text-minimessage:4.1.0-SNAPSHOT")
```
MrIvanPlays: Why don't you expose mini message to the API?

astei: The inclusion of MiniMessage in this PR was a mistake. MiniMessage won't be added to Velocity until Adventure 4.10 when it will become stable.

astei:
[Here's the test plugin I made](https://github.com/astei/origins-velocity-adapter) to test this PR. This enables the popular [Origins mod](https://www.curseforge.com/minecraft/mc-mods/origins) to work on Velocity (combined with the CrossStitch mod) without further modifications.

tadhgboyle (approved):

proxy/build.gradle
```diff
@@ -80,6 +80,9 @@ dependencies {
 
     implementation 'com.github.ben-manes.caffeine:caffeine:3.0.3'
 
+    implementation 'space.vectrix.flare:flare:2.0.0'
+    implementation 'space.vectrix.flare:flare-fastutil:2.0.0'
```
CoreyShupe, line 84: https://mvnrepository.com/artifact/it.unimi.dsi/fastutil is there a reason behind using flare other than this fastutil impl which is used in paper as well iirc?

astei: Flare provides a general thread-safe wrapper around any kind of map with superior performance over synchronizing on the map, especially in this case where the keys are disjoint.

The Flare map in this case wraps a normal Int2ObjectOpenHashMap.

PR #588, opened 2021-11-02
Introduce VelocityBrigadierMessage

MrIvanPlays:
This gives support to commands' argument suggestions for tooltip messages by implementing `com.brigadier.Message` to adapt with adventure. Allows to do things like this:

image

Code used to test:
```java
final LiteralArgumentNode<CommandSource> node = BrigadierCommand.literalArgumentBuilder("test")
      .then(BrigadierCommand.requiredArgumentBuilder("something", StringArgumentType.string())
              .suggests(
                  (context, builder) -> {
                    builder.suggest(
                        "iamfancy",
                        VelocityBrigadierMessage.tooltip(
                            Component.text("This is a fancy tooltip", NamedTextColor.GREEN)
                        )
                    );
                    return builder.buildFuture();
                 })
              .executes(
                  context -> {
                    context.getSource().sendMessage(Component.text("Hello!"));
                    return 1;
                  })
              .build();
proxy.getCommandManager().register(new BrigadierCommand(node));
```

Tested on 1.17.1

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/command/VelocityBrigadierMessage.java
+++ b/api/src/main/java/com/velocitypowered/api/command/VelocityBrigadierMessage.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright (C) 2018 Velocity Contributors
+ *
+ * The Velocity API is licensed under the terms of the MIT License. For more details,
+ * reference the LICENSE file in the api top-level directory.
+ */
+
+package com.velocitypowered.api.command;
+
+import com.google.common.base.Preconditions;
+import com.mojang.brigadier.Message;
+import net.kyori.adventure.text.Component;
+import net.kyori.adventure.text.serializer.legacy.LegacyComponentSerializer;
+
+/**
+ * Represents an implementation of brigadier's {@link Message}, providing support for {@link
+ * Component} messages.
+ */
+public final class VelocityBrigadierMessage implements Message {
+
+  private final Component message;
+
+  public VelocityBrigadierMessage(Component message) {
+    this.message = Preconditions.checkNotNull(message, "message");
+  }
+
+  /**
+   * Returns the message as a {@link Component}.
+   *
+   * @return message as component
+   */
+  public Component asComponent() {
+    return message;
+  }
+
+  /**
+   * Returns the message as a legacy text.
+   *
+   * @return message as legacy text
+   */
+  @Override
+  public String getString() {
+    return LegacyComponentSerializer.legacySection().serialize(message);
+  }
+}
--- a/api/src/main/java/com/velocitypowered/api/command/VelocityBrigadierMessage.java
+++ b/api/src/main/java/com/velocitypowered/api/command/VelocityBrigadierMessage.java
@@ -26,7 +26,7 @@ public final class VelocityBrigadierMessage implements Message, ComponentLike {
 
   private final Component message;
 
-  public VelocityBrigadierMessage(Component message) {
+  private VelocityBrigadierMessage(Component message) {
     this.message = Preconditions.checkNotNull(message, "message");
   }
 
--- a/api/src/main/java/com/velocitypowered/api/command/VelocityBrigadierMessage.java
+++ b/api/src/main/java/com/velocitypowered/api/command/VelocityBrigadierMessage.java
@@ -20,6 +20,10 @@ import org.jetbrains.annotations.NotNull;
  */
 public final class VelocityBrigadierMessage implements Message, ComponentLike {
 
+  public static VelocityBrigadierMessage tooltip(Component message) {
+    return new VelocityBrigadierMessage(message);
+  }
+
   private final Component message;
 
   public VelocityBrigadierMessage(Component message) {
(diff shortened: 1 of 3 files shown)


api/src/main/java/com/velocitypowered/api/command/VelocityBrigadierMessage.java
```diff
@@ -0,0 +1,45 @@
...
+  public VelocityBrigadierMessage(Component message) {
+    this.message = Preconditions.checkNotNull(message, "message");
+  }
+
+  /**
+   * Returns the message as a {@link Component}.
+   *
+   * @return message as component
+   */
+  public Component asComponent() {
```
hugmanrique, line 38: Implement `ComponentLike`

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientPlaySessionHandler.java
```diff
@@ -497,15 +499,19 @@ private boolean handleCommandTabComplete(TabCompleteRequest packet) {
...
+    server.getCommandManager().offerBrigadierSuggestions(player, command)
         .thenAcceptAsync(suggestions -> {
           if (suggestions.isEmpty()) {
             return;
           }
 
           List<Offer> offers = new ArrayList<>();
-          for (String offer : suggestions) {
-            offers.add(new Offer(offer));
+          for (Suggestion offer : suggestions.getList()) {
```
hugmanrique: Nit: a suggestion is not an offer.

hugmanrique (changes requested):
Nice catch! I didn't even know about these tooltips. Left two minor nits.

api/src/main/java/com/velocitypowered/api/command/VelocityBrigadierMessage.java
```diff
@@ -0,0 +1,45 @@
...
+    return message;
+  }
+
+  /**
+   * Returns the message as a legacy text.
+   *
+   * @return message as legacy text
+   */
+  @Override
+  public String getString() {
```
hugmanrique, line 48: Is this used somewhere? Users should convert components to text themselves, as there's multiple representations.

MrIvanPlays: This is not used by velocity's code. Could be used by brigadier's code (in that case /shrug). That's the way it came to my mind to implement this method, rather for it to stay null.

hugmanrique: Ah, my bad, didn't notice the `@Override`. I would use `PlainTextComponentSerializer` instead, since the messages are shown to players (see `LiteralMessage`)

hugmanrique:
Sorry, I didn't explain myself: I meant that `VelocityBrigadierMessage` should implement `ComponentLike` since you already wrote the `asComponent` method yourself.

MrIvanPlays:
> Sorry, I didn't explain myself: I meant that `VelocityBrigadierMessage` should implement `ComponentLike` since you already wrote the `asComponent` method yourself.

Why should it if it is going to be used only in tooltips?

hugmanrique:
Because a class that wraps a `Component` (and nothing else) is a `ComponentLike` in most cases.

Also, the comment was marked as completed, but I would use PlainTextComponentSerializer instead, since the messages are shown to players (see LiteralMessage)

MrIvanPlays:
Is it good now?

MrIvanPlays:
What about changing this https://github.com/VelocityPowered/Velocity/blob/dev/3.0.0/proxy/src/main/java/com/velocitypowered/proxy/command/VelocityCommandManager.java#L216 to send the error message more like the client does? something like what I do here: https://github.com/MrIvanPlays/IvanCord/blob/master/Waterfall-Proxy-Patches/0026-Brigadier-command-api.patch#L1438-L1501

hugmanrique (approved):

hugmanrique:
I'm not too sure the amount of introduced code would justify the minimal change. Anyways, that should go in a different PR. We can further discuss this on Discord.

Again, thank you for the contribution.

MrIvanPlays:
> We can further discuss this on Discord.

Send me a friend request: MrIvanPlays#2830 ~~(ill tell you why after we get in PMs so just do it)~~

api/src/main/java/com/velocitypowered/api/command/VelocityBrigadierMessage.java
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Represents an implementation of brigadier's {@link Message}, providing support for {@link
+ * Component} messages.
+ */
+public final class VelocityBrigadierMessage implements Message, ComponentLike {
+
+  private final Component message;
+
+  public VelocityBrigadierMessage(Component message) {
```
astei: Please expose this constructor as a `static` method, perhaps `VelocityBrigadierMessage.tooltip()`?

astei (changes requested):

api/src/main/java/com/velocitypowered/api/command/VelocityBrigadierMessage.java
astei: Also make the constructor proper `private`.

astei (changes requested):

astei (approved):

PR #590, opened 2021-11-03
Expose world name

### LINKED ISSUE #521: World Name by Player Support
Hello, I wrote with Five and he told me it is doable to add a method for getting the name of the current world which the player is in. (Only 1.16 or above) 
I would really appreciate if you can add this. Then I don't need a plugin for my backend servers.

MrIvanPlays:
I decided this method is best suited in ServerConnection so I put it there. 

I have a few questions though:

- Should it be `Optional<Key>` , or just return a Key and mark it `@Nullable`?
- Should it have a warning that this works on 1.16+? 

Fixes #521

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/proxy/ServerConnection.java
+++ b/api/src/main/java/com/velocitypowered/api/proxy/ServerConnection.java
@@ -11,6 +11,8 @@ import com.velocitypowered.api.proxy.messages.ChannelMessageSink;
 import com.velocitypowered.api.proxy.messages.ChannelMessageSource;
 import com.velocitypowered.api.proxy.server.RegisteredServer;
 import com.velocitypowered.api.proxy.server.ServerInfo;
+import java.util.Optional;
+import net.kyori.adventure.key.Key;
 
 /**
  * Represents a connection to a backend server from the proxy for a client.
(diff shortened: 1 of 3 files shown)


api/src/main/java/com/velocitypowered/api/proxy/ServerConnection.java
```diff
@@ -37,4 +39,12 @@
...
    */
   Player getPlayer();
+
+  /**
+   * Returns the world the current connection's associated {@link Player} is playing on.
+   *
+   * @return an {@link Optional} the name of the world that the player is playing on, which may be
+   *     empty.
+   */
+  Optional<Key> getCurrentWorldName();
```
hugmanrique, line 49: I would also document the optional is empty if the client's version is <1.16 (I'm not too sure that's the right version, going with what the linked issue says).

MrIvanPlays: that's what I questioned; whether to document that it will be an empty optional if the version is < 1.16

Xernium:
Here are the things wrong with this:
1. It changes upon Respawn
2. It’s not guaranteed to be a key
3. It doesn’t tell you anything about the dimension type, only the world “name”  If this gets added it needs to be wrapped in serious API and not like this

MrIvanPlays:
> It doesn’t tell you anything about the dimension type, only the world “name” If this gets added it needs to be wrapped in serious API and not like this

What should such API include/expose?

Xernium:
> > It doesn’t tell you anything about the dimension type, only the world “name” If this gets added it needs to be wrapped in serious API and not like this
> 
> What should such API include/expose?

The entire registry Dimension info package ideally, but that’s not the best of ideas. The data won’t be accurate over 90% of the time anyway because APIs like Forge Fabric or Spigot don’t actually correctly use this system 
So to put my personal opinion on the table, I’m against this feature, but I will not stand in the way if another project member is for it.

MrIvanPlays:
I'm not going to try and make such an API given the fact the data won't be accurate most of the times.

PR #591, opened 2021-11-08
Do not track plugin channels registered per-player on the proxy

astei:
We don't need to track this information since Velocity uses the `JoinGame` packet, which is about as good of a server rejoin mechanism we're likely to get in vanilla Minecraft.

This change should not impact modding frameworks, at least [Fabric API for Minecraft 1.17.1 should work](https://github.com/FabricMC/fabric/blob/daa367217679ffae237a7216154255bd840edc47/fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/client/ClientPlayNetworkAddon.java#L66) but compatibility with older clients and other modding frameworks must be ensured.


AoElite:
@astei Why was this not merged? Would probably fix https://github.com/PaperMC/Velocity/issues/819

astei:
I haven't had time to verify this works all the way back to 1.7.

kashike (approved):

PR #609, opened 2021-12-10
Add version information for 1.18.1

kosma:
I am not entirely sure if this is everything that has to be changed - please verify. According to Wiki the protocol version has not changed: https://minecraft.fandom.com/wiki/Protocol_version#Java_Edition_2 so this is a rather cosmetic change.


astei:
You’re correct, the protocol has not changed one iota.

Issue #624, opened 2021-12-30
Add Event for Resource Pack Send Packets


CoreyShupe:
This issue is representing the work for adding in an event for the [Resource Pack Send](https://wiki.vg/Protocol#Resource_Pack_Send) packet.

This allows the owner of a proxy to transform the resource pack request coming from the backend servers at the proxy level rather than sending a resource pack request at the proxy level. This can help synchronize server resource pack sends without needing to push the logic into the proxy level.

astei:
This is long-implemented, so closing.

PR #625, opened 2021-12-30
Implement resource pack send event.

CoreyShupe:
This resolves issue #624


api/src/main/java/com/velocitypowered/api/event/player/PlayerResourcePackSendEvent.java
```diff
@@ -0,0 +1,97 @@
...
+import net.kyori.adventure.text.Component;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+/**
+ * This event is fired when the downstream server tries to send a player a ResourcePackSend packet.
+ * Velocity will wait on this event to finish before forwarding the packet to the user. If this
+ * event is denied, it will retroactively send a DENIED status to the downstream server in response.
+ * If the downstream server has it set to "forced" it will forcefully disconnect the user.
+ */
+public class PlayerResourcePackSendEvent implements ResultedEvent<ResultedEvent.GenericResult> {
```
CoreyShupe: once we move to java 16+ if we do, we might want to just turn this into a record, it doesn't need to be a class realistically.

astei (approved):

Issue #566, opened 2021-08-24
Backend Servers Rearchitecture


CoreyShupe:
Right now there's a very strict way of viewing servers as only a "registered server" which is tied to the server map (which is already a broken contract).

This server only contains a `server-info` with the IP, port, and information regarding the server itself, while the registered server contains things like players, and the ability to ping the remote.

This is an issue since there's a lot of things which go on in a server which there's either no control over or it's statically defined which takes away from the idea of a "server", such as forwarding method, forwarding secrets, etc... which all go along with the server's handshake.

The goal is to properly restructure servers to reflect that some servers won't be registered, along with allowing servers to maintain their own configuration while using the main velocity configuration as a basic default.

The goal:
```
server (as an audience) => {
    remote info,
    handshake data? => {
        forwarding method
        forwarding secret?
    }
    name
}
```
This server type should be able to provide the basic defaults of what a server actually is. The registration adds a layer on it, but it should not actually be required as something for a velocity player to connect to.

Players should also be able to connect to the raw server rather than simply a "VelocityRegisteredServer" which it's locked into by virtue of how it's written. This can be avoided by allowing this interface to be accepted by the player as a valid connection point. If a user decides to misconfigure or send a player an improperly made server it should be on them to understand the use case outside of the immediate API (creating their own implementation of a "server").

```
Velocity Registered Server => {
  <connected players>
  <server instance>
  <server interface>
}
```

Registered servers should comprise solely of registered servers, and the sub-server implementation which those players are connected to.

The server map should work as a default but should not be heavily enforced into the idea of what a server is. There are plenty of use-cases for a system where server registration is a secondary use-case to a preferred stateless setup. Although this idea was implemented earlier it still does not fit well with what it is actually doing as in bad naming and improper handling of how the server is actually dealt with.

With this new setup servers can have memoized configurations with customized forwarding secrets and forwarding modes. This specific use-case looks at a server where it needs:

```
server a = 1.12.2 forge
server b = 1.17.1 fabric
```

This setup is impossible with the current velocity layout but becomes possible when player info forwarding is dynamic.

Another use-case for this kind of structure is such:
```
server a owned by person a = modern
server b owned by person b = modern

server a should not be accessible via person b
```
[...]

PR #626, opened 2022-01-02
Make RegisteredServer implementations stateless

roccodev:
This is a first attempt at solving #566.

This shifts the stateful logic behind `VelocityRegisteredServer` (keeping track of connected players and sending plugin messages to those players) to a new class (`VelocityServerState`, backed by the `RegisteredServerState` API).

This way one could make a custom `RegisteredServer` that doesn't have to be registered to the proxy, perhaps forwarding the stateful logic to a server that's actually registered. (server aliasing)  
For instance, you can now write this:

```java
public class AliasedServer implements RegisteredServer {
    private final RegisteredServer handle;
    private final ServerInfo serverInfo;

    public AliasedServer(RegisteredServer handle, String alias) {
        this.handle = handle;
        this.serverInfo = new ServerInfo(alias, handle.getServerInfo().getAddress());
    }

    @Override
    public ServerInfo getServerInfo() {
        return serverInfo;
    }

    @Override
    public CompletableFuture<ServerPing> ping() {
        return handle.ping();
    }

    @Override
    public RegisteredServerState getState() {
        return handle.getState();
    }
}

```

This is a breaking API change, but backwards compatibility is kept through deprecated methods.

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/proxy/server/MutableServerState.java
+++ b/api/src/main/java/com/velocitypowered/api/proxy/server/MutableServerState.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright (C) 2018 Velocity Contributors
+ *
+ * The Velocity API is licensed under the terms of the MIT License. For more details,
+ * reference the LICENSE file in the api top-level directory.
+ */
+
+package com.velocitypowered.api.proxy.server;
+
+import com.velocitypowered.api.proxy.Player;
+
+/**
+ * A mutable view on a server's state.
+ *
+ * <p>This interface's methods may be called from different threads. As such, <strong>it is important to consider
+ * thread-safety in implementations.</strong>
+ */
+public interface MutableServerState extends ServerState {
+  /**
+   * Called when a player has switched to the {@link RegisteredServer} associated with this state.
+   *
+   * @param player the player
+   */
+  void addPlayer(Player player);
+
+  /**
+   * Called when a player has disconnected from {@link RegisteredServer} associated with this state.
+   *
+   * @param player the player
+   */
+  void removePlayer(Player player);
+}
--- a/api/src/main/java/com/velocitypowered/api/proxy/server/RegisteredServer.java
+++ b/api/src/main/java/com/velocitypowered/api/proxy/server/RegisteredServer.java
@@ -8,15 +8,18 @@
 package com.velocitypowered.api.proxy.server;
 
 import com.velocitypowered.api.proxy.Player;
+import com.velocitypowered.api.proxy.messages.ChannelIdentifier;
 import com.velocitypowered.api.proxy.messages.ChannelMessageSink;
 import java.util.Collection;
 import java.util.concurrent.CompletableFuture;
 import net.kyori.adventure.audience.Audience;
 
 /**
- * Represents a server that has been registered with the proxy. The {@code Audience} associated with
- * a {@code RegisteredServer} represent all players on the server connected to this proxy and do not
- * interact with the server in any way.
+ * Represents a server that has been registered with the proxy.
+ *
+ * <p>This is an immutable and stateless view over a server's information. To get the server's current state
+ * (such as players connected), use {@link RegisteredServer#getState()}.
+ * </p>
  */
 public interface RegisteredServer extends ChannelMessageSink, Audience {
 
@@ -31,8 +34,12 @@ public interface RegisteredServer extends ChannelMessageSink, Audience {
    * Returns a list of all the players currently connected to this server on this proxy.
    *
    * @return the players on this proxy
+   * @deprecated use {@link ServerState#getPlayersConnected()} instead
    */
-  Collection<Player> getPlayersConnected();
+  @Deprecated
+  default Collection<Player> getPlayersConnected() {
+    return getState().getPlayersConnected();
+  }
 
   /**
    * Attempts to ping the remote server and return the server list ping result.
@@ -40,4 +47,25 @@ public interface RegisteredServer extends ChannelMessageSink, Audience {
    * @return the server ping result from the server
    */
   CompletableFuture<ServerPing> ping();
+
+  /**
+   * Returns the current state of this server on the proxy.
+   *
+   * @return the server state
+   */
+  ServerState getState();
+
+  /**
+   * Sends a plugin message to all the players connected this server.
+   *
+   * @param identifier the channel identifier to send the message on
+   * @param data the data to send
+   * @return whether or not the message could be sent
+   * @deprecated use {@link ServerState#sendPluginMessage(ChannelIdentifier, byte[])} instead
+   */
+  @Override
+  @Deprecated
+  default boolean sendPluginMessage(ChannelIdentifier identifier, byte[] data) {
+    return getState().sendPluginMessage(identifier, data);
+  }
 }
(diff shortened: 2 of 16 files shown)


api/src/main/java/com/velocitypowered/api/proxy/server/RegisteredServerState.java
```diff
@@ -0,0 +1,30 @@
...
+ * connected to this proxy and do not interact with the server in any way.
+ * </p>
+ */
+public interface RegisteredServerState extends ChannelMessageSink, Audience {
+  /**
+   * Returns a list of all the players currently connected to this server on this proxy.
+   *
+   * @return the players on the proxy connected to this server
+   */
+  Collection<? extends Player> getPlayersConnected();
```
astei: Not sure how I feel about this change. While this is going to be binary-compatible due to erasure, it does add some friction for external users.

proxy/src/main/java/com/velocitypowered/proxy/connection/backend/BungeeCordMessageResponder.java
```diff
@@ -172,7 +172,7 @@ private void processMessageRaw(ByteBufDataInput in) {
   }
 
   private void processMessage0(ByteBufDataInput in,
-      ComponentSerializer<Component, ?, String> serializer) {
+                               ComponentSerializer<Component, ?, String> serializer) {
```
astei: Lots of unnecessary reformatting. Please keep your PR focused on the feature at hand.

proxy/src/main/java/com/velocitypowered/proxy/server/VelocityServerState.java
```diff
@@ -0,0 +1,85 @@
...
+import io.netty.buffer.Unpooled;
+import java.util.Collection;
+import java.util.Map;
+import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
+import net.kyori.adventure.audience.Audience;
+import net.kyori.adventure.audience.ForwardingAudience;
+import org.checkerframework.checker.nullness.qual.NonNull;
+
+public class VelocityServerState implements MutableServerState, ForwardingAudience {
```
astei: Nit: I'm not such a fan of `VelocityRegisteredServer` and `VelocityServerState` nowadays. If I were naming these classes today, I would go with `RegisteredServerImpl` (the `RegisteredServer` interface is effectively a side-show at this point) and `DefaultMutableServerState`.

api/src/main/java/com/velocitypowered/api/proxy/server/MutableServerState.java
```diff
@@ -0,0 +1,29 @@
...
+ * The Velocity API is licensed under the terms of the MIT License. For more details,
+ * reference the LICENSE file in the api top-level directory.
+ */
+
+package com.velocitypowered.api.proxy.server;
+
+import com.velocitypowered.api.proxy.Player;
+
+/**
+ * A mutable view on a server's state.
```
astei, line 13: This needs to also specify that said view should be thread-safe.

api/src/main/java/com/velocitypowered/api/proxy/server/MutableServerState.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+public interface MutableServerState extends RegisteredServerState {
+  /**
+   * Called when a player should be added to the server.
```
astei:
```suggestion
   * Called when a player has switched to the {@link RegisteredServer} associated with this state.
```

api/src/main/java/com/velocitypowered/api/proxy/server/MutableServerState.java
```diff
@@ -0,0 +1,29 @@
...
+   *
+   * @param player the player
+   */
+  void addPlayer(Player player);
+
+  /**
+   * Called when a player should be removed from the server.
```
astei:
```suggestion
   * Called when a player has disconnected from {@link RegisteredServer} associated with this state.
```

astei (changes requested):
Some initial thoughts on this. I like the direction this is going.

api/src/main/java/com/velocitypowered/api/proxy/server/RegisteredServer.java
```diff
@@ -31,13 +34,49 @@
...
+  /**
+   * Returns whether the server is registered in the proxy's server list. (i.e., in the config)
+   *
+   * <p>If this is {@code false}, then {@link com.velocitypowered.api.proxy.ProxyServer#getServer(String)}
+   * might return an empty {@link java.util.Optional}.
+   * </p>
+   *
+   * @return whether the server is in the config
+   */
+  boolean isRegistered();
```
hugmanrique: This seems like an implementation detail. A `RegisteredServer` surely is registered. We could add a note to `ProxyServer#getServer` documenting that it may return an empty optional depending on the registration means.

api/src/main/java/com/velocitypowered/api/proxy/server/RegisteredServerState.java
```diff
@@ -0,0 +1,30 @@
...
+import com.velocitypowered.api.proxy.messages.ChannelMessageSink;
+import java.util.Collection;
+import net.kyori.adventure.audience.Audience;
+
+/**
+ * A server's current state. This differs from {@link RegisteredServer} for providing a stateful view
+ * on a backend server instance.
+ *
+ * <p>The {@code Audience} associated with a {@code RegisteredServerState} represent all players on the server
+ * connected to this proxy and do not interact with the server in any way.
```
hugmanrique, line 20: What's the meaning of "do not interact with the server in any way"?

api/src/main/java/com/velocitypowered/api/proxy/server/RegisteredServerState.java
```diff
@@ -0,0 +1,30 @@
...
+ * </p>
+ */
+public interface RegisteredServerState extends ChannelMessageSink, Audience {
```
hugmanrique: Perhaps a tad controversial, but I would call this `ServerState`. This decouples the notion of registration from the actual purpose of the class: contain the state of a server. A comment above hints at the fact that we are conflating two different concepts of "registration": the server's parameters like e.g. IP:port pair being known by Velocity; and these parameters appearing in `velocity.toml`. I wish we could go further and rename `RegisteredServer` (since Velocity should be the only one creating objects of this type, a `RegisteredServer`/`Server` is registered by nature), but that's not an option due to API compat.

hugmanrique (commented):
Left some remarks on API design. Implementation LGTM.

api/src/main/java/com/velocitypowered/api/proxy/server/RegisteredServer.java
roccodev: True, though ultimately we'd want to rename the class. The main premise of the change is to allow servers to not be "registered" entirely, and just work as abstractions over a server connection.

hugmanrique (approved):
API side now LGTM :)

XLordalX:
What's the status on this? This feature is pretty critical for servers utilizing something like Kubernetes where ip addresses are constantly changing.

CoreyShupe:
> What's the status on this? This feature is pretty critical for servers utilizing something like Kubernetes where ip addresses are constantly changing.

I think it's being taken a different way in design - I'm going to write up a UML + draft PR for the overhaul this weekend, see the initial issue in the post, it should be able to be done without any breaking changes at all.

PR #628, opened 2022-01-08
Limit chat message size

roccodev:
This allows server owners to block oversized messages on the proxy's end.  
They can either choose to kick the player or to just cut the message to fit within limits.


tjalp:
I don't think this should be something built-in to Velocity, but rather something a plugin should do

roccodev:
That's also something I was wondering, but Velocity already checks for invalid characters, and it also rejects oversized messages in the API's `Player#spoofChatInput(String)`.

4drian3d:
> I don't think this should be something built-in to Velocity, but rather something a plugin should do

I agree, this pr would limit the ability of modded servers to receive chat messages of any length they require

PR #631, opened 2022-01-21
Add missing check in MinecraftVarintFrameDecoder

ghost:
People can exploit this missing check by sending a too short VarInt

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/protocol/netty/MinecraftVarintFrameDecoder.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/protocol/netty/MinecraftVarintFrameDecoder.java
@@ -30,6 +30,8 @@ public class MinecraftVarintFrameDecoder extends ByteToMessageDecoder {
       new QuietDecoderException("Bad packet length");
   private static final QuietDecoderException VARINT_BIG_CACHED =
       new QuietDecoderException("VarInt too big");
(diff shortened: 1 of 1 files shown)


Xernium:
The state of too-short simply means that the reader hasn‘t yet read a valid VarInt.
Furthermore, if you were to add this change then you‘re potentially breaking the decoder as a whole.
There is no security issue here, a VarInt is either too long or invalid.

ghost:
> The state of too-short simply means that the reader hasn‘t yet read a valid VarInt. Furthermore, if you were to add this change then you‘re potentially breaking the decoder as a whole. There is no security issue here, a VarInt is either too long or invalid.

I already tested this change in a production server with 70 players

KoxSosen:
Could you please send an example of how the MinecraftVarintFrameDecoder can be exploited without this "check"? I'm just curious.

electronicboy:
This change is the type of change that given the data involved will likely work fine, until one day it doesn't with 0 warning, networking gets fun that way.

This is not really the place for sharing exploits, but, I somewhat see a way in which this could be exploited, but, I don't think that introducing a behavioral bug in the quest of solving a DoS exploit is a good fix here.

Shevchik:
This isn't a missing check. That's how TCP works, it's quite possible to receive 1 byte per frame, that's why you add up the read data and discard only after successful read (Literally what ByteToMessageDecoder does).

PR #637, opened 2022-02-03
Disable Log4J's built-in shutdown hook which broke our shutdown sequence.

kyngs:
When exiting by pressing CTRL + C, log4j shuts down before we do. Which causes any calls to the logger in our shutdown sequence to throw an exception. This resulted in Velocity (and running plugins) not shutting down gracefully.

I have disabled the built-in hook and added a line to shutdown Log4J ourselves. You might consider moving it to earlier parts of the shutdown sequence.


kyngs:
Any update on this? I am still suffering from this issue and I would love to stop using my fork in production.

astei (approved):

PR #643, opened 2022-02-11
Add passthrough-default-server-ping option

BBaoVanC:
If `passthrough-default-server-ping` is set to false, then Velocity will respond with its own MOTD and version for pings that go to a host not listed in `forced-hosts`. This means that you can make it so Velocity has its own generic network-wide MOTD which is shown on the proxy's main IP(s), but also see the backend's MOTD and version if connecting to it specifically (using a forced host).

This option doesn't have any effect if `ping-passthrough` is disabled.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
@@ -78,6 +78,8 @@ public class VelocityConfiguration implements ProxyConfig {
   private boolean onlineModeKickExistingPlayers = false;
   @Expose
   private PingPassthroughMode pingPassthrough = PingPassthroughMode.DISABLED;
+  @Expose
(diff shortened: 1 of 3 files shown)


BBaoVanC:
Rebased because of the addition of `enablePlayerAddressLogging`, I think this is ready for review

JT122406:
Can conflicts be resolved and then this merged this would be a great feature

BBaoVanC:
Would be nice to know if this has a chance at being merged directly to Velocity (I hope it can) first, before I spend time rebasing it but yes

Xernium:
I’m putting this, and any other PR that changes something in the config, on hold. We absolutely need a config rewrite before this gets merged. Changes like this to the old system are simply unsustainable now

JT122406:
> I’m putting this, and any other PR that changes something in the config, on hold. We absolutely need a config rewrite before this gets merged. Changes like this to the old system are simply unsustainable now

Any idea when this rewrite will happen for the config

PR #647, opened 2022-02-14
Add support for custom class loaders.

kyngs:
**This PR does not contain changes made by myself in #637**

If Velocity was loaded by a different class loader (for example, URLClassLoader), when attempting to load plugins, it would throw an exception, saying, that the plugin that is being loaded cannot find Velocity API classes. I've investigated a bit and found out, that this is caused by a class loader issue. Adding the class loader which was used to load Velocity, as a parent to our class loader, seems to fix the issue.

### How I encountered this

I am doing a plugin updater which unlike most, doesn't start the server by just running the java -jar... command. But, it loads the server jar classes and invokes them. To do that I need to use a custom URLClassLoader which raised the issues described above.

### Why should this get merged
It is kinda obvious, that this is an incredibly specific scenario, however, I can imagine it would help some other developers doing a similar thing. Also, I didn't find this breaking something or introducing some new issues, also the performance should not be affected at all.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
@@ -502,6 +502,9 @@ public class VelocityServer implements ProxyServer, ForwardingAudience {
         Thread.currentThread().interrupt();
       }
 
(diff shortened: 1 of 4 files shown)


astei (approved):
I'm not entirely convinced by the value of this patch, but given that it's a relatively minor change, I'll let that slide. Approving.

Issue #614, opened 2021-12-17
Option to disable logging of ip addresses


TheLeCrafter:
When using Paper there is a setting in the paper.yml to disable logging of player ip addresses. The ip addresses will be replaced with "\<ip address withheld\>" in the log. Add an option to the velocity.toml to disable the logging of ip addresses in the same way.

kyngs:
I believe this issue may be closed since it was resolved by #650

PR #650, opened 2022-02-20
Add option to disable player ip logging (#614)

ebony-lamp:
Added `enable-player-address-logging` to velocity config. If disabled (default is true) player IP addresses will be replaced by <ip address withheld> in logs.

Lmk if there was anything I missed as I am new to open source :). Thanks!

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
@@ -73,6 +73,7 @@ public class VelocityConfiguration implements ProxyConfig {
   @Expose private final Advanced advanced;
   @Expose private final Query query;
   private final Metrics metrics;
+  @Expose private boolean enablePlayerAddressLogging = true;
   private net.kyori.adventure.text.@MonotonicNonNull Component motdAsComponent;
   private @Nullable Favicon favicon;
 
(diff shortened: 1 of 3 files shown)


proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
```diff
@@ -88,8 +89,9 @@ private VelocityConfiguration(Servers servers, ForcedHosts forcedHosts, Advanced
   private VelocityConfiguration(String bind, String motd, int showMaxPlayers, boolean onlineMode,
       boolean preventClientProxyConnections, boolean announceForge,
       PlayerInfoForwarding playerInfoForwardingMode, byte[] forwardingSecret,
-      boolean onlineModeKickExistingPlayers, PingPassthroughMode pingPassthrough, Servers servers,
-      ForcedHosts forcedHosts, Advanced advanced, Query query, Metrics metrics) {
+      boolean onlineModeKickExistingPlayers, PingPassthroughMode pingPassthrough,
+      boolean disablePlayerIpLogging, Servers servers,ForcedHosts forcedHosts,
```
ebony-lamp: After giving this a second look, I feel like I shouldn't be changing this constructor. Not sure since I am new to this codebase. Feedback on this would be appreciated :)

hugmanrique (commented):
Nit: instead of the double negative, rename the setting to `enable-player-address-logging` and set it to `true` by default.

The addition of a setting for this has been questioned in the past, since the result can also be achieved with a slf4j filter. I'm in favor of this PR since the filter 1. is difficult to configure for regular server admins, and 2. can affect performance.

proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
hugmanrique: This is fine 👍

PR #657, opened 2022-03-01
Fix forced-hosts

### LINKED ISSUE #653: Forced hosts are ignored
When connecting to the server, velocity always forwards the connection to the `lobby` server. The config:
```toml
[servers]
  lobby = "172.18.0.1:25565"
  prvt = "172.18.0.1:3540"
  prvt1 = "172.18.0.1:3541"

[forced-hosts]
  "prvt.mc.example.com" = "prvt"
  "medieval.mc.example.com" = "prvt1"
  ```

### PR DESCRIPTION
**Pull request by Kebab11noel**
Fix #653 by splitting the host string at `/` and taking only the first sub-string

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
@@ -726,6 +726,7 @@ public class ConnectedPlayer implements MinecraftConnectionAssociation, Player {
     if (serversToTry == null) {
       String virtualHostStr = getVirtualHost().map(InetSocketAddress::getHostString)
           .orElse("")
+          .split("/", 2)[0]
           .toLowerCase(Locale.ROOT);
       serversToTry = server.getConfiguration().getForcedHosts().getOrDefault(virtualHostStr,
           Collections.emptyList());

### CODE REVIEW

**Comment by electronicboy** [comment]
have you tested this, as this makes 0 sense to me?

**Comment by kyngs** [comment]
> have you tested this, as this makes 0 sense to me?

I also think that this should be tested. However, this makes sense to me if it was the issue. My guess is, that InetSocketAddress::getHostString **could** return something like mc.example.com/ instead of mc.example.com, which would cause the issue.

**Comment by electronicboy** [comment]
it can't

image

**Comment by kyngs** [comment]
I still believe that this is worth testing.

**Comment by electronicboy** [comment]
This makes 0 sense when you look at the logic, /unresolved only ever shows up when toString is called, which is not the case here; I don't have the setup to test this, but, I fail to see how this would show up as /unresolved given that it's reading directly from the hostname field which is passed in; if unresolved IS getting in there, this would be an issue elsewhere

**Comment by Kebab11noel** [comment]
Yes, I did test it and I use that build now. The host string could look like `domain/<unresolved>:XXXXX` or even `domain///XXX.XXX.XXX.XXX:XXXXX///XXXXXXXXXX///long-base64-encoded-binary-data` (`X` represents a single digit, domain is well obviously the domain name used to connect)

**Comment by Kebab11noel** [comment]
Yes, I did test it and I use that build now. The host string could look like `domain/<unresolved>:XXXXX` or even `domain///XXX.XXX.XXX.XXX:XXXXX///XXXXXXXXXX///long-base64-encoded-binary-data` (`X` represents a single digit, domain is well obviously the domain name used to connect)

**Comment by electronicboy** [comment]
what's the host string in the handshake packet itself? base64 data in there sounds suspicious, non-vanilla client?

**Comment by kyngs** [comment]
> Yes, I did test it and I use that build now. The host string could look like `domain/<unresolved>:XXXXX` or even `domain///XXX.XXX.XXX.XXX:XXXXX///XXXXXXXXXX///long-base64-encoded-binary-data` (`X` represents a single digit, domain is well obviously the domain name used to connect)

I was unable to reproduce this on my **local** environment with both vanilla and fabric clients (1.18&1.18.1), the hostname which I obtain the same way as Velocity is: `localhost` with no slash or something other.
However, I find this quote from InetSocketAddress#getHostString suspicious.
```
Returns the hostname, or the String form of the address if it doesn't have a hostname (it was created using a literal). This has the benefit of not attempting a reverse lookup.
```

Maybe in some cases, it defaults to the IP of the connection which could cause the issue?

**Comment by kyngs** [comment]
When I look at this, it looks like velocity doesn't use the address provided in the handshake packet, maybe it should have a priority? And if it wasn't provided, just fallback to this?

EDIT: I've done some searching and it appears that the InetSockedAddress is created from the information provided in the handshake packet.

**Comment by kyngs** [comment]
I confirm, that we (me and @Kebab11noel) were able to reproduce this on his test proxy.
The code on the proxy was the following:
```java
@Subscribe
public void handler(PreLoginEvent e) {
    InboundConnection conn = e.getConnection();
    String vhost = conn.getVirtualHost().map(InetSocketAddress::getHostString).orElse("").toLowerCase(Locale.ROOT);
    if (vhost.contains("mct")) {
        e.setResult(PreLoginEvent.PreLoginComponentResult.denied(Component.text(vhost)));
    }
}
```
When I connected, the proxy kicked me (as expected) with the host string consisting of:
image
(If you are having trouble reading the image, open it up in a new tab)

The proxy is build 115
My client is vanilla 1.18

I was connecting to: test-a.mct.noelnemeth.com

The thing at the end which looks like base64 (it is not to our knowledge), changes every time I attempt to join.
I would approve this PR (I realize that this is not up to me)

**EDIT: It seems that this is caused by TCPShield**

Issue #654, opened 2022-02-28
Bukkit PlayerResourcePackStatusEvent doesn't work on Velocity


DaJokni:
A "recent" commit seems to have broken the "PlayerResourcePackStatusEvent" in the bukkit api, it just doesnt fire ever. I'd assume that the packets related to accepting the resource packs arent being sent back to the backend.

The pack gets sent just fine to the client, but the backend server doesnt seem to get that info.

The same thing works just fine on the build (build 110) before that commit.

The commit in question: https://github.com/PaperMC/Velocity/commit/596d4758ba9efedc60ede2789644ea10263c48e4

EDIT: And no, errors don't show up, not in the backend or velocity.

codeHusky:
This is causing a critical issue with our custom gamemode server - no errors from what I can see either and we're on the latest dev build available tonight (build 115). Important this event works so we can kick players who don't accept the resource pack.

DeprecatedLuke:
Confirmed, having the same issue.

ytnoos:
Same issue

CoreyShupe:
I'll test this out and work on a fix on Saturday. If anyone finds anything before then feel free to PR and ref the issue.

If anyone wants to give it a crack, I think the issue lies in the changes to the `Origin` stuff for the resource packs. That's where I'd recommend looking at first for the control flow of status sends.

electronicboy:
Just to confirm, the pack is being sent from the backend server and not the proxy?

DaJokni:
> Just to confirm, the pack is being sent from the backend server and not the proxy?

That is indeed the case, with Player#setResourcePack

CoreyShupe:
https://github.com/PaperMC/Velocity/pull/659 Can you see if the issue still resides in this PR.

codeHusky:
can confirm the issue is fixed with #659, posted picture there

CoreyShupe:
Note: This should be closed, #659 merged

OverBrave:
I am still having this problem. Can you check again?

Velocity 3.1.2-SNAPSHOT (git-74edac96-b153)

PR #659, opened 2022-03-02
Flip output for queued resource pack status passthrough conditional

CoreyShupe:
This should solve #654 

It looks like the conditional should be:
* false if we want the packet to passthrough to the downstream server
* true if velocity sent the packet

The current implementation has it `true` for when the downstream server sent the resource pack, and `false` for when the proxy sent the resource pack.


electronicboy (approved):

codeHusky:
image
Can confirm this fixes it

codeHusky:
image
Can confirm this fixes the issue, should be merged in if it tests good otherwise

PR #661, opened 2022-03-06
Set player list in ping response

yawek9:
Adds writing online players list to the ping status response.


astei:
...why not use a plugin?

electronicboy:
imho, no; This would at the very least need to implement the recent "privacy" settings added in vanilla

4drian3d:
superseded by #811

PR #662, opened 2022-03-12
Replace opt-out dark pattern with opt-in.

tomlister:
This PR removes the "opt-out" dark pattern and replaces it with privacy first "opt-in" to ensure that users consent to having telemetry sent.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/Metrics.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/Metrics.java
@@ -74,7 +74,7 @@ public class Metrics {
       logger.info("bStats collects some basic information for plugin"
           + " authors, like how many people use");
       logger.info("their plugin and their total player count.");
-      logger.info("If you'd like to enable this, please opt-in"
+      logger.info("If you're comfortable with this, please opt-in"
           + " by editing the config.txt file in");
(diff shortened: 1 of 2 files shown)


tomlister:
Any update on this?

astei:
Closing this, and while this will sound nihilistic, sadly that is how the world works right now.

I'm closing this for two main reasons:

1. The most practical reason is because this deviates from bStats' standards, which is by default an opt-out system. We know some people are paranoid and would like bStats off - this is an option we do not even try to hide and explicitly warn about on startup, so we're not trying to be evasive in any way.
2. You've already chosen to give up a significant amount of your freedoms and privacy when you use closed-source software, like _Minecraft_. Combine that with the fact that most Minecraft players play using Windows, an operating system not well-known for respecting your privacy, sending a few stats to bStats has _nothing_ on what Microsoft can do to you. You may say that Velocity is open source, but the truth is that Velocity is completely useless without Mojang's proprietary implementation of the _Minecraft_ game. We might as well have just written `System.out.println("Hello, Velocity!");` and called it a day.

None of this is to say that we don't care about privacy, but if you're concerned about Velocity sending statistics to a third party service, there are _far_ bigger concerns you also have, and turning off this data collection by default will have at best a negligible positive impact on privacy, but harms the Velocity project as a whole far more.

PR #666, opened 2022-03-15
Added MiniMessage support to built-in motd

4drian3d:
- MiniMessage support has been added to build-in motd replacing legacy color codes and json usage
- Bump config-version to 2.6

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
@@ -274,11 +274,7 @@ public class VelocityConfiguration implements ProxyConfig {
   @Override
   public net.kyori.adventure.text.Component getMotd() {
     if (motdAsComponent == null) {
-      if (motd.startsWith("{")) {
-        motdAsComponent = GsonComponentSerializer.gson().deserialize(motd);
-      } else {
-        motdAsComponent = MiniMessage.miniMessage().deserialize(motd);
-      }
+      motdAsComponent = MiniMessage.miniMessage().deserialize(motd);
     }
     return motdAsComponent;
   }
@@ -534,18 +530,29 @@ public class VelocityConfiguration implements ProxyConfig {
 
     String motd = config.getOrElse("motd", "<#09add3>A Velocity Server");
 
-    // MOTD Migration
+    // Old MOTD Migration
     if (configVersion < 2.6) {
-      if (!motd.startsWith("{")) {
-        final String migratedMotd = MiniMessage.miniMessage().serialize(
-                LegacyComponentSerializer.legacyAmpersand().deserialize(motd));
+      String migratedMotd = motd;
 
-        config.set("motd", migratedMotd);
-        motd = migratedMotd;
+      // JSON Format Migration
+      final String strippedMotd = migratedMotd.strip();
+      if (strippedMotd.startsWith("{\"") || strippedMotd.startsWith("[\"")) {
+        migratedMotd = MiniMessage.miniMessage().serialize(
+                GsonComponentSerializer.gson().deserialize(migratedMotd))
+                .replace("\\", "");
       }
+
+      // Legacy '&' Format Migration
+      migratedMotd = MiniMessage.miniMessage().serialize(
+              LegacyComponentSerializer.legacyAmpersand().deserialize(migratedMotd))
+              .replace("\\", "");
+
+      config.set("motd", migratedMotd);
+      motd = migratedMotd;
+
       config.setComment("motd",
               " What should be the MOTD? This gets displayed when the player adds your server to\n"
-                      + "their server list. MiniMessage format and JSON are accepted.");
+                      + " their server list. Only MiniMessage format are accepted.");
       config.set("config-version", "2.6");
       mustResave = true;
     }
--- a/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
@@ -553,16 +553,17 @@ public class VelocityConfiguration implements ProxyConfig {
     String motd = config.getOrElse("motd", "<#09add3>A Velocity Server");
 
     if ("1.0".equals(configurationVersion)) {
-      final String migratedMotd = MiniMessage.miniMessage().serialize(
-          LegacyComponentSerializer.legacy('&').deserialize(motd));
+      if (!motd.startsWith("{")) {
+        final String migratedMotd = MiniMessage.miniMessage().serialize(
+            LegacyComponentSerializer.legacy('&').deserialize(motd));
 
-      config.set("motd", migratedMotd);
+        config.set("motd", migratedMotd);
+        motd = migratedMotd;
+      }
       config.setComment("motd", 
           " What should be the MOTD? This gets displayed when the player adds your server to\n" 
           + " their server list. MiniMessage format and JSON are accepted.");
       config.set("config-version", "1.1");
-
-      motd = migratedMotd;
     }
 
     int maxPlayers = config.getIntOrElse("show-max-players", 500);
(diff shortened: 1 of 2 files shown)


proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
```diff
@@ -59,7 +61,7 @@
   private static final Logger logger = LogManager.getLogger(VelocityConfiguration.class);
 
   @Expose private String bind = "0.0.0.0:25577";
-  @Expose private String motd = "&3A Velocity Server";
+  @Expose private String motd = "<dark_aqua>A Velocity Server";
```
astei: Nit: needs to be `<mm><dark_aqua>A Velocity Server`

astei:
I'm happy with this but would like some input from others before I merge.

KoxSosen (approved):
I'm not a contributor, but this seems like a great change.

kezz:
A prefix is really weird - is there any reason not to migrate the config?

kyngs:
> A prefix is really weird - is there any reason not to migrate the config?

That would mean peoples motd's breaking on a minor update. Not having the prefix would be a good idea on velocity 4.x IMO.

4drian3d:
> A prefix is really weird - is there any reason not to migrate the config?

I thought it would be the best option, to migrate the formatting of that option in a "patch" version like 3.1.2 would be strange

kezz:
> That would mean peoples motd's breaking on a minor update. Not having the prefix would be a good idea on velocity 4.x IMO.

Do you know what migration means? The point of migration is that nothing breaks.

> to migrate the formatting of that option in a "patch" version like 3.1.2 would be strange

Then let's bump the minor version? I assume Velocity, being a modern proxy, would want to migrate away from legacy formatting eventually, so why not do it in one step?

kyngs:
> > That would mean peoples motd's breaking on a minor update. Not having the prefix would be a good idea on velocity 4.x IMO.
> 
> Do you know what migration means? The point of migration is that nothing breaks.
> 

I'm sorry, I understood it as if it were to just switch to mm. How do you want to migrate such a thing?

Astralchroma:
I made a similar pull request to Paper, and after discussion there the course of action was to forcibly migrate the config messages instead of maintain compatibility with both, would it not be best to be consistent and forcibly migrate the format with Velocity?

4drian3d:
> > That would mean peoples motd's breaking on a minor update. Not having the prefix would be a good idea on velocity 4.x IMO.
> 
> Do you know what migration means? The point of migration is that nothing breaks.
> 
> > to migrate the formatting of that option in a "patch" version like 3.1.2 would be strange
> 
> Then let's bump the minor version? I assume Velocity, being a modern proxy, would want to migrate away from legacy formatting eventually, so why not do it in one step?

Yes, the ideal would be to migrate at once from all legacy color code uses to MiniMessage, but being a minor version to be released in proximity, I thought better to use a prefix, but if required, I could try to migrate to MiniMessage in that option now

kezz:
> How do you want to migrate such a thing?

Velocity has a versioned config, there's a million and one ways you could easily setup migrations.

4drian3d:
> > How do you want to migrate such a thing?
> 
> Velocity has a versioned config, there's a million and one ways you could easily setup migrations.

Honestly, I never paid attention to this configuration option, I just remembered it existed, now I am making the migration

Astralchroma:
MiniMessage can convert components to MiniMessage so it's not hard to convert it.

4drian3d:
migration1
The migration and the motd format are working correctly, it should be ready for another review

astei:
I'm happy with this but am going to hold this off until we get a new minor version out.

wiktor-jozwiak:
will this be implemented? (About when this version of Velocity will come out)

4drian3d:
> I'm happy with this but am going to hold this off until we get a new minor version out.

Considering that pull request #692 was accepted and that very soon pull request #690 will be accepted, I believe that the next version will no longer be considered as a "patch" and this pull request could be accepted directly.

Also, it would be nice if the first version that includes MiniMessage already has its use implemented in all the places where the legacy format was used (where possible) which is the only place that would be missing

proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
```diff
@@ -276,7 +277,7 @@ public net.kyori.adventure.text.Component getMotd() {
       if (motd.startsWith("{")) {
```
kezz: Might it be a good time to ditch JSON support too in this tbh. There are valid JSON text components that don't start with `{` and MiniMessage can represent everything that JSON can so there's no need to still have it.

4drian3d: According to what you say, I just checked and any slightly complex JSON formatting in this option was unusable since the [same check](https://github.com/PaperMC/Velocity/blob/dev/3.0.0/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java#L276) was used to parse JSON 😅

I just added migration support for the JSON format, but I had to add 2 additional `.replace("\\", "")` because otherwise the Gson serializer would add that character, I don't know if it's related to the fix https://github.com/KyoriPowered/adventure/pull/789 but it doesn't work

proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
```diff
@@ -274,11 +274,7 @@ public boolean shouldQueryShowPlugins() {
   @Override
   public net.kyori.adventure.text.Component getMotd() {
     if (motdAsComponent == null) {
-      if (motd.startsWith("{")) {
-        motdAsComponent = GsonComponentSerializer.gson().deserialize(motd);
-      } else {
-        motdAsComponent = MiniMessage.miniMessage().deserialize(motd);
-      }
+      motdAsComponent = MiniMessage.miniMessage().deserialize(motd);
```
kezz, line 277: There's a deserializeOrNull method you can use just to save the if check here

proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
```diff
@@ -534,18 +530,29 @@ public static VelocityConfiguration read(Path path) throws IOException {
...
-        final String migratedMotd = MiniMessage.miniMessage().serialize(
-                LegacyComponentSerializer.legacyAmpersand().deserialize(motd));
-
-        config.set("motd", migratedMotd);
-        motd = migratedMotd;
+      String migratedMotd = motd;
+
+      // JSON Format Migration
+      final String strippedMotd = migratedMotd.strip();
+      if (strippedMotd.startsWith("{\"") || strippedMotd.startsWith("[\"")) {
```
kezz: imo it's only worth doing exactly the same check that was done when the component was retrieved originally (just the `{` check), as nothing else would've been displayed as JSON anyway

proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
4drian3d: I think it is better to leave it as it is, since the only time the motd of the configuration changes, is when the whole configuration is reloaded, so it would be an unnecessary waste of resources to always call the `MiniMessage#deserialize` or `#deserializeOrNull` method if the motd will always be the same

proxy/src/main/resources/default-velocity.toml
```diff
@@ -1,12 +1,12 @@
...
-config-version = "2.5"
+config-version = "2.6"
 
 # What port should the proxy be bound to? By default, we'll bind to all addresses on port 25577.
 bind = "0.0.0.0:25577"
 
 # What should be the MOTD? This gets displayed when the player adds your server to
-# their server list. Legacy color codes and JSON are accepted.
-motd = "&#09add3A Velocity Server"
+# their server list. MiniMessage format and JSON are accepted.
```
kezz: need to remove the JSON bit here

astei (approved):

PR #671, opened 2022-03-30
Added ServerPing.Builder#hoverMessage

4drian3d:
Added a direct method to set a hover message in the ping response. This will make it easier to set a custom hover in the motd using formats like ~~legacy~~ or **MiniMessage**.

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/proxy/server/ServerPing.java
+++ b/api/src/main/java/com/velocitypowered/api/proxy/server/ServerPing.java
@@ -189,13 +189,9 @@ public final class ServerPing {
       SamplePlayer[] players = new SamplePlayer[components.length];
       for (int i = 0; i < components.length; i++) {
         players[i] = new SamplePlayer(
-          net.kyori.adventure.text.serializer.legacy.LegacyComponentSerializer
-            .builder()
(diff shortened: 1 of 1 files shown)


kezz:
You shouldn't really be using the unusual hex code format - it is not part of the native legacy format, it was a hacky spigot addition. This should be downsampling the colours using the methods available on the builder (plus it should be being stored somewhere rather than being recreated in every single loop).

hugmanrique:
This isn't the intended use of the sample players feature, and it can be recreated by a plugin using the current API, so I'm not sold on merging this.

4drian3d:
> This isn't the intended use of the sample players feature, and it can be recreated by a plugin using the current API, so I'm not sold on merging this.

I knew someone would say that, and yes, I know it's not a "proper" use and that at any time mojang could remove the legacy color codes, but I leave the pull request open in case someone has something else to say

electronicboy:
I'm inclined to agree, I don't think that stuff relying on deprecated behavior which is already marked for removal should be part of the core API, this API would basically be deprecated on day 1 (but, this is on the list of stuff which needs to be consulted with mojang about potentially providing a more rich experience here, but 🤷 )

astei:
Closing this PR. Unfortunately, what @electronicboy said is right: this is something Mojang is likely to deprecate and it would be better if there was a more rich, native experience here instead.

PR #675, opened 2022-04-01
Fix implements in class ResponseResult

microwin7:
Fixed inheritance of interface on ResultedEvent.Result in ResponseResult class

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/event/player/ServerLoginPluginMessageEvent.java
+++ b/api/src/main/java/com/velocitypowered/api/event/player/ServerLoginPluginMessageEvent.java
@@ -114,7 +114,7 @@ public class ServerLoginPluginMessageEvent implements ResultedEvent<ResponseResu
         + '}';
   }
 
-  public static class ResponseResult implements Result {
+  public static class ResponseResult implements ResultedEvent.Result {
 
     private static final ResponseResult UNKNOWN = new ResponseResult(null);
 


astei:
I'm concerned that this could be a breaking change. What issue does this fix?

4drian3d:
> I'm concerned that this could be a breaking change. What issue does this fix?

Currently there is always a compiler error because "Result" is not found among the imports of class
Screenshot_1
For some reason, when compiling the project, no errors appear and everything works normally

Replacing it with ResultedEvent.Result eliminates the error because the ResultedEvent class is imported in the class.

This is not a big change and will not break anything because if you look at javadocs https://jd.velocitypowered.com/3.0.0/com/velocitypowered/api/event/player/ServerLoginPluginMessageEvent.ResponseResult.html there you actually use the ResultedEvent.Result
Screenshot_2

astei:
> Currently there is always a compiler error because "Result" is not found among the imports of class

This sounds like a bug in the Java extension for Visual Studio Code. IntelliJ, Eclipse, and the Java compiler all resolve the import correctly.

> This is not a big change and will not break anything

I'm convinced but this is the wrong fix - the correct fix is to file a bug with the authors of the Java extension for Visual Studio Code.

hugmanrique:
This is indeed a bug in the VS Code extension. Quoting [JLS 6.3](https://docs.oracle.com/javase/specs/jls/se8/html/jls-6.html#jls-6.3),
> The scope of a declaration of a member m declared in or inherited by a class type C ([§8.1.6](https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.1.6)) is the entire body of C, including any nested type declarations.

The `Result` interface is declared in the `ResultedEvent` interface, which `ServerLoginPluginMessageEvent` implements (and thus inherits, see [JLS 8.2](https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.2)).

PR #678, opened 2022-04-05
Add "stop" as an alias to "shutdown"

kyngs:
Many (Minecraft) server hosts shut down servers by entering the "stop" command, instead of sending SIGINT to the process. This is because many server softwares have issues with it. (Velocity also had issues with it until #637). This should not break anything and is just a quality of life patch.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
@@ -209,7 +209,7 @@ public class VelocityServer implements ProxyServer, ForwardingAudience {
     // Initialize commands first
     commandManager.register("velocity", new VelocityCommand(this));
     commandManager.register("server", new ServerCommand(this));
-    commandManager.register("shutdown", new ShutdownCommand(this),"end");
+    commandManager.register("shutdown", new ShutdownCommand(this),"end", "stop");
     new GlistCommand(this).register();
 
     this.doStartupConfigLoad();


tomcraft:
How do you stop your backend server if the proxy also uses the /stop command ? 😅

kyngs:
> How do you stop your backend server if the proxy also uses the /stop command ? sweat_smile

Oh no, I haven't thought of that :). I will leave this PR open in case someone still thinks it is worth it.

4drian3d:
Unless the command/alias is exclusively executable from the console, I think it could cause many more problems than it solves

Astralchroma:
> /stopv

This would be good as it would show up in auto complete for /stop, while being clear it is for the proxy. You could also do /stopvelocity which would have a similar advantage.

kyngs:
> > 
> 
> This would be good as it would show up in auto complete for /stop, while being clear it is for the proxy. You could also do

But it kinda defeats the whole point of this PR :D

kyngs:
When I think about it (as @4drian3d) indicated, if this was a console-only alias, it would be fine. Tho I'm not sure if it is possible.

4drian3d:
> When I think about it (as @4drian3d) indicated, if this was a console-only alias, it would be fine. Tho I'm not sure if it is possible.

~~Perhaps creating a new command, identical to the shutdown/end command, but that in its hasPermission method checks for the corresponding permission and that the CommandSource is an instance of the console.~~ The shutdown command already makes this check

tomcraft:
But whats the point of this ? Any host provider should just adjust their scripts/things to use "/shutdown" as stop command instead. BungeeCord had "/end" as its stop command for years.

kyngs:
> But whats the point of this ? Any host provider should just adjust their scripts/things to use "/shutdown" as stop command instead. BungeeCord had "/end" as its stop command for years.

Velocity is still a pretty niche solution, that most people do not use. Therefore, most providers do not have a separate script for it.

tomcraft:
I don't think that's on Velocity to do their job, and the "end" alias also work to help moving from BungeeCord.

astei:
I'm going to be somewhat controversial here but I approve of this, primarily because Velocity will only accept the stop command from the console anyway (it will otherwise forward the command to the server), so Velocity won't get in front of the server.

I've gone ahead and committed this myself (primarily because this PR has a merge conflict).

PR #679, opened 2022-04-05
Added translatable console messages and more

4drian3d:
Translations have been added to the console messages along with an extra translation for the "could-not-move-to-fallback" error.

_Originally I was only going to add a translation to the offline mode warning message as I thought it was important to have it translated, but I ended up adding more translations, that's why the branch is called this way_


astei:
Not sure of the value of this. The messages are in English for a good reason: the primary developers of Velocity and Paper speak English as their native tongue or speak it fluently, and we need to be able to read errors to diagnose issues.

4drian3d:
Yes, I know that several console messages should never be translated such as errors and exceptions in order to give universal support to everyone. For that reason, in this pull request I only added translations to certain console messages that are not circumstantial and only serve to give information, such as loaded plugins and other things, especially the warning message about the use of Offline Mode, which is the basis of this pull request

olijeffers0n:
I can see the value of this, if the vital support messages are not translated then this is the best of both worlds as far as I can see.

PR #688, opened 2022-04-13
Limiting inbound EncryptionResponse packets

hevav:
Prohibiting players from sending more than one EncryptionResponse packet, as this could cause DoS-attacks. Attackers don't even need to have multiple IPs to perform this attack, as we can send too many EncryptionResponse packets from one player. The attack overloads both CPU and RAM.

Before, during the attack:
Firstly, this error appears
image
Then that error
image
And then (even when the attack is stopped, as I sent 10000 EncryptionResponses from one player)
image

After this patch:
Just works well.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/client/LoginSessionHandler.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/client/LoginSessionHandler.java
@@ -82,6 +82,7 @@ public class LoginSessionHandler implements MinecraftSessionHandler {
   private @MonotonicNonNull ServerLogin login;
   private byte[] verify = EMPTY_BYTE_ARRAY;
   private @MonotonicNonNull ConnectedPlayer connectedPlayer;
+  private boolean receivedEncryptionResponse = false;
 
   LoginSessionHandler(VelocityServer server, MinecraftConnection mcConnection,
       LoginInboundConnection inbound) {
@@ -113,6 +114,12 @@ public class LoginSessionHandler implements MinecraftSessionHandler {
     if (verify.length == 0) {
       throw new IllegalStateException("No EncryptionRequest packet sent yet.");
     }
+
+    if (receivedEncryptionResponse) {
+      throw new IllegalStateException("Too many EncryptionResponse packets received.");
+    }
+
+    receivedEncryptionResponse = true;
 
     try {
       KeyPair serverKeyPair = server.getServerKeyPair();


proxy/src/main/java/com/velocitypowered/proxy/connection/client/LoginSessionHandler.java
```diff
@@ -82,6 +82,7 @@ public class LoginSessionHandler implements MinecraftSessionHandler {
   private @MonotonicNonNull ServerLogin login;
   private byte[] verify = EMPTY_BYTE_ARRAY;
   private @MonotonicNonNull ConnectedPlayer connectedPlayer;
+  private boolean receivedEncryptionResponse = false;
```
Xernium, line 85: While this works I think this class should have a state enum for tracking the process rather. Just a note for future refactors

proxy/src/main/java/com/velocitypowered/proxy/connection/client/LoginSessionHandler.java
```diff
@@ -114,6 +115,12 @@ public boolean handle(EncryptionResponse packet) {
       throw new IllegalStateException("No EncryptionRequest packet sent yet.");
     }
 
+    if (receivedEncryptionResponse) {
+      throw new IllegalStateException("Too many EncryptionResponse packets received.");
```
Xernium, line 119: Also begs the question wether or not we should throw cached exceptions here

Xernium (approved):
I don’t see any issue with this but I’d still like your opinion @astei before merging this

proxy/src/main/java/com/velocitypowered/proxy/connection/client/LoginSessionHandler.java
electronicboy: This was honestly my thought too, just figured I'd leave this PR to those more involved with this

proxy/src/main/java/com/velocitypowered/proxy/connection/client/LoginSessionHandler.java
electronicboy: potentially, idk how it will handle it if a flood of packets come in, will it throw on just the first or will it drain the others out? if it's just a single throw then it's not *too* bad, but, this is a stupidly easy area to send packets to, so, maybe much of this stuff should be cached exceptions?

astei:
I would have much rather we implemented a real state machine instead of this - but alas, I'm short on time these days.

hevav:
What do you mean by "a real state machine"? What should we do since Velocity has a DoS vulnerability?

electronicboy:
by a real state machine they mean tracking the state of the connection properly

Unless somebody has any objections, or the state machine can be done fast, I am somewhat inclined to pull this, that exception should however be cached/stackless (I forget if velocity does stackless ones, if not, ignore that)

proxy/src/main/java/com/velocitypowered/proxy/connection/client/LoginSessionHandler.java
astei: this should be a stackless exception

astei (changes requested):

hevav:
Shouldn't the other exceptions in this method be replaced with stackless ones? Should we disconnect the player if we receive a "duplicated" EncryptionResponse?

ghost:
> Shouldn't the other exceptions in this method be replaced with stackless ones? Should we disconnect the player if we receive a "duplicated" EncryptionResponse?

In my opinion, yes, the client cannot send more than an encryption packet afaik

kyngs:
> Shouldn't the other exceptions in this method be replaced with stackless ones? Should we disconnect the player if we receive a "duplicated" EncryptionResponse?

I thought we are, it makes perfect sense to kick the player.

astei:
I have reimplemented this using a proper state machine.

Issue #248, opened 2019-09-20
Forge 1.14 support


astei:

astei:
There has been a general lack of interest in supporting Forge versions past 1.12.2, so I'm taking this off the roadmap. We will still entertain support for Forge 1.13+ in Velocity if there is sufficient commitment to supporting it and the Forge team offers a viable path towards adding proxy support with their new handshaking protocol.

phit:
That's sad to hear, can't really gather interest if there is no support. A lot of people are stuck on older versions with modded right now due to stuff like Sponge still not having updated. 
Are there any open Forge issues since you seem to hint at Forge adding proxy support?

XinKe-Conch:
We really need forge. Please don't give it up. Plz

astei:
@XinKe-Conch Velocity still supports Forge 1.7-1.12.2 and we do not intend to drop support for them any time soon. We are referring to support for Forge 1.14+.

XinKe-Conch:
Yes, although the 1.14 + forge is gradually replaced by fabric, there are still some excellent mod developers using forge for development. I hope you can support it

Dec1lent:
> @XinKe-Conch Velocity still supports Forge 1.7-1.12.2 and we do not intend to drop support for them any time soon. We are referring to support for Forge 1.14+.

Is there a way to support 1.4.7 forge on ProtocolSupportVelocity? Im running multiple modpacks in all different versions (💯

astei:
It is extremely unlikely Velocity will support ancient versions of Minecraft, let alone ancient versions of Forge.

astei:
I wanted to provide a little update on this. There is currently a work-in-progress mod to try and enable this on Forge 1.16+ by reintroducing the old-style handshakes in a newer version of Forge that is in cahoots with the proxy to allow switching over to the login state and redo the handshake. I do not know when this work will be complete but there is already a proof of concept.

PurpleIsEverything:
While this is good news one should remember any such solution will absolutely require a client side mod.

You then have to get modpacks and the like to include that mod otherwise it doesn't help all that much. Also the possibility of such a solution causing a great many mod specific issues is staggering.

This solution mentioned is more functional as a proof of concept to give direction to a proper native Forge solution.

PurpleIsEverything:
I should also add we had discussions with Lex he has been willing to accept a PR containing this functionality for a while he simply was not willing to put the time into the creation himself.

[...]

PR #690, opened 2022-04-15
Modern Forge (1.13+) support

danorris709:
Adds a fix for https://github.com/PaperMC/Velocity/issues/248

Updates the Forge connection logic for the new Forge client handshake and adds the "reset packet" sending implementation. There is a pending Forge fix for interpreting this packet, however this has to come first (especially given the fix doesn't have to exist in Forge for the solution to work to begin with).

This has all been tested with Vanilla, Spigot, and Forge (and Forge with mods) clients and servers interchangeably.


Xernium:
While I really welcome this I really don’t want another forge-specific implementation.
I’m not putting in my veto though and I’m far too busy at the moment to oversee this so I’m leaving it in the hands of either @electronicboy or @astei or another paper dev team member to decide.
What I don’t like however is how the brigadier types are handled, in my opinion a capsulation should be negotiated so this can work with the existing CrossStitch code.
And the breaking api change has to go for this to be considered within the next minor version releases, please correct that.
But other than that I certainly didn’t expect this and it’s a welcome surprise.

api/src/main/java/com/velocitypowered/api/util/ModInfo.java
```diff
@@ -62,19 +64,23 @@ public static final class Mod {
...
+      this.version = version;
     }
 
     public String getId() {
       return id;
     }
 
-    public String getVersion() {
-      return version;
+    public Optional<String> getVersion() {
```
astei: This is a breaking API change. I would probably just return an empty string in this case.

proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/AvailableCommands.java
```diff
@@ -81,8 +89,30 @@ public RootCommandNode<CommandSource> getRootNode() {
   @Override
   public void decode(ByteBuf buf, Direction direction, ProtocolVersion protocolVersion) {
     int commands = ProtocolUtils.readVarInt(buf);
+
+    // When Minecraft serializes this packet the nodes are written in order of id,
+    // as the RootCommandNode is the first node processed it should also be the first node in the array,
+    // Velocity can read the first node of the array to check the type of the node,
+    // if it is the RootCommandNode then we can manipulate this packet without deserializing every node.
+
+    // In order to support Forge, Velocity cannot deserialize every node as it does not support custom parsers,
```
astei: Velocity does, in fact, support custom argument parsers. All you have to do is wrap non-vanilla commands [like we do in CrossStitch](https://github.com/VelocityPowered/CrossStitch/blob/master/src/main/java/com/velocitypowered/crossstitch/mixin/command/CommandTreeSerializationMixin.java). This is probably portable enough to carry over into Forge too.

The fact this "support" completely neuters `PlayerAvailableCommandsEvent` is a non-starter for merging this.

astei (changes requested):
I am actually familiar with this work. There are some show-stopper issues with this, though.

The biggest showstopper of all is if the Forge team is willing to accept this. We are more than willing to accept this and put in the legwork - by popular request, up to the point where I have received offers from people willing to compensate me in cold hard cash for this.

Rasgnarok:
Pixelmon has been promoting this solution for which we've been discussing with Lex on Forge's role in this. Much like Velocity, they will have their own changes to propose, but Forge is absolutely willing to accept a solution proxies will use @astei .

proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/AvailableCommands.java
danorris709: So pretty much this section is redundant and can be removed and then a PR can be made to the CrossStitch repository to add Forge support?

LexManos:
Just  to be clear on Forge's end. We are not against adding a hook for the proxy community. As long as it's sane/not bespoke. If we have to add a custom thing for every proxy project out there it's not going to happen. So if the proxy community is able to agree on a solution, and the implementation on Forge's end is sane. Then I'll be willing to entertain it.

From my perspective, the main requirement I have is it MUST fully disconnect the user and let him reconnect. Any other solution would just cause more potential headakes.

The rest of it is up for you guys to decide:
Is there data that needs to be synced to the proxy?
Is this allowing reconnecting to a different ip/port?
How to keep it secure for end users, so that random servers cant send them to other random servers?
Anything else you guys need to agree on.

Once then it should be written up as a spec outside of thee implementation so that other proxy {and modloader}  services can implement it.

Yes, the Pixelmon people have talked to me about this, and are pushing for it, Same with the guy who offered large sums of cash to fix it. I have not agreed to include any solution because I have no seen a spec that has been agreed upon by your community. But I am willing to entertain it if you guys can actually agree on something.

proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/AvailableCommands.java
astei: I believe Forge has its own custom argument types - we'd have to be able to parse them. I would be fine with adding these directly into Velocity.

astei: CrossStitch should be the solution for mods with custom argument types, yes.

danorris709: Okay, perfect. Thank you for the clarification I'll take a look at getting this resolved tonight

astei:
> As long as it's sane/not bespoke. If we have to add a custom thing for every proxy project out there it's not going to happen. 

Agreed on this one.

> Is there data that needs to be synced to the proxy?

As long as Forge is able to re-negotiate the client-server handshake, then we're fine. In fact, it would be best if Velocity was _not_ involved beyond forwarding messages and sending the reset packet, the more to make the support portable.

> Is this allowing reconnecting to a different ip/port?

No. From the client side, the player always remains connected to the proxy, which is acting as the server. The only change is that Velocity is sending the Join Game and Respawn packet to reset the client state for the new server. From there, Velocity just forwards packets back and forth.

> How to keep it secure for end users, so that random servers cant send them to other random servers?

Server switches are _always_ mediated by the proxy, which will only do the switch in response to a disconnect from the server (in which case Velocity will attempt to fall back to a different server) or someone firing a connection request (be that through the API, or most likely, the BungeeCord plugin messaging channel).

danorris709:
I believe this solution (and the soon to follow Forge side) matches all the above requirements and doesn't create any vulnerabilities. I am confident it's not specific to one proxy solution and am perfectly happy to write up the specification for others to follow if necessary

LeStegii:
grafik

Sorry for the meme, but is this really just happening?

proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/AvailableCommands.java
danorris709: I believe I resolved this with the latest commit

Here is the Forge registration: https://github.com/MinecraftForge/MinecraftForge/blob/1.18.x/src/main/java/net/minecraftforge/common/ForgeMod.java#L175
Here are the Forge classes: 
https://github.com/MinecraftForge/MinecraftForge/blob/1.18.x/src/main/java/net/minecraftforge/server/command/ModIdArgument.java
https://github.com/MinecraftForge/MinecraftForge/blob/1.18.x/src/main/java/net/minecraftforge/server/command/EnumArgument.java

proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/brigadier/EnumArgumentProperty.java
```diff
@@ -0,0 +1,23 @@
+package com.velocitypowered.proxy.protocol.packet.brigadier;
```
astei: Nit: I'd move this to `com.velocitypowered.proxy.protocol.packet.brigadier.forge`

astei: The license header is missing here too.

proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/brigadier/EnumArgumentPropertySerializer.java
```diff
@@ -0,0 +1,43 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package com.velocitypowered.proxy.protocol.packet.brigadier;
```
astei: Nit: I'd move this to `com.velocitypowered.proxy.protocol.packet.brigadier.forge`

proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/brigadier/ArgumentPropertyRegistry.java
```diff
@@ -136,6 +136,21 @@ public void serialize(BoolArgumentType object, ByteBuf buf) {
     // Crossstitch support
     register("crossstitch:mod_argument", ModArgumentProperty.class, MOD);
 
+    // Forge support
+    register("forge:enum", EnumArgumentProperty.class, EnumArgumentPropertySerializer.ENUM);
+    register("forge:modid", ModIdArgumentProperty.class,
+            new ArgumentPropertySerializer<>() {
```
astei: The indenting needs to be fixed here

proxy/src/main/java/com/velocitypowered/proxy/connection/backend/LoginSessionHandler.java
```diff
@@ -136,8 +166,27 @@ public boolean handle(ServerLoginSuccess packet) {
...
+      success.setUsername(player.getUsername());
+      success.setUuid(playerUniqueId);
+      pmc.write(success);
+
+      pmc.setState(StateRegistry.PLAY);
+    }
+
     // Switch to the transition handler.
-    smc.setSessionHandler(new TransitionSessionHandler(server, serverConn, resultFuture));
+    smc.setSessionHandler(new BackendTransitionSessionHandler(server, serverConn, resultFuture));
```
astei, line 199: I like this new name

proxy/src/main/java/com/velocitypowered/proxy/connection/backend/LoginSessionHandler.java
```diff
@@ -65,6 +71,26 @@ public class LoginSessionHandler implements MinecraftSessionHandler {
     this.resultFuture = resultFuture;
   }
 
+  @Override
+  public void activated() {
+    if (serverConn.getPlayer().getConnection().getType() == ConnectionTypes.MODERN_FORGE) {
```
astei, line 87: It would be helpful to explain _why_ we need this logic. I know why (the client is renegotiating the handshake with backend mods, and if someone goes wrong on the server side then there is no way to recover) but this needs to be documented.

astei (changes requested):
A couple more changes and we can merge this.

proxy/src/main/java/com/velocitypowered/proxy/connection/backend/LoginSessionHandler.java
danorris709: All credit to @LXGaming for that :)

proxy/src/main/java/com/velocitypowered/proxy/connection/backend/LoginSessionHandler.java
danorris709: I believe what I've written is accurate but probably worth double checking

astei: Aside from some typos that explanation looks good.

LexManos:
> No. From the client side, the player always remains connected to the proxy, which is acting as the server. The only change is that Velocity is sending the Join Game and Respawn packet to reset the client state for the new server. From there, Velocity just forwards packets back and forth.

I understand how proxies currently work yes, that does not work in a modded environment because you can not get back into the handshake mode. And things like registry data is still in use. Which is why I said:
>From my perspective, the main requirement I have is it MUST fully disconnect the user and let him reconnect. Any other solution would just cause more potential headakes.

So if you maintain the connection it's a non-starter. I've stated this clearly.

Also, those following questions were not something you should answer to me. They are something other proxy providers have brought up as concerns and you guys have to decide upon themselves. Hence, `Is there data that needs to be synced to the proxy?` As it is a fresh connection there may be data the proxy needs to know to figure out which backend server you're trying to connect to.

I have tried to explain this as clearly as I can to both you guys, and the Pixelmon folks. If they did not understand that and have started this process with the wrong approach then that is on them. I am leaving it to your side of the community {not JUST Paper, it has to take the other proxy providers into account} to design a spec. On Forge's end it's essentially:
```
client.disconnect();
ConnnectScreen.startConnecting(currentServer, currentPort)
```
The entire point is that you get kicked out to essentially the main menu, so that the world, player instance, and everything associated with it gets destroyed. Thus making sure that the rendering thread or anything else are not using registry data. Because registry data needs to be swapped out.

I recommend before you even consider merging this, you do as I instructed the Pixelmon folks, and design a spec that is agreed upon by the other proxy communities, and is documented outside of some random project's implementation.

danorris709:
> > No. From the client side, the player always remains connected to the proxy, which is acting as the server. The only change is that Velocity is sending the Join Game and Respawn packet to reset the client state for the new server. From there, Velocity just forwards packets back and forth.
> 
> I understand how proxies currently work yes, that does not work in a modded environment because you can not get back into the handshake mode. And things like registry data is still in use. Which is why I said:
> 
> > From my perspective, the main requirement I have is it MUST fully disconnect the user and let him reconnect. Any other solution would just cause more potential headakes.
> 
> So if you maintain the connection it's a non-starter. I've stated this clearly.
> 
> Also, those following questions were not something you should answer to me. They are something other proxy providers have brought up as concerns and you guys have to decide upon themselves. Hence, `Is there data that needs to be synced to the proxy?` As it is a fresh connection there may be data the proxy needs to know to figure out which backend server you're trying to connect to.
> 
> I have tried to explain this as clearly as I can to both you guys, and the Pixelmon folks. If they did not understand that and have started this process with the wrong approach then that is on them. I am leaving it to your side of the community {not JUST Paper, it has to take the other proxy providers into account} to design a spec. On Forge's end it's essentially:
> 
> ```
> client.disconnect();
> ConnnectScreen.startConnecting(currentServer, currentPort)
> ```
> 
> The entire point is that you get kicked out to essentially the main menu, so that the world, player instance, and everything associated with it gets destroyed. Thus making sure that the rendering thread or anything else are not using registry data. Because registry data needs to be swapped out.
> 
> I recommend before you even consider merging this, you do as I instructed the Pixelmon folks, and design a spec that is agreed upon by the other proxy communities, and is documented outside of some random project's implementation.

I am sorry, but you are totally and utterly incorrect here @LexManos. I have been in a call with you before where you have outright agreed on statements that directly contradict what you've stated here.

I strongly recommend that you further investigate this issue further, and consider conversations of the past, before making anymore incomprehensible claims.

danorris709:
> I understand how proxies currently work yes, that does not work in a modded environment because you can not get back into the handshake mode. And things like registry data is still in use. Which is why I said:
> 

This statement is incorrect. We have an implementation that does not require a complete disconnect and reconnect packet (using IPs etc.) and that is just not hpw a proxy would work as @astei said.





> So if you maintain the connection it's a non-starter. I've stated this clearly.

As I said above, it is a complete starter and it works. 



> I have tried to explain this as clearly as I can to both you guys, and the Pixelmon folks. If they did not understand that and have started this process with the wrong approach then that is on them. I am leaving it to your side of the community {not JUST Paper, it has to take the other proxy providers into account} to design a spec. On Forge's end it's essentially:

We did not start this process with the wrong approach. We spoke to you two months ago about this, and since we have contacted every single developer possible to attempt to meet your insane requirements. The fact is there is not a single developer in the community who wants to work on this as a result of your attitude towards it.

The proof of that being it has taken two months for us to get anywhere with it, and you've not only ignored our personal messages to you for this but you've chosen to publicly attack the pull request that is on a completely different platform. You could've waited for us to PR Forge and then shut us down rather than going out of your way to shut it down as early as possible. 



> The entire point is that you get kicked out to essentially the main menu, so that the world, player instance, and everything associated with it gets destroyed. Thus making sure that the rendering thread or anything else are not using registry data. Because registry data needs to be swapped out.

Again, you're completely wrong here and it further proves you really do not know what you're talking about in relation to this. The implementation here is perfectly fine and can be replicated anywhere else. I will have a specification document ready before I make the PR to Forge so please be prepared.

astei:
We are not going to have mud-slinging in this thread. I _truly_ want this to succeed - users want it, it's a big hole we want filled, and we cannot let petty arguments get in the way.

I would recommend that @danorris709 document how this is going to fully work, on both the proxy and Forge sides before we proceed any further. I believe this is a reasonable request, and it'll also assist in adding modern Forge client support to Waterfall and to any other proxies that wish to support modern Forge.

As for the implementation of this feature, implementing this as a full disconnection and reconnection to the proxy is _possible_, but not desirable from our view. (I see a path forward with disconnecting the client and reconnecting them with a "cookie", but there are significant problems with this - I cannot elaborate further due to IRL circumstances.) The approach outlined by @danorris709 is consistent with how Forge support has been implemented in BungeeCord and Velocity prior to Forge for Minecraft 1.13. I think the registry sync issues can be worked around - Velocity has been successfully resetting the client state using methods native to the game, and I feel that there has to be a path with native Forge support as well.

ME1312:
> I understand how proxies currently work yes, that does not work in a modded environment because you can not get back into the handshake mode. And things like registry data is still in use. Which is why I said:
> 
> > From my perspective, the main requirement I have is it MUST fully disconnect the user and let him reconnect. Any other solution would just cause more potential headakes.
> 
> So if you maintain the connection it's a non-starter. I've stated this clearly.
>
> *...*
> 
> On Forge's end it's essentially:
> 
> ```
> client.disconnect();
> ConnnectScreen.startConnecting(currentServer, currentPort)
> ```
> 
> The entire point is that you get kicked out to essentially the main menu, so that the world, player instance, and everything associated with it gets destroyed. Thus making sure that the rendering thread or anything else are not using registry data. Because registry data needs to be swapped out.

I've heard you state these requirements before, but in the end, I'm always left with a single question. If it is important that these objects in forge be thrown away so that they may be rebuilt, why don't you throw them away so they can be rebuilt?

In other words, what is keeping the disconnect from being simulated?

electronicboy:
I think that the big general issue is the maintainability of it all, it's much easier to just not deal with flow changes and to just restart the connection vs tryna get stuff to line up later and generally replaying a lot of logic in an unexpected state. Theres a good chunk of state in the server and client which is sync'd up during connection, and to keep the existing connection you'd need to caox the client back into a sane state, prime the client to be able to yeet and resync that data, and be hospitable with the current behavioural expectations of mods

This is generally what we discussed years ago as to how it would be best done, as it generally wouldn't require any special consideration from mods, etc; The only caveat I had is if there should be some raw data that should be exchanged back to the proxy for routing purposes to safely identify individual clients across a "switch", i.e. the proxy would send a message to reconnect to the client with some payload with a client ID, or a signed payload akin to {"target": "magicalserver1"}, ofc, the client *shouldn't* care what that is, IMHO, and should just pass back whatever was sent there

ME1312:
I'll admit that I'm armchair-ing it here but I don't see the harm in faking a disconnect at the "server's" request. From my point of view the flow wouldn't even have to meaningfully change. Is the following just something that can't happen using the existing libraries? If so, why?

0. take note of a request to reset
1. existing disconnect workflow
2. leave the underlying tcp connection open
3. initialize new high-level connection instances & workflow using the existing tcp connection
4. existing connect workflow
5. success!

electronicboy:
you've gotta simulate the entire teardown/up process for this, that flow will need to be maintained to ensure that the correct state resets occur, that the registries are cleaned up properly, and then restart the connection, you've gotta maintain all of the logic for that to occur, vs the logic and the flow which has existed for years and is generally expected by mods, etc

You're just introducing a lot more code that needs to be maintained for a somewhat kinda niche setup

ME1312:
> you've gotta simulate the entire teardown/up process for this, that flow will need to be maintained to ensure that the correct state resets occur, that the registries are cleaned up properly, and then restart the connection, you've gotta maintain all of the logic for that to occur, vs the logic and the flow which has existed for years and is generally expected by mods, etc
> 
> You're just introducing a lot more code that needs to be maintained for a somewhat kinda niche setup

Right &ndash; that would be painful &ndash; so, I'm suggesting that we actually use the existing logic, except without closing the socket. Any protocol, registry, and state tracking would not be carried over to the new connection instance &ndash; just the socket.

electronicboy:
But, that requires changes deep within the connection flow to do that, it's still a chunk of extra code vs disconnect/connect
Theres still caveats of stuff hooking into the networking stack to potentially contend with

ME1312:
It sounds like you're trying to tell me that that's too much of a hack to be considered viable. That's fine if true. I just wanted to throw it out there.

electronicboy:
yea, as I said, it's just a good chunk of extra code for a feature for a niche set of users, the general agreement here between us years ago was K.I.S.S; keeping the existing connection would have been nice, but, it introduces much more complexity than what it solves

danorris709:
> yea, as I said, it's just a good chunk of extra code for a feature for a niche set of users, the general agreement here between us years ago was K.I.S.S; keeping the existing connection would have been nice, but, it introduces much more complexity than what it solves

I think potentially you're over estimating how complicated it truly is. [Here](https://github.com/EnvyWare/MinecraftForge/commits/add-reset-packet) is the implementation and it is really quite simple to do. It leaves the connection open, displays a "negotiating screen" and then re-runs the login sequence for the client. 
In personal testing using our 1.16.5 implementation of Pixelmon (which I believe you'd class as a larger than average mod) there wasn't any significant additional delay for joining the server either

electronicboy:
It's probably not all that much more complex, I never said it was; I just said it was extra complexity and code to maintain for a niche set of the community

LeStegii:
Don't want to annoy anyone, but since there has been no activity since 5 days, do you think that this PR has a chance of getting merged?

johnfriedrich:
> Don't want to annoy anyone, but since there has been no activity since 5 days, do you think that this PR has a chance of getting merged?

From reading the conv it seems like a detailed concept/spec document is still missing. This would help other devs to understand it and maintain it in the future and might help to get that into Forge/Bungee

ME1312:
It seems like "*cookie & reconnect*" is the way forward, so if all parties are okay with that here, I wouldn't mind drafting up some protocol spec for it if that's what we're waiting for.

> This is generally what we discussed years ago as to how it would be best done, as it generally wouldn't require any special consideration from mods, etc; The only caveat I had is if there should be some raw data that should be exchanged back to the proxy for routing purposes to safely identify individual clients across a "switch", i.e. the proxy would send a message to reconnect to the client with some payload with a client ID, or a signed payload akin to {"target": "magicalserver1"}, ofc, the client _shouldn't_ care what that is, IMHO, and should just pass back whatever was sent there

I think this is a good place to start, but I'd like to remind everyone of the extremely rare but non-zero possibility that the proxy instance the user disconnects from might not be the same one they eventually reconnect to (because of load-balancing techniques such as SRV records).

With that in mind, I'm thinking there should be maybe one or two more fields added to allow proxies to determine what *brand* and version of proxy was disconnected from. Thoughts? Additional considerations?

electronicboy:
There has been a small amount of discussion internally regarding this, still waiting to see if some parties respond and such, but, a primative view of how the process should work has been discussed; I've personally mentioned JWT tokens as what I think would be preferred to be communicated, but, also expressed that personally, the client shouldn't care; advantage of the JWT is that, well, they've already solved this specific concern

ME1312:
~Also, what is our stance on adding new packets? I don't really see how we could use [`Login Plugin Response`](https://wiki.vg/Protocol#Login_Plugin_Response) on its own as, given my previous concern, there's no way for the proxy to know which `Message ID` to intercept... unless we define that as a constant, but that's not really how this packet is meant to be interpreted.~

The proper way around that is to just ask forge to state its intentions every time a user connects, but that's not very portable. Though, to be fair, neither is adding packets.

astei:
@ME1312 Adding a new packet is a hard no from my perspective. All the functionality needed for this can be accomplished through login plugin messages.

ME1312:
> @ME1312 Adding a new packet is a hard no from my perspective. All the functionality needed for this can be accomplished through login plugin messages.

2nd paragraph it is then. OK on everything else?

ME1312:
Alright, the draft has been started. Let me know what you think! Also, <ins>please don't quote the entire post</ins> because that would make this thread totally unreadable.

### 4/25/2022
The following is a draft specification for new plugin messaging channel(s) that proxies could use to reconnect forge clients, using [data types and formatting specified by wiki.vg](https://wiki.vg/Protocol#Data_types). It is not yet implementable in its current state as the channel identifier constants have yet to be decided upon.

# [Clientbound] (Play) Plugin Message: Disconnect
This plugin message is sent by the proxy to initiate the server switch process. Upon receiving this, implementing clients are expected to cache the data, initiate a disconnect from the proxy, and reconnect as soon as possible.

<table><thead>
<tr><td><b> Field </b></td><td><b> Type </b></td><td><b> Description </b></td></tr>
</thead><tbody>
<tr><td> Channel </td><td> Identifier </td><td> An agreed upon, constant, namespaced identifier for this plugin message channel. </td></tr>
<tr><td> Data </td><td> Byte Array </td><td><br><table><thead>
<tr><td><b> Field </b></td><td><b> Type </b></td><td><b> Description </b></td></tr>
</thead><tbody>
<tr><td> Brand </td><td> String (32767) </td><td> A constant value that the proxy uses to identify itself. </td></tr>
<tr><td> Data </td><td> Byte Array </td><td> Data that supporting proxies will use to complete the transfer. <br><br> The format is undefined beyond this point, so <br> implementing clients should not attempt to parse the contents. </td></tr>
</tbody></table></td></tr>
</tbody></table>

# [Clientbound] Login Plugin Request: Reconnect
This plugin message is sent by the proxy before connecting to a backend server to determine where to send the user. Implementing proxies are expected to send this to any client for which they detect the FML2 header, but there is no harm in sending it to clients of any type should the need arise.

<table><thead>
<tr><td><b> Field </b></td><td><b> Type </b></td><td><b> Description </b></td></tr>
</thead><tbody>
<tr><td> Message ID </td><td> VarInt </td><td> An ID for which the proxy will await a response. </td></tr>
<tr><td> Channel </td><td> Identifier </td><td> An agreed upon, constant, namespaced identifier for this plugin message channel. </td></tr>
<tr><td> Data </td><td> Byte Array </td><td> A zero-length array must be interpereted as a <code>GET</code> request. <br><br> Otherwise, this is to be interpereted as a <code>RESET</code> request. <code>RESET</code> requests function and <br> contain the same internal data structure as <code>(Play) Plugin Message: Disconnect</code>. </td></tr>
</tbody></table>


# [Serverbound] Login Plugin Response: Connect
This plugin message is sent in response to the above <code>GET</code> request by <ins>all</ins> clients. It is also sent in the event a <code>RESET</code> request fails to disconnect from the proxy.

<table><thead>
<tr><td><b> Field </b></td><td><b> Type </b></td><td><b> Description </b></td></tr>
</thead><tbody>
<tr><td> Message ID </td><td> VarInt </td><td> The ID that was received in the previous plugin message. </td></tr>
<tr><td> Success </td><td> Boolean </td><td> Clients that don't support this specification will output <code>false</code>. <br> Implementing clients that are <ins>not</ins> in the process of reconnecting must output <code>false</code>. <br> Implementing clients that are in the process of reconnecting must output <code>true</code>.</td></tr>
<tr><td> Data </td><td> Byte Array </td><td> This value is only present when the above value is <code>true</code>. <br><br><table><thead>
<tr><td><b> Field </b></td><td><b> Type </b></td><td><b> Description </b></td></tr>
</thead><tbody>
<tr><td> Brand </td><td> String (32767) </td><td> The value the previous proxy identified itself with. <br><br> It is reasonable for proxies to reject the following<br> transfer data when it comes from an unsupported brand.</td></tr>
<tr><td> Data </td><td> Byte Array </td><td> Data the previous proxy supplied to complete the transfer. </td></tr>
</tbody></table></td></tr>
</tbody></table>

Xernium:
Even though I have absolutely no time for this I need to throw this in here; this was my original draft 
Packet design-wise your ideas are in the right direction. But till take more than that, like negotiating a capsulation identifier for vanilla/modified features like brigadier 
D65E63C5-70EC-45FA-A835-712DCC49FEDF

ME1312:
Hmm... well, the outside (right-most and left-most) paths look like what I specified, but I can't say I understand the logic behind the middle bit at all. Also, isn't that the kind of problem namespaced identifiers were introduced to solve?

ME1312:
After thinking it over, my main issue with the above chart really is that 2nd disconnect. We should be able to accomplish this functionality with the single disconnect we're given. Although, I have since updated the draft spec to allow for better exception handling. Also, if we really need extra data to move forward, then let me know.<br><br>

Currently, I believe we should be able to successfully connect forge (and vanilla) users to servers with the following logic:
https://svgshare.com/i/gaa.svg

ME1312:
>> What I don’t like however is how the brigadier types are handled, in my opinion a capsulation should be negotiated so this can work with the existing CrossStitch code.
>
>> Packet design-wise your ideas are in the right direction. But till take more than that, like negotiating a capsulation identifier for vanilla/modified features like brigadier
>
> Isn't that the kind of problem namespaced identifiers were introduced to solve?

Viewing these together clued me into the fact that you're probably trying to tell me that Forge might be modifying the internal data structure of something under a mojang-standard namespace. If that is indeed the case, then my response would be:

Instead of having Forge negotiate this with the proxy (`client - proxy`), Forge should negotiate this with itself (`client - server`). My suggestion for how that should be implemented is:

0. Forge servers should theoretically know ahead-of-time whether or not protocol-breaking mods exist on them, so
1. When protocol-breaking mods are active, move any default vanilla-modified command structures to [`forge:`](https://wiki.vg/Command_Data#forge:modid)
2. Forge clients should be able to accept either identifier for the command structure, naturally

electronicboy:
mods really shouldn't be modifying the built in types, that would just basically be bad practice, namespaces exist for the reason that mods can create their own types without sitting in the vanilla namespace

The proxy needs to modify these data structures in order to do things, i.e. adding commands to the client for things like /server, and thus, we need to be able to decode the structure in some form of manner; The idea of an encapsulation format is basically to wrap said non-standard types so that we don't need to know how to de/encode those specific types, we can just pass them through which will be unwrapped on the other side, akin to what cross stitch does

ME1312:
I feel like we said the same thing, just with different words. Essentially, what I suggested would make it so anything not [`minecraft:` or `brigadier:`](https://wiki.vg/Command_Data) standard could just be forwarded as-is with no further parsing proxy-side.

Xernium:
As @electronicboy said this is where it gets interesting.
## The proxy may be required to intercept the packets sent from downstream OR upstream:
Such packets may include but arent limited to **scoreboards, client settings, resource packs, tablist, respawn** ... etc.
There needs to be an encapsulation support inside these packets if the modded server/client interfere or else the proxy cant function correctly.
### Secondly, your way of interaction with the server completely breaks both Velocity and Bungeecords event-flow.
In a normal scenario:
`Client connect > Login > Login success > Determine server > Send to server`
In your scenario:
`Client connect > Login > Determine server > Send to server > Login success`
Which not only is inconsistent with the flow in future events or server switches but also doesn't account for the reconnections
### Addendum:
What I failed to mention in my chart is that once the client disconnects to initiate a reconnect, the connection should be counted as successful and the ChooseInitialServer sequence would run on Velocity in my scenario.   

That initial server would then be present already for the reconnection, no events prior to the ChooseInitialServerEvent would be fired and the data from the previous connection would be used + the determined server.

**Additionally:** You cannot hold the server or client in pre-login-success for long enough that some of the heavier plugins complete blocking events while choosing a server. Vanilla clients in particular have way too restrictive timeouts for that. It would only work for a modded client that expects to wait.

This way you're not breaking the entire architecture of both proxy and possible plugins.
**We cannot consider this feature if a complete break in architecture and event-flow is the price**

This feature simply isn't important enough for that.

Xernium:
> I feel like we said the same thing, just with different words. Essentially, what I suggested would make it so anything not [`minecraft:` or `brigadier:`](https://wiki.vg/Command_Data) standard could just be forwarded as-is with no further parsing proxy-side.

Ideally yes, but that breaks the API for those on the proxy-side

ME1312:
If the forge team isn't willing to budge on this...
> > From my perspective, the main requirement I have is it MUST fully disconnect the user and let him reconnect. Any other solution would just cause more potential headakes.
> 
> So if you maintain the connection it's a non-starter. I've stated this clearly.


...and proxies aren't willing to budge on this...

> ### Secondly, your way of interaction with the server completely breaks both Velocity and Bungeecords event-flow.
> 
> In a normal scenario: `Client connect > Login > Login success > Determine server > Send to server` In your scenario: `Client connect > Login > Determine server > Send to server > Login success` Which not only is inconsistent with the flow in future events or server switches but also doesn't account for the reconnections

Well then, I'm afraid this is once-again the end of the line, unfortunately. We can't just send Login Plugin Requests/Responses in `Play` state, after all.

Xernium:
> Well then, I'm afraid this is once-again the end of the line, unfortunately. We can't just send Login Plugin Requests/Responses in `Play` state, after all.

That's not what I was trying to get at. It is both possible, this is the issue I created my sequence chart for. The complexity is just a lot higher. 
Essentially: If a verdict on reconnection is reached, the login is marked as "Successful" and an initial server is chosen. The connection is then closed with a reconnect token for the client while its kept as a dummy-connection (waiting) on the proxy.
Upon the reconnection and identification of the client the proxy will re-attach the connection to the waiting connection. From there the connection to the server is established, plugin messages are synced, login is successful, and the server post-connect events are triggered.

For every other client the flow doesnt change.

ME1312:
That could work, yes, but it still seems overly complicated when compared to just treating the reconnect as a reconnect.

Also, thoughts regarding the following breaking a connection's memory-tracking?
> ... but I'd like to remind everyone of the extremely rare but non-zero possibility that the proxy instance the user disconnects from might not be the same one they eventually reconnect to (because of load-balancing techniques such as SRV records).

ME1312:
Also, now that I'm reading this for the 2nd time, I'm not so sure I agree with the statement that my logic has forced a break in event flow. A Login Success event could be sent at the tail end of the first "Normal Login Sequence" instruction as, by that point, you have all the vanilla data &ndash; and any plugins that attempt to send data the client beyond that point could have their data withheld until the client actually enters the `Play` state.

electronicboy:
that's what the reconnect token is for, hence why I said that if the type of data should be determined by spec, it should probs be something akin to a JWT token

TepelStreeltjeEoA:
Hi there, no more updates on this?

Rasgnarok:
For anyone following up on progress here, Pixelmon's added https://github.com/Just-Chaldea/Forge-Client-Reset-Packet/ to the base jar to allow compatibility with this PR, and we will continue to keep the PR up to date with latest Velocity to provide an alternative. We hope that your teams do find a solution for those that have waited since 2018 for it, but until then, Pixelmon will do its best to maintain it for the communities that rely on this beyond 1.13+.

wildmaster84:
Has there been any more progress (or conversations) about a proper way to add modern forge support or is there still conflict between ideas?

Rasgnarok:
> Has there been any more progress (or conversations) about a proper way to add modern forge support or is there still conflict between ideas?

As far as I've seen, no further discussion has happened. This PR is still pending, and Forge <-> Velocity <-> Paper have not publicly set down on a way to approach this issue without resorting to this PR. 

This sadly means you need to rely on community effort to maintain modded proxies until an effort is made across these teams.

electronicboy:
Forge reached out to several proxy Devs under the basis that they wanted to get people to work together to come up with *a* solution, afaik nobody other than us have responded to such a thing and so we're in this general joyous position of this being a low priority for us Vs everything else we need to do, and seemingly nobody else caring

Rasgnarok:
That's simply because you've not community-outreached this. This PR is currently working as expected for several big networks as they've had to go ahead without proxy devs on the matter. You won't see any traction until there is a public, decisive effort from the teams involved to solve it. 

As far as everyone knows, this issue has been pending since 2018 with no progress. It's only natural you don't hear about them, when modded networks have been considered dead to proxies, @electronicboy . 

I've had a _lot_ of people contact me based on my comment above alone. Perhaps the Paper team could write an FAQ or announcement if they want this interest funneled through its channels.

electronicboy:
> That's simply because you've not community-outreached this. This PR is currently working as expected for several big networks as they've had to go ahead without proxy devs on the matter. You won't see any traction until there is a public, decisive effort from the teams involved to solve it.

Forge has outreached to the relevant parties that they want to work with to come up with a spec which they agree with, out of all the parties involved, we're the only group who has cared, and generally even work a spec up, but it's pretty hard to get a community agreed upon consensus of what to do here when the proxy dev community just doesn't care. public traction is irrelevant here as that's not exactly what we're after, unless somebody can actually get md to respond to his emails, that's moot.

> As far as everyone knows, this issue has been pending since 2018 with no progress. It's only natural you don't hear about them, when modded networks have been considered dead to proxies, @electronicboy .

I mean caring on the side of people who can actually get stuff done here, public support doesn't magically make this PR tenable to us. supporting a solution which goes against the interests of actually getting this supported in the core itself would just be backwards, imho.

Rasgnarok:
While I understand your POV, I think you misunderstand what public views would mean for priority setting and agenda-making. Making this a tenable PR starts by knowing just how needed or wanted it is, and I think that's where differences lie. As long as it's not a priority, it won't be further developed either.

electronicboy:
public views wouldn't change my own ability to commit to this, this pr as is is NOT tenable, getting the marching band together is very unlikely to change forges hard line on this PR being untenable

Xernium:
I really don’t care about the politics side of things here.    
The velocity team has come to terms with the fact that we likely won’t get a mod-loader independent implementation of this kind of protocol.    
That said, some of this PR are parts we won’t/can’t accept or budge on, namely:
1. Alteration of the event flow; It breaks api in a way that would require a major new version of velocity. Possible but out of scope currently 
2. Forge specific argument types. We already offer a way to capsulate custom argument types- these have to go.
3. Custom pipeline logic. Same reason as with events
4. Client reliance (too heavily) and the api collisions with existing login api.     
      
If we can work out a solution that gets around all of that, which is hard but not impossible, then the velocity team won’t stand in the way.

danorris709:
Firstly, it's not just about "getting the marching band together". This PR, alongside the Forge clientside mod, works. You, and the Forge team, are yet to provide any form of solution that works despite your constant barrage of complaints where you claim "we could've done this three years ago". It's not about finding the perfect solution, or finding a solution everyone agrees with because ultimately that will never happen.

Aside from the fact that this needs rebasing why is it that the Velocity team has suddenly changed their mind on accepting this PR? As per last contact it was perfectly fine and the only person standing in the way was Lex. I already went over the breaking API requests and fixed those so I find your point mute? And the same stands for custom Forge arguments.

electronicboy:
I'm not aware anywhere where the team agreed on this PR in this form being tenably accepted, as individuals we want forge support but we want an agreed upon manner to do this

as it currently stands, we want a generic means which can be implemented by various parties from a proper specification, this needs to be done in a manner that forge and even fabric would be happy to implement, and as it stands, this is currently not it; I'm closing this, as this PR is not tenable to us, and the back and forth is not going to solve anything here unless somebody actually gets the applicable parties invested into a solution which fits at least appeases the most people

PR #696, opened 2022-04-18
Improved Scheduler API

### LINKED ISSUE #276: Expose scheduler data via API
The scheduler's implementation has information such as the current list of tasks available that is not (but could be) exposed via the API.

Additionally, it would be useful to have a `builder()` method with no args that returns a `TaskBuilder`. That way, all task properties can be set via the `TaskBuilder` which is more consistent than requiring two properties in order to construct the `TaskBuilder` in the first place.

4drian3d:
- Added `Scheduler#buildTask(plugin, Consumer<ScheduledTask>)`
- Added `Scheduler#tasksByPlugin(plugin)`
Allows to obtain the tasks that a plugin has schedule to execute and that are currently active

Resolves #276

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/scheduler/Scheduler.java
+++ b/api/src/main/java/com/velocitypowered/api/scheduler/Scheduler.java
@@ -33,9 +33,10 @@ public interface Scheduler {
    * Initializes a new {@link TaskBuilder} for creating a task on the proxy.
    *
    * @param plugin the plugin to request the task for
+   * @param consumer the task to be run when scheduled with the capacity to cancel itself
    * @return the task builder
    */
-  TaskBuilder builder(@NotNull Object plugin);
+  TaskBuilder buildTask(@NotNull Object plugin, @NotNull Consumer<ScheduledTask> consumer);
 
   /**
    * Get the {@link ScheduledTask} for a specific plugin.
@@ -49,22 +50,6 @@ public interface Scheduler {
    * Represents a fluent interface to schedule tasks on the proxy.
    */
   interface TaskBuilder {
-
-    /**
-     * Specifies the task to be executed.
-     *
-     * @param runnable the task to run
-     * @return this builder, for chaining
-     */
-    TaskBuilder task(@NotNull Runnable runnable);
-
-    /**
-     * Specifies the task to be executed.
-     *
-     * @param consumer the task to be run with the capacity to cancel itself
-     * @return this builder, for chaining
-     */
-    TaskBuilder task(@NotNull Consumer<ScheduledTask> consumer);
 
     /**
      * Specifies that the task should delay its execution by the specified amount of time.
--- a/api/src/main/java/com/velocitypowered/api/scheduler/Scheduler.java
+++ b/api/src/main/java/com/velocitypowered/api/scheduler/Scheduler.java
@@ -8,8 +8,12 @@
 package com.velocitypowered.api.scheduler;
 
 import java.time.Duration;
+import java.util.Collection;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Consumer;
+
 import org.checkerframework.common.value.qual.IntRange;
+import org.jetbrains.annotations.NotNull;
 
 /**
  * Represents a scheduler to execute tasks on the proxy.
@@ -23,12 +27,44 @@ public interface Scheduler {
    * @param runnable the task to run when scheduled
    * @return the task builder
    */
-  TaskBuilder buildTask(Object plugin, Runnable runnable);
+  TaskBuilder buildTask(@NotNull Object plugin, @NotNull Runnable runnable);
+
+  /**
+   * Initializes a new {@link TaskBuilder} for creating a task on the proxy.
+   *
+   * @param plugin the plugin to request the task for
+   * @return the task builder
+   */
+  TaskBuilder builder(@NotNull Object plugin);
+
+  /**
+   * Get the {@link ScheduledTask} for a specific plugin.
+   *
+   * @param plugin the plugin object
+   * @return the list of {@link ScheduledTask} corresponding to a specific plugin
+   */
+  @NotNull Collection<ScheduledTask> tasksByPlugin(@NotNull Object plugin);
 
   /**
    * Represents a fluent interface to schedule tasks on the proxy.
    */
   interface TaskBuilder {
+
+    /**
+     * Specifies the task to be executed.
+     *
+     * @param runnable the task to run
+     * @return this builder, for chaining
+     */
+    TaskBuilder task(@NotNull Runnable runnable);
+
+    /**
+     * Specifies the task to be executed.
+     *
+     * @param consumer the task to be run with the capacity to cancel itself
+     * @return this builder, for chaining
+     */
+    TaskBuilder task(@NotNull Consumer<ScheduledTask> consumer);
 
     /**
      * Specifies that the task should delay its execution by the specified amount of time.
@@ -37,7 +73,7 @@ public interface Scheduler {
      * @param unit the unit of time for {@code time}
      * @return this builder, for chaining
      */
-    TaskBuilder delay(@IntRange(from = 0) long time, TimeUnit unit);
+    TaskBuilder delay(@IntRange(from = 0) long time, @NotNull TimeUnit unit);
 
     /**
      * Specifies that the task should delay its execution by the specified amount of time.
@@ -45,7 +81,7 @@ public interface Scheduler {
      * @param duration the duration of the delay
      * @return this builder, for chaining
      */
-    default TaskBuilder delay(Duration duration) {
+    default TaskBuilder delay(@NotNull Duration duration) {
       return delay(duration.toMillis(), TimeUnit.MILLISECONDS);
     }
 
@@ -57,7 +93,7 @@ public interface Scheduler {
      * @param unit the unit of time for {@code time}
      * @return this builder, for chaining
      */
-    TaskBuilder repeat(@IntRange(from = 0) long time, TimeUnit unit);
+    TaskBuilder repeat(@IntRange(from = 0) long time, @NotNull TimeUnit unit);
 
     /**
      * Specifies that the task should continue running after waiting for the specified amount, until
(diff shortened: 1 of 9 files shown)


proxy/src/main/java/com/velocitypowered/proxy/scheduler/VelocityScheduler.java
```diff
@@ -65,12 +68,28 @@ public VelocityScheduler(PluginManager pluginManager) {
...
+    checkArgument(pluginManager.fromInstance(plugin).isPresent(), "plugin is not registered");
+    return new TaskBuilderImpl(plugin, null);
+  }
+
+  @Override
+  public @NonNull Collection<ScheduledTask> tasksByPlugin(@NonNull Object plugin) {
+    requireNonNull(plugin, () -> "the plugin cannot be null");
+    checkArgument(pluginManager.fromInstance(plugin).isPresent(), "plugin is not registered");
+    synchronized (tasksByPlugin) {
+      return Set.copyOf(tasksByPlugin.get(plugin));
```
hugmanrique: Nit: the `get` call can be moved outside of the synchronized block (https://javadoc.io/doc/com.google.guava/guava/latest/com/google/common/collect/Multimaps.html#synchronizedMultimap(com.google.common.collect.Multimap) (the synchronized block would then contain `Set.copyOf(tasks)`.

proxy/src/test/java/com/velocitypowered/proxy/scheduler/VelocitySchedulerTest.java
```diff
@@ -65,4 +66,39 @@ void repeatTaskWorks() throws Exception {
...
+        .schedule());
+  }
+
+  @Test
+  void obtainTasksFromPlugin() throws Exception {
+    VelocityScheduler scheduler = new VelocityScheduler(new FakePluginManager());
+
+    scheduler.builder(FakePluginManager.PLUGIN_A)
+      .task(ScheduledTask::cancel)
+      .delay(100, TimeUnit.MILLISECONDS)
```
hugmanrique: This test is time-dependent, and could fail when running with a single core. Could we perform the assertion inside the task itself (and so the task no longer needs to be delayed)?

proxy/src/test/java/com/velocitypowered/proxy/scheduler/VelocitySchedulerTest.java
```diff
@@ -65,4 +66,39 @@ void repeatTaskWorks() throws Exception {
...
+
+    assertEquals(scheduler.tasksByPlugin(FakePluginManager.PLUGIN_A).size(), 1);
+  }
+
+  @Test
+  void testConsumerCancel() throws Exception {
+    VelocityScheduler scheduler = new VelocityScheduler(new FakePluginManager());
+
+    ScheduledTask task = scheduler.builder(FakePluginManager.PLUGIN_B)
+        .task(ScheduledTask::cancel)
```
hugmanrique: Same as above: we could schedule a task inside the original task that checks its status.

api/src/main/java/com/velocitypowered/api/scheduler/Scheduler.java
```diff
@@ -23,29 +27,61 @@ public interface Scheduler {
...
-  TaskBuilder buildTask(Object plugin, Runnable runnable);
+  TaskBuilder buildTask(@NotNull Object plugin, @NotNull Runnable runnable);
+
+  /**
+   * Initializes a new {@link TaskBuilder} for creating a task on the proxy.
+   *
+   * @param plugin the plugin to request the task for
+   * @return the task builder
+   */
+  TaskBuilder builder(@NotNull Object plugin);
```
hugmanrique: In the original issue, @astei argued that methods to create a builder that take all mandatory arguments are more idiomatic. Could we use the same approach and replace this by a `buildTask(Object, Consumer<ScheduledTask>)` method?

I rarely see the need for the `TaskBuilder#task` methods below, but I guess it makes the builders reusable (this should be documented). On the other hand, e.g. once a `Runnable` task has been specified, the `Consumer` variant can no longer be called and vice versa. This is a bit inconsistent. An alternative is for the `task` methods to clear the other field (e.g. `TaskBuilder#task(Runnable)` sets the `Consumer` field to `null`). That way we can also remove the precondition in the `schedule` method (it confused me at first).

Somewhat related: I would change the conditions of the form `runnable == null ? use consumer : use runnable` to `runnable != null ? use runnable : use consumer` to prevent bugs in the future.

4drian3d: Now that you mention it, I would have liked the idea that the builder could be reusable, but I find its usage very rare and somewhat confusing

proxy/src/test/java/com/velocitypowered/proxy/scheduler/VelocitySchedulerTest.java
4drian3d: _With this specific test I doubt very much that it could fail, since I doubt that going from one simple method to another would take 100 milliseconds_

I just understood what you mean with the following comment

proxy/src/main/java/com/velocitypowered/proxy/scheduler/VelocityScheduler.java
```diff
@@ -65,10 +68,28 @@ public VelocityScheduler(PluginManager pluginManager) {
 
   @Override
   public TaskBuilder buildTask(Object plugin, Runnable runnable) {
-    checkNotNull(plugin, "plugin");
-    checkNotNull(runnable, "runnable");
+    requireNonNull(runnable, () -> "the scheduled task cannot be null");
```
astei: why change this?

proxy/src/test/java/com/velocitypowered/proxy/scheduler/VelocitySchedulerTest.java
```diff
@@ -65,4 +67,36 @@ void repeatTaskWorks() throws Exception {
     task.cancel();
   }
 
+  @Test
+  void obtainTasksFromPlugin() throws Exception {
+    VelocityScheduler scheduler = new VelocityScheduler(new FakePluginManager());
```
astei, line 74: I'm nervous about this test. How do we know it will pass or fail? Would feel more comfortable with a `CountDownLock` and doing an assertion in the test thread.

proxy/src/test/java/com/velocitypowered/proxy/scheduler/VelocitySchedulerTest.java
```diff
@@ -65,4 +67,36 @@ void repeatTaskWorks() throws Exception {
...
+        ).schedule();
+      }
+    }).delay(50, TimeUnit.MILLISECONDS)
+      .repeat(Duration.ofMillis(5))
+      .schedule();
+  }
+
+  @Test
+  void testConsumerCancel() throws Exception {
+    VelocityScheduler scheduler = new VelocityScheduler(new FakePluginManager());
```
astei, line 96: ditto

astei (changes requested):

proxy/src/test/java/com/velocitypowered/proxy/scheduler/VelocitySchedulerTest.java
```diff
@@ -65,4 +67,36 @@ void repeatTaskWorks() throws Exception {
...
+
+    scheduler.buildTask(FakePluginManager.PLUGIN_B, actualTask -> {
```
astei: I'd also have to have a condition to check if `actualTask` is the same value as what `TaskBuilder#scheduler()` will return.

astei (changes requested):

proxy/src/main/java/com/velocitypowered/proxy/scheduler/VelocityScheduler.java
4drian3d: I thought it would be better to use a Supplier to avoid creating an unnecessary string each time this method is called and no argument is null

proxy/src/test/java/com/velocitypowered/proxy/scheduler/VelocitySchedulerTest.java
4drian3d: If you mean a `CountDownLatch`, I just implemented it in the last commit

proxy/src/main/java/com/velocitypowered/proxy/scheduler/VelocityScheduler.java
astei: The string is stored in the constant pool. There is no need to "create an unnecessary string".

proxy/src/test/java/com/velocitypowered/proxy/scheduler/VelocitySchedulerTest.java
```diff
@@ -65,4 +67,75 @@ void repeatTaskWorks() throws Exception {
...
+    }).repeat(5, TimeUnit.MILLISECONDS).schedule();
+    latch.await();
+  }
+
+  @Test
+  void testConsumerEquality() throws Exception {
+    VelocityScheduler scheduler = new VelocityScheduler(new FakePluginManager());
+    CountDownLatch latch = new CountDownLatch(1);
+
+    TaskContainer consumerTask = new TaskContainer();
```
astei: Why not use `AtomicReference<ScheduledTask>`, which is safer?

proxy/src/test/java/com/velocitypowered/proxy/scheduler/VelocitySchedulerTest.java
astei: This is not acceptable. We should always assert OUTSIDE of any tasks.

proxy/src/test/java/com/velocitypowered/proxy/scheduler/VelocitySchedulerTest.java
astei: This is not acceptable. We should NEVER be asserting inside a task.

astei (changes requested):

astei (approved):

PR #697, opened 2022-04-18
Implement custom Modern Forwarding protocol limit

### LINKED ISSUE #602: Add allowed client version range as a configuration option
When running a network that only accepts the latest versions (or at the very least, not the full range of 1.8-1.18), it would be great if there was a configuration option that defined the client versions that are allowed to connect. If you connect with a version not allowed, it would disconnect you with a message showing "This server is only compatible with Minecraft versions 1.17.1-1.18" (or whatever version range you defined). This is already partially possible with a plugin, but it becomes a bit messy if you use modern forwarding because [Velocity drops connections from clients older than 1.13 before plugins get a chance to change the message](https://github.com/PaperMC/Velocity/blob/292b2f96ab8d65a9982dd787c626d1060e5472be/proxy/src/main/java/com/velocitypowered/proxy/connection/client/HandshakeSessionHandler.java#L139).

4drian3d:
Added a configurable limit for the lowest protocol acceptable by velocity in case Modern Forwarding mode is active

Currently it is impossible to control the limit of versions that can enter through Velocity in the modern forwarding mode, this option allows, besides changing the disconnection message, to add a limit of protocols higher than 1.13 which will also be disconnected in case they are lower than the established protocol (protocols lower than 1.13 will still be kicked out)

Resolves #602


electronicboy:
sounds like this PR is tryna catch the wrong issue; original PR looks to be tryna ask more for a custom connection range, vs, well, this. Not sure that this is solving the actual issue they want, nor am I sure if this really belongs in velocity core

Xernium:
Since this version limit can be achieved by using the API we will not consider this, it makes no sense in a normal server environment and only caters to a very specific use that I still fail to understand.

PR #707, opened 2022-05-12
Clarify the invalid ID message.

kyngs:
The current message is confusing for many new people. TBH it even confused me when getting started with velocity.

On a bit side note, what was the reason to have the id lowercase only?


kennytv:
A generally better phrasing would be something along the lines of
`Ids must be lowercase, start with a letter, and otherwise only contain alphanumerical characters, dashes and underscores`.

Having the actual id (vs. a name) be lowercase means you don't have to worry about capitalization when depending on other plugins or start getting into weird shenangigans when two plugins have the same name but differently capitalized

hugmanrique:
Someone today asked what ID the message was referring to. Perhaps we could log that as well, or note that it should be changed in the `@Plugin` annotation.

PR #712, opened 2022-05-26
Add config support for using file as forwarding secret

### LINKED ISSUE #463: Save forwarding secrets in a different file from the regular proxy config
This will provide a major improvement for proxy security.

logsym:
Resolves #463 by enabling the user to configure Velocity to use a file for the forwarding secret. The configuration parameter is `forwarding-secret-file` and takes a path to the file to use. Currently, the configuration parameter is not included by default and requires the user to manually add it. 

In interest of backwards-compatibility, the order at which Velocity chooses where to source the secret from is the following:
1. Environment Variable
2. `forwarding-secret` config parameter
3. `forwarding-secret-file` config parameter

In order to use the `forwarding-secret-file` config parameter, the environment variable and `forwarding-secret` config parameter must not be set.

Considerations:
1. Should the configuration include the `forwarding-secret-file` parameter by default so the user is aware it exists without having to go online to reference?
2. Should there be some sort of notification to let the user know which of these three values Velocity is sourcing from? I can see potential headaches of users wondering why their desired secret isn't be used.
3. When the secret file is read to string it contains the EOF character(s); unsure if that is of concern or not.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
@@ -415,6 +415,12 @@ public class VelocityConfiguration implements ProxyConfig {
       throw new RuntimeException("Default configuration file does not exist.");
     }
 
+    // Create the forwarding-secret file on first-time startup if it doesn't exist
+    Path defaultForwardingSecretPath = Path.of("forwarding.secret");
+    if (!path.toFile().exists() && !defaultForwardingSecretPath.toFile().exists()) {
+      Files.writeString(defaultForwardingSecretPath, generateRandomString(12));
+    }
+
     boolean mustResave = false;
     CommentedFileConfig config = CommentedFileConfig.builder(path)
         .defaultData(defaultConfigLocation)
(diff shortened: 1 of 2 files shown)


hugmanrique:
I think the original issue envisioned moving the key option from the `velocity.toml` file to another file altogether. This change introduces complexity, and its non-default nature defeats the main purpose: prevent inexperienced users from sharing their key online. 

As it stands, a simple `VELOCITY_FORWARDING_SECRET=$(cat secret.txt) java -jar ...` can emulate the current diff.

logsym:
> I think the original issue envisioned moving the key option from the `velocity.toml` file to another file altogether. This change introduces complexity, and its non-default nature defeats the main purpose: prevent inexperienced users from sharing their key online.
> 
> As it stands, a simple `VELOCITY_FORWARDING_SECRET=$(cat secret.txt) java -jar ...` can emulate the current diff.

Consider how many server applications allow the user to enter paths for secrets and keys directly in the main configuration file. Think Nginx, Apache Server, and SSH. Need to share the config file with others due to issues? No problem, it only includes the file names.

I feel like the config should only be the config. No secrets or sensitive information (as best as can be managed). Adding an entirely new configuration file for more secrets still seems like bad practice to me. Secrets are for the application, not for the configuration.

You are correct in that setting the variable using cat works just as well as this diff. But it's more of a hack than first-class support. It might not be so simple in different environments as well. Using environment variables rather than files isn't best security practice either. See [https://diogomonica.com/2017/03/27/why-you-shouldnt-use-env-variables-for-secret-data/](https://diogomonica.com/2017/03/27/why-you-shouldnt-use-env-variables-for-secret-data/)

The PR can be amended like so maybe:

1. Inform the user that using `forwarding-secret` in the config is deprecated, insecure, and they should prefer using a file or environment variable.
2. By default add the `forwarding-secret-file` config option and set it to something like `"secret.txt"`, create that file in the working directory, and add some random text much like Velocity currently does.
3. Edit: Could also completely remove `forwarding-secret` from the config and automatically move the secret to a file. This wouldn't allow users time to transition and be aware of the change and doesn't seem near as elegant as the above mentioned solution.

Appreciate the feedback. Please let me know if you have any more thoughts. :)

hugmanrique:
I agree with your first three paragraphs: I was also proposing adding a `forwarding-key.txt`-esque file, not a new config file (still, do we need the secret file path to be configurable?).

I think (1) and (2) are a good compromise for now :) The config entry could be dropped in a future major version.

logsym:
> I agree with your first three paragraphs: I was also proposing adding a `forwarding-key.txt`-esque file, not a new config file (still, do we need the secret file path to be configurable?).
> 
> I think (1) and (2) are a good compromise for now :) The config entry could be dropped in a future major version.

Ohhh okay. I wasn't entirely sure what you meant there. Glad we're on the same page now :)

I think in the interest of being flexible to the user, they should be able to specify a full path for the file. In Docker for example, secrets are mounted in `/run/secrets/<secret_name>` by default. The implementation complexity wouldn't change significantly to enable anyways.

logsym:
Tweaks since first commit:

1. A notification warning the user to migrate to using `forwarding-secret-file` instead of `forwarding-secret`.
2. Velocity will create a `forwarding.secret` default file with a 12 character random string (much like before) if the `forwarding-secret-file` parameter isn't present or empty.
3. The `default-velocity.toml` configuration file has been updated to reflect the new preference for the separate secret file.

proxy/src/main/resources/default-velocity.toml
```diff
@@ -33,8 +33,8 @@ prevent-client-proxy-connections = false
 #                  Velocity's native forwarding. Only applicable for Minecraft 1.13 or higher.
 player-info-forwarding-mode = "NONE"
 
-# If you are using modern or BungeeGuard IP forwarding, configure a unique secret here.
-forwarding-secret = ""
+# If you are using modern or BungeeGuard IP forwarding, configure a file that contains a unique secret here.
+forwarding-secret-file = ""
```
hugmanrique: Nit: could we default this to `forwarding.secret` as in `VelocityConfiguration`? This will save the extra re-save during the initial start-up.

proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
```diff
@@ -435,16 +435,34 @@ public static VelocityConfiguration read(Path path) throws IOException {
...
     byte[] forwardingSecret;
     String forwardingSecretString = System.getenv()
         .getOrDefault("VELOCITY_FORWARDING_SECRET", config.get("forwarding-secret"));
-    if (forwardingSecretString == null || forwardingSecretString.isEmpty()) {
-      forwardingSecretString = generateRandomString(12);
-      config.set("forwarding-secret", forwardingSecretString);
+
+    // Ensure that the "forwarding-secret-file" configuration parameter and auto-created file are present.
+    String forwardingSecretFileName = config.get("forwarding-secret-file");
+    if (forwardingSecretFileName == null || forwardingSecretFileName.isEmpty()) {
```
hugmanrique: If the config contains the `forwarding-secret` option, we shouldn't create the secret file: its contents will not be used and the user could be confused because the config and the file contents are different. Instead, we can move this down to line 462.

proxy/src/main/resources/default-velocity.toml
logsym: > Nit: could we default this to `forwarding.secret` as in `VelocityConfiguration`? This will save the extra re-save during the initial start-up.

That makes sense. Though, we'd have to change the logic that generates the `forwarding.secret` file; unless we depend on the user to manage that?

Currently it checks if the `forwarding-secret-file` parameter is null or empty, and if it is it'll create the file. If it will default to be present and not empty, the current logic will not work and the user will be expected to create the file with the secret contents.

I feel depending on the user to create the file isn't such a bad idea (in terms of security) though. It gives them the option of choosing where to put the file and what file permissions (chmod/chown) to give it? There will be an extra step though and maybe that won't go over well for the majority of users?

proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
logsym: > If the config contains the `forwarding-secret` option, we shouldn't create the secret file: its contents will not be used and the user could be confused because the config and the file contents are different. Instead, we can move this down to line 462.

That would make sense as well. Though we need to iron out the file creation details (mentioned above) before that logic is moved around.

logsym:
Here's a summary of the latest commit:

1. Change default config version to `2.0`, allowing us to handle different versions.
2. There is now a check to determine if the config version is `1.0` and if it is, it handles the `forwarding-secret` parameter just like you would expect pre-PR. Otherwise, it will use only the `forwarding-secret-file` parameter as expected in version `2.0`.
3. The file reading has been changed to ensure new line characters are not included in the secret (to maintain consistency with Paper and be at peace with VS Code and its add new-line character when saving feature). Please comment if you feel this shouldn't be default behavior; I was a little unsure about this change.
4. Version `2.0` doesn't modify any values in the config file at all. There is no need to resave (currently).
5. The user is expected to upgrade the config file themselves. There isn't a huge push beyond the console warning, and upgrading really is trivial. I felt like trying to programmatically update the old config was too convoluted and prone to future headaches. Having the user update themselves I feel will also help them understand the changes better.
6. A runtime exception will occur if the secret file is empty and Velocity is actively using it. It's not secure so the user should be informed.
7. Velocity generates a secret file automatically when it is first started (meaning there is no `velocity.toml` and `forwarding.secret` files present yet).

astei (approved):
I'm happy with this

PR #716, opened 2022-06-01
Clean up and Updates

JT122406:


KoxSosen (changes requested):
Velocity follows the [Google Code Style](https://google.github.io/styleguide/javaguide.html) for Java, see the contributing.md file. 

It seems like many of these changes don't exactly follow that, for as an example wildcard imports.

Also, did you test these changes?

4drian3d:
You can't update dependencies without reason, especially if it is between major versions. I'm sure many people use Configurate 3 and Guava in their plugins and this change may affect compatibility with these plugins, being that the next version will not be a major version that has to break compatibility.
Ideally Velocity would like to upgrade to Configurate 4, but it is not currently the time to do so.

Also, I'm sure you didn't even try to compile the project since several of your changes make it impossible to compile Velocity (mostly because of the checkstyle)

electronicboy:
Updating dependencies is something which needs to be done within reason; i.e. making sure that they don't break ABI or are handled on a release which isn't a late in cycle snapshot build

wildcard imports are evil and generally, this doesn't even compile because you broke the checkstyle rules

as a side note, NEVER submit a PR from the same branch as you're pushing to

xism4:
> Velocity follows the [Google Code Style](https://google.github.io/styleguide/javaguide.html) for Java, see the contributing.md file.
> 
> It seems like many of these changes don't exactly follow that, for as an example wildcard imports.
> 
> Also, did you test these changes?

Basicly he's IDE did that refactor for him. He just upgraded versions because Jetbrains IDE said it.

kyngs:
> > Velocity follows the [Google Code Style](https://google.github.io/styleguide/javaguide.html) for Java, see the contributing.md file.
> > It seems like many of these changes don't exactly follow that, for as an example wildcard imports.
> > Also, did you test these changes?
> 
> Basicly he's IDE did that refactor for him. He just upgraded versions because Jetbrains IDE said it.

Yep, you should install the appropriate extension when working with Velocity.

PR #717, opened 2022-06-01
Bump Bstats to latest 3.0.0

JT122406:


4drian3d:
Between bStats version 2.2.1 and 3.0.0 only the code in the sponge module has changed, this change would not really have any effect on Velocity

JT122406:
> Between bStats version 2.2.1 and 3.0.0 only the code in the sponge module has changed, this change would not really have any effect on Velocity

Yeah so bumping it should be no big deal

PR #718, opened 2022-06-02
Implement ComponentLogger

4drian3d:
superseded by https://github.com/PaperMC/Velocity/pull/1022


xIsm4:
> _ No description provided. _

Me guen the when, epic

4drian3d:
Tested, works correctly 
component-logger

4drian3d:
I have been trying for quite some time to implement a system similar to the one Paper has with a custom LogEventPatternConverter, but due to https://issues.apache.org/jira/browse/LOG4J2-673 it is impossible at the moment. Paper uses a non-"conventional" method to create its builds, so it is not affected

electronicboy:
I can't look right now, but, iirc there is a transformer somewhere for
merging the plugin config things, in the past we used to just exclude that
file which sends l4j back to a class path scan

PR #721, opened 2022-06-04
Implemented customizable translation formatting

### LINKED ISSUE #720: HEX/RGB color codes in "com\velocitypowered\proxy\l10n"
I tried everything to call RGB colors but failed so I think other than basic minecraft colors are not supported.

So i want to make a suggestion because of the RGB colors.

My whole server would be made up of RGB colors, and it would be embarrassing if my proxy server weren't exactly that.
(velocity.jar\com\velocitypowered\proxy\l10n)

4drian3d:
Added ability to customize the translation formatting

img

resolves #720


proxy/src/main/java/com/velocitypowered/proxy/util/ClosestLocaleMatcher.java
```diff
@@ -54,4 +61,43 @@ public void registerKnown(final Locale locale) {
...
+        ))
+        .format(format(arguments), new StringBuffer(), null).toString())
+      .colorIfAbsent(color);
+  }
+
+  private static final Object[] format(Object... arguments) {
+    if (arguments == null || arguments.length == 0) {
+      return arguments;
+    }
+    for (int i = 0; i <= 0; i++) {
```
Joo200: This does only work when you're providing exactly one argument.
Your ArrayIndexOutOfBoundException is related to that: The correct iteration for an array is `for (int i = 0; i < array.length; i++)`

4drian3d: Thanks for commenting, I really didn't notice that mistake, thank you very much

Joo200: Now you'll receive every time a ArrayIndexOutOfBoundException...

4drian3d: I made the change without seeing it much, thanks again

4drian3d:
superseded by #1108

PR #727, opened 2022-06-08
Hotfix/patch modern forwarding v2

CoreyShupe:


hugmanrique:
Both branches return the same value, is this intended?

PR #730, opened 2022-06-08
Fix unsigned chat breaking decode on PlayerChat.

CoreyShupe:


codeHusky:
Appears to have fixed this particular issue I was facing, leaving here in case anyone else was running into a similar bug.
image

CoreyShupe:
> Appears to have fixed this particular issue I was facing, leaving here in case anyone else was running into a similar bug. image

that bug was fixed here: https://github.com/PaperMC/Velocity/pull/729

PR #731, opened 2022-06-08
Don't send "shrunk" removed from 1.16.2

LemonCaramel:
This data is not used in Minecraft 1.16.2 and later versions


LemonCaramel:
This PR is no longer need since commit 93b6376

PR #732, opened 2022-06-08
Fix system messages shown in action bar for 1.19 clients

Foorcee:
CHAT type is not supported for system messages.


electronicboy (dismissed):

electronicboy:
Actually looking at this, I'd need to have a deeper look, but, this doesn't seem proper? That chat types seem to use the same registry, so, idk why -1 would fix anything here?

Gerrygames:
I think -1 works for <=1.18.2 servers with ViaVersion. 
https://github.com/ViaVersion/ViaVersion/blob/7ab0c41be6fcf0c3cea0231a20a46973975bedf8/common/src/main/java/com/viaversion/viaversion/protocols/protocol1_19to1_18_2/packets/EntityPackets.java#L52

The proper fix would be to look up the correct id in the dimension registry sent in the join game packet.

electronicboy:
Ah, yea, so, this PR is actually only technically correct on Via due to that hard coding, in reality, MC shares the registry for the two, and so, chat is a valid registered thing in there

Issue #150, opened 2018-11-25
exception when connecting to proxy


BukkitTNT:
Velocity throws an exception while connecting to paperspigot 1.13.2 server via "modern" mode.

Exception: https://gist.github.com/BukkitTNT/66016fe445e9f2a96753cafc574c8bdc

Paperspigot isn't recognising a new connection.
Occured after reconnecting the proxy.

astei:
Did you enable the Velocity forwarding support in `paper.yml`?

BukkitTNT:
Yes

astei:
Paste your `velocity.toml`, your `paper.yml`, your `spigot.yml`, and any logs you have.

BukkitTNT:
https://gist.github.com/BukkitTNT/20c51dcd0880488bd739f851debafd25

astei:
I am unable to reproduce this issue, using the latest version of Velocity from master (edfdcf3) and the latest build of Paper (at this time, the latest 1.13.x version is build 463).

hugmanrique:
You have `bungeecord` set to `false` in your `spigot.yml` config. I believe it should be set to true (https://docs.velocitypowered.com/en/latest/users/player-info-forwarding.html#spigot-paper )

astei:
@hugmanrique Nope, that's for legacy forwarding. They want to use modern forwarding.

PR #736, opened 2022-06-08
move force key checking into a system property

electronicboy:
Primarily added to allow plugins to override this option for their own "gains", from a quick discussion with Five

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
@@ -315,6 +315,7 @@ public class VelocityServer implements ProxyServer, ForwardingAudience {
       }
 
       commandManager.setAnnounceProxyCommands(configuration.isAnnounceProxyCommands());
+      System.setProperty("auth.forceSecureProfiles", String.valueOf(configuration.isForceKeyAuthentication()));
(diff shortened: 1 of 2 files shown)


Iseason2000:
There is a problem at line 78 [link](https://github.com/PaperMC/Velocity/blob/df20d81f0f98f2b4704bc4c8a5750dabf48a301d/proxy/src/main/java/com/velocitypowered/proxy/connection/client/InitialLoginSessionHandler.java#L78) that the value of
Boolean.getBoolean("auth.forceSecureProfiles") is always false
which was broke geyser

electronicboy:
works perfectly fine here, no idea how it being always false would even break geyser

Iseason2000:
sorry i misread the java api  ,but when i use the version of velocity later than #150 whit geyser , i got a error message "Missing profile public key.This server requires secure profiles."  when join the server with bedrock version
That's why I've come to this conclusion

electronicboy:
Which would make that value true, not false; that issueshould have also been fixed in geyser/floodgate or whatever by now

Camotoy:
Floodgate works around key authentication in Velocity: https://github.com/GeyserMC/Floodgate/commit/59b37f20e40391a878ca870c394d612fbcf66298
Make sure you're using the latest Floodgate build.

PR #737, opened 2022-06-09
Enable modern forwarding v2

Xernium:
This adds a (byte) qualifier the server can provide to indicate the highest supported modern forwarding mode.
If not provided the proxy will always assume v1.
The proxy may respond with any version up to and equal to the requested version depending on the data present.
This is done to keep compatibility with servers that have their protocol translated by a Plugin like ViaVersion.


electronicboy (approved):

electronicboy:
fwiw, tested that this works as expected on paper, so, all gucci for me

PR #743, opened 2022-06-12
Fixed possible IOException if forwarding-secret-file path is a directory

### LINKED ISSUE #740: Can't start Velocity 3.1.2-SNAPSHOT (git-74edac96-b153)
https://pastebin.com/a0EcSqdM

4drian3d:
Actually an IOException will occur in some cases where the config-version option does not exist, is not exactly "1.0", among other reasons.
This change will give an informational error in case the forwarding-secret-file is not a valid file, does not exist or is a directory, or in case the config-version does not exist or is not exactly "1.0" as for example if an alternative version has been tested that changes that previous value (i.e #666 )

resolves #740

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
@@ -424,7 +424,7 @@ public class VelocityConfiguration implements ProxyConfig {
 
     // Create the forwarding-secret file on first-time startup if it doesn't exist
     Path defaultForwardingSecretPath = Path.of("forwarding.secret");
-    if (!Files.exists(path) && !Files.exists(defaultForwardingSecretPath)) {
+    if (Files.notExists(path) && Files.notExists(defaultForwardingSecretPath)) {
       Files.writeString(defaultForwardingSecretPath, generateRandomString(12));
     }
 
@@ -448,8 +448,16 @@ public class VelocityConfiguration implements ProxyConfig {
     CommentedFileConfig defaultConfig = CommentedFileConfig.of(tmpFile, TomlFormat.instance());
     defaultConfig.load();
 
+    // TODO: migrate this on Velocity Polymer
+    double configVersion;
+    try {
+      configVersion = Double.parseDouble(config.getOrElse("config-version", "1.0"));
+    } catch (NumberFormatException e) {
+      configVersion = 1.0;
+    }
+
     // Whether or not this config version is older than 2.0 which uses the deprecated "forwarding-secret" parameter
-    boolean legacyConfig = Double.parseDouble(config.getOrElse("config-version", "1.0")) < 2.0;
+    boolean legacyConfig = configVersion < 2.0;
 
     String forwardingSecretString;
     byte[] forwardingSecret;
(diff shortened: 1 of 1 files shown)


logsym (approved):
Looks good to me! Really appreciate you tweaking this.

proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
```diff
@@ -448,8 +448,8 @@ public static VelocityConfiguration read(Path path) throws IOException {
     CommentedFileConfig defaultConfig = CommentedFileConfig.of(tmpFile, TomlFormat.instance());
     defaultConfig.load();
 
-    // Whether or not this config is version 1.0 which uses the deprecated "forwarding-secret" parameter
-    boolean legacyConfig = config.getOrElse("config-version", "").equalsIgnoreCase("1.0");
+    // Whether or not this config version is older than 2.0 which uses the deprecated "forwarding-secret" parameter
+    boolean legacyConfig = Double.parseDouble(config.getOrElse("config-version", "1.0")) < 2.0;
```
hugmanrique: Nit: a version is not guaranteed to follow the `major.minor` format. The Polymer branch included the semver4j dependency to perform plugin dependency matching (see [diff](https://github.com/PaperMC/Velocity/commit/2a39ddb03e5a25dafaa3949859e2860ee3650656#diff-f091b0b868fccf7d828aa915340bf2d65c580d71bd5d23e0af8f44fcd1a231c1R138)). While this will work for now, I would add a `// todo` annotation.

proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
```diff
@@ -424,7 +424,7 @@ public static VelocityConfiguration read(Path path) throws IOException {
 
     // Create the forwarding-secret file on first-time startup if it doesn't exist
     Path defaultForwardingSecretPath = Path.of("forwarding.secret");
-    if (!path.toFile().exists() && !defaultForwardingSecretPath.toFile().exists()) {
+    if (!Files.exists(path) && !Files.exists(defaultForwardingSecretPath)) {
```
hugmanrique: Caution: `!Files.exists` is not equivalent to [`Files.notExists`](https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/nio/file/Files.html#notExists(java.nio.file.Path,java.nio.file.LinkOption...)) when the existence cannot be determined.

4drian3d: That's why I used the !Files.exists in many projects because I read its documentation, I think that in this case it is worth to put Files.notExists (now that I think about it, maybe I misunderstood the documentation the first time I read it)

proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
4drian3d: Currently the config-version option accepts a String, in case it is not a double, the version will be set to 1.0. I also put a TODO to change the logic in Polymer

astei (approved):

PR #751, opened 2022-06-17
Improve plugin id and dependency id handling

jacoballen1:


api/src/ap/java/com/velocitypowered/api/plugin/ap/SerializedPluginDescription.java
```diff
@@ -14,14 +14,15 @@
...
+import java.util.Locale;
 import java.util.Objects;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import org.checkerframework.checker.nullness.qual.Nullable;
 
 public final class SerializedPluginDescription {
 
-  public static final Pattern ID_PATTERN = Pattern.compile("[a-z][a-z0-9-_]{0,63}");
+  public static final Pattern ID_PATTERN = Pattern.compile("[a-z-A-Z][a-z-A-Z-0-9-_]{0,63}");
```
hugmanrique: This regex contains extra dashes between the z and the A (e.g. "-foo" is incorrectly accepted as a valid ID).

hugmanrique (changes requested):
Otherwise lgtm

hugmanrique:
Sorry, also missed the extra dash in `Z-0`.

PR #771, opened 2022-06-23
Implement the ServerData packet by firing ProxyPingEvent

astei:
Mojang introduced the enable-status server property with Minecraft 1.19, which if enabled causes servers to close the connection when a client tries to ping them. Mojang wants to show the MOTD and favicon on the server select screen for those who manage to log in, so we need to implement this packet as well.

The good news is that we can send this packet as many times as needed on the same connection

This matches the behavior of pinging the server. This is a minor, but completely backwards-compatible, API breakage: Player inherits from InboundConnection so we do not have to change ProxyPingEvent, however plugins not expecting a Player might get confused.

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/event/proxy/ProxyPingEvent.java
+++ b/api/src/main/java/com/velocitypowered/api/event/proxy/ProxyPingEvent.java
@@ -13,10 +13,11 @@ import com.velocitypowered.api.proxy.InboundConnection;
 import com.velocitypowered.api.proxy.server.ServerPing;
 
 /**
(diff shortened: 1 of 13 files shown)


Xernium (approved):
This fell off of my list, lgtm.
We should also consider how LoginInbound is handled internally, but thats for me to think about at a later date

thecmdradama:
I know this is just a comment but should "when handling this event" be removed from the comment in the ProxyPingEvent.java file?
```
* you are urged to handle this event as quickly as possible when handling this event due to the
* amount of ping packets a client can send.
```

PR #779, opened 2022-06-29
Fix PlayerChatEvent.

### LINKED ISSUE #762: Changing the message in PlayerChatEvent has no effect
Changing the message in PlayerChatEvent does not actually change the final message. Noticed this when we merged your 1.19 commits since our Chat Filter system stopped working. Please check [this ](https://github.com/PaperMC/Velocity/blob/c2b237f19665ff482a2948d0afe9db54a0197f11/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientPlaySessionHandler.java#L178) code. Instead of event.getMessage() you shall use eventMsg.

mdxd44:
Fixes #762


mdxd44:
Was fixed in 1.19.1 update

PR #787, opened 2022-07-05
Update command server and remove aliases for avoid conflicts

packetspoofer:
Update command server with a command that no proxy has so far, and to avoid conflicts with spigot / paper remove the shutdown aliases.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
@@ -209,9 +209,8 @@ public class VelocityServer implements ProxyServer, ForwardingAudience {
 
     // Initialize commands first
     commandManager.register("velocity", new VelocityCommand(this));
-    commandManager.register("server", new ServerCommand(this));
-    commandManager.register("shutdown", ShutdownCommand.command(this),
-        "end", "stop");
+    commandManager.register("goto", new ServerCommand(this));
+    commandManager.register("shutdown", ShutdownCommand.command(this));
     new GlistCommand(this).register();
 
     this.doStartupConfigLoad();


roccodev:
To be fair, you could always make your own `/goto` command with a plugin, and revoke the permission to use Velocity's. (which is disabled by default, btw)
The same goes for `/end`, etc. Why would you even give MC users permission to stop the proxy?

Moreover, I think `/server` is generally accepted as the proxy command to switch between servers.

Loremente:
Yes, but it's nicer if it's native to the proxy.

proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
```diff
@@ -209,9 +209,8 @@ void start() {
 
     // Initialize commands first
     commandManager.register("velocity", new VelocityCommand(this));
-    commandManager.register("server", new ServerCommand(this));
-    commandManager.register("shutdown", ShutdownCommand.command(this),
-        "end", "stop");
+    commandManager.register("goto", new ServerCommand(this));
+    commandManager.register("shutdown", ShutdownCommand.command(this));
```
4drian3d, line 213: This change is completely unnecessary and pointless as currently only the ConsoleCommandSource (the console) can execute the command, so there are no compatibility issues with Minecraft server commands

proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
```diff
@@ -209,9 +209,8 @@ void start() {
 
     // Initialize commands first
     commandManager.register("velocity", new VelocityCommand(this));
-    commandManager.register("server", new ServerCommand(this));
-    commandManager.register("shutdown", ShutdownCommand.command(this),
-        "end", "stop");
+    commandManager.register("goto", new ServerCommand(this));
```
4drian3d, line 212: Changing the `server `command to `goto` makes no sense, plus it breaks many setups of several servers, and it seems that it could only be useful for your individual setup

electronicboy:
this is a no go
1) /server is the general defacto expected command by server owners and users on how to do this stuff, the *fact* that other proxy environments use that is why we're using that, especially as it makes it easy for servers to transition
2) Breaking environments mid cycle, especially this far in, would just be stupid
3) Only caveat I have here is "stop", I thought that that only worked on the console, but, I can't see the logic for that, so, that might be something to look into

roccodev:
>     3. Only caveat I have here is "stop", I thought that that only worked on the console, but, I can't see the logic for that, so, that might be something to look into

If everything is configured properly, this should be prevented by this line here: https://github.com/PaperMC/Velocity/blob/dev/3.0.0/proxy/src/main/java/com/velocitypowered/proxy/command/builtin/ShutdownCommand.java#L41

proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
Loremente: Bruhh if give permission with lpv or alternative plugin for velocity permission yes

Loremente:
@electronicboy then I create another pull request by removing only stop? Or you commit this?

Xernium:
> @electronicboy then I create another pull request by removing only stop? Or you commit this?

Neither will be pulled by us. The “stop” alias only exists so more generic environments also work with velocity (ones that would normally contain a normal Minecraft server).   
We will not remove this alias.    
There is no reason for us to do so; you cannot run it in game anyway since the shutdown command is limited to only running from the console command source(s).
Because of that there was no conflict in the first place.

PR #772, opened 2022-06-23
Update to 1.19.1

Xernium:
Pull-request to track 1.19.1 changes

PR #788, opened 2022-07-09
Improved configuration version handling

### LINKED ISSUE #733: Velocity doesn't add new options to the velocity.toml config file
After updating to velocity-3.1.2-SNAPSHOT-147 I needed to change option `force-key-authentication` but saw that it was missing.  After renaming the config file and restarting Velocity to generate a new one, I notice there were quite a few options missing from the old config file.

I would expect Velocity to add in any new options and not have to manually regenerate the file then copy across existing settings.

Our dev server is running on Windows 10, in case that makes a difference.

4drian3d:
Currently the user is "forced" to handle the configuration update on their own, deleting the configuration they have in place to add the new options, whereas the job should be handled by Velocity, especially with a patch version being developed `3.1.2`.

This pull request replaces the old logic for handling version 1.0 and 2.0 configuration to a single migration. This would also allow more migrations to be done in case more configuration options or changes are added in the future, e.g. #666. Possibly resolves #733 , but only for the new options that may be added from now on

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
@@ -428,7 +428,6 @@ public class VelocityConfiguration implements ProxyConfig {
       Files.writeString(defaultForwardingSecretPath, generateRandomString(12));
     }
 
-    boolean mustResave = false;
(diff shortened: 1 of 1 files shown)


Xernium:
Collides with #772 
Please either wait and rebase or target the PR branch so we can pull this later.

4drian3d:
superseded by #1111

PR #798, opened 2022-07-16
don't error out when forwarding-file doesn't exist and is not needed

derklaro:
#743 optimized the way the forwarding secret file is read. While the change is fine, it also introduced a small annoyance I stumbled upon. When the configured and default forwarding file are missing (for whatever reason) the proxy won't start and throw an exception even when using legacy or no forwarding where no forwarding secret is needed.

There is a check a bit further down which validates that the forwarding secret is present when using bungeeguard or modern forwarding, I just extended the error message by appending a hint about the missing file there.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
@@ -422,12 +422,6 @@ public class VelocityConfiguration implements ProxyConfig {
       throw new RuntimeException("Default configuration file does not exist.");
     }
 
-    // Create the forwarding-secret file on first-time startup if it doesn't exist
-    Path defaultForwardingSecretPath = Path.of("forwarding.secret");
-    if (Files.notExists(path) && Files.notExists(defaultForwardingSecretPath)) {
-      Files.writeString(defaultForwardingSecretPath, generateRandomString(12));
-    }
-
     boolean mustResave = false;
     CommentedFileConfig config = CommentedFileConfig.builder(path)
         .defaultData(defaultConfigLocation)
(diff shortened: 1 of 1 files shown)


4drian3d (approved):

hugmanrique:
The `VelocityConfiguration#read` method needs a rework. Imo, if the secret file doesn't exist, a default one should be created even when the `velocity.toml` file exists (i.e. I would remove the first condition in [this line](https://github.com/PaperMC/Velocity/blob/d8198783739f214917f957ad70a5c3698436d5f5/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java#L427) and move the create call further down when we know that the file is indeed needed -- e.g. when the `VELOCITY_FORWARDING_SECRET` env var is unset).

Unrelated, but I don't understand the purpose of [these lines](https://github.com/PaperMC/Velocity/blob/d8198783739f214917f957ad70a5c3698436d5f5/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java#L444-L449). The `defaultConfig` variable looks unused after loading :confused:

derklaro:
After looking again it seems like the forwarding-secret presence check is there two times: during reading and during validation. Question is if we just create the file if it is missing without content and remove the check during reading 🤔

hugmanrique:
That's what I meant, i.e. remove [these lines](https://github.com/PaperMC/Velocity/blob/d8198783739f214917f957ad70a5c3698436d5f5/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java#L425-L429) (425-429) and add an `else` branch [here](https://github.com/PaperMC/Velocity/blob/d8198783739f214917f957ad70a5c3698436d5f5/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java#L499) that looks
```java
forwardingSecretString = generateRandomString(12);
Files.writeString(secretPath, forwardingSecretString);
```

This preserves the behavior prior to #712.

derklaro:
Yea that, and I'm talking about [this check](https://github.com/PaperMC/Velocity/blob/dev/3.0.0/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java#L535-L539) and [this check](https://github.com/PaperMC/Velocity/blob/dev/3.0.0/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java#L138-L152)

hugmanrique (approved):

Issue #769, opened 2022-06-22
MOTD behavior


wiktor-jozwiak:
After setting the MOTD in velocity.toml and after shutting down the server, the MOTD from velocity changes to a motd like the one in the image when shutting it down.
![image](https://user-images.githubusercontent.com/61411400/175044485-a3667cdb-0792-4fa6-8945-4a41ae6118f3.png)

Im using Velocity build #156

4drian3d:
Make sure you are connecting to Velocity and not to an individual server, that motd is the default for Minecraft Vanilla, not Velocity.

Otherwise, check how you have configured the https://github.com/PaperMC/Velocity/blob/dev/3.0.0/proxy/src/main/resources/default-velocity.toml#L67 option and check what each alternative is for

wiktor-jozwiak:
@4drian3d  I'm sure I connect via velocity, after entering the server from this ip are all the commands velocity, etc, the problem is that when I turn off the lobby for example, MOTD Velocity for a moment changes to the MOTD of ordinary servers that are connected to Velocity.

wiktor-jozwiak:
It's my configuration:
![image](https://user-images.githubusercontent.com/61411400/175047092-6ca1fa97-761a-45c3-8b18-d8ed4e67ce32.png)

56738:
Since 1.19 servers send their MOTD + icon to players after they connect. Velocity probably needs to remove those values from the [server data packet](https://wiki.vg/Protocol#Server_Data) to prevent this.

wiktor-jozwiak:
@56738 good, thanks very much for your answer

astei:
Oh, Mojang, why'd you have to do this? I'll get a PR up for this.

astei:
This will be resolved by #771

astei:
merged as 662fbc4e3cc1ebbd817af6bb01a12cc46f479c2a

Aurelien30000:
Unsure if this is expected, but this doesn't seem to be fixed when a client if kicked out of the minecraft server and then gets disconnected from the proxy.
With a normal disconnection, proxy MOTD and favicon are shown, however, if caused by a server kick, server-specific MOTD and favicon are shown, even if modified listening to the ``ProxyPingEvent``. Note: other info like sample players are always correct (proxy side).

PR #799, opened 2022-07-19
Decode ServerData packet

56738:
#771 implemented handling for the ServerData packet, but decoding was disabled, so the handler was never called and #769 still happened.


Xernium:
Thanks

PR #808, opened 2022-07-31
Set proxy to build with Java 11 compatibility

BBaoVanC:
Not sure if this is the correct way to do it, but I use Java 18 on my system and building Velocity results in the proxy jar not being compatible with Java 17.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -32,8 +32,9 @@ allprojects {
     }
 
     java {
-        sourceCompatibility = JavaVersion.VERSION_11
-        targetCompatibility = JavaVersion.VERSION_11
+        toolchain {
+            languageVersion = JavaLanguageVersion.of(11)
+        }
     }
 
     repositories {
--- a/proxy/build.gradle
+++ b/proxy/build.gradle
@@ -13,6 +13,11 @@ license {
     header = project.rootProject.file('HEADER.txt')
 }
 
+java {
+    sourceCompatibility = JavaVersion.VERSION_11
+    targetCompatibility = JavaVersion.VERSION_11
+}
+
 jar {
     manifest {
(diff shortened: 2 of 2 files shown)


4drian3d:
I don't think this change is so necessary, it would only affect custom builds, and in that case, everyone can modify Velocity as required

BBaoVanC:
What do you mean by custom builds?

4drian3d:
I mean that if you clone the project directly or through a fork it will affect when compiling on your own, but for the official Velocity distribution it will not have any effect

electronicboy:
to my understanding, one should set it through the toolchain, that way it can do other stuff like ensuring that pulled in stuff is compatible with 11 and that the final output supports 11

BBaoVanC:
> to my understanding, one should set it through the toolchain, that way it can do other stuff like ensuring that pulled in stuff is compatible with 11 and that the final output supports 11

So do you mean that I should set my system's Java version instead when building? It's kind of annoying to do that, I feel like it makes the most sense for it to build properly automatically, instead of needing to be set properly in CI, or on your own system or whatever.

electronicboy:
No, I mean using he gradle toolchain system, which basically properly sets the JVM release, etc, properly

BBaoVanC:
Like this, in 8b87ca3e87bb5ac189c5170a904f02619572686f?

electronicboy:
yes

BBaoVanC:
I also see that in `api/build.gradle` and `native/build.gradle`, it uses `sourceCompatibility` and `targetCompatibility` to set the Java version. Should this be changed to Gradle `toolchain` as well?

astei:
I went ahead and made this change as part of the conversion of Velocity to use Gradle's Kotlin DSL

PR #810, opened 2022-08-01
Fix command packet wrong mark as unsigned

james58899:
Fix player kick for `Chat message validation failure` when using F3+F4 switch game mode after sending some message.


Xernium:
`previousMessages.length == 0` is a valid state. Happens when you log in first.     
I did not have time to look at this but from the looks we should not have an issue here

james58899:
If the packet is marked as unsigned, the `previousMessages` will be dropped and the chat chain will be broken.
So only mark as unsigned if `previousMessages` is empty.

Maybe making packets forward to the server always preserve `previousMessages` is better but changes could be larger.

PR #811, opened 2022-08-02
Show proxy-wide online players in server ping

BBaoVanC:
This makes it so you can see online players on the entire proxy when hovering over the player count (only has an effect when ping passthrough is disabled, otherwise the backend server's player list is already used).

Preview is limited to 12 players, same as vanilla:

image

Players that have "Allow Server Listings" set to false show up as "Anonymous Player" (same as vanilla):

image

And when refreshing the server list, the list reshuffles.

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/proxy/server/ServerPing.java
+++ b/api/src/main/java/com/velocitypowered/api/proxy/server/ServerPing.java
@@ -429,6 +429,10 @@ public final class ServerPing {
    */
   public static final class SamplePlayer {
 
+    public static final SamplePlayer ANONYMOUS = new SamplePlayer(
+        "Anonymous Player",
+        new UUID(0L, 0L)
+    );
     private final String name;
     private final UUID id;
 
--- a/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
@@ -78,6 +78,8 @@ public class VelocityConfiguration implements ProxyConfig {
   private boolean onlineModeKickExistingPlayers = false;
   @Expose
   private PingPassthroughMode pingPassthrough = PingPassthroughMode.DISABLED;
+  @Expose
+  private boolean samplePlayersInPing = false;
   private final Servers servers;
   private final ForcedHosts forcedHosts;
   @Expose
@@ -105,6 +107,7 @@ public class VelocityConfiguration implements ProxyConfig {
       boolean preventClientProxyConnections, boolean announceForge,
       PlayerInfoForwarding playerInfoForwardingMode, byte[] forwardingSecret,
       boolean onlineModeKickExistingPlayers, PingPassthroughMode pingPassthrough,
+      boolean samplePlayersInPing,
       boolean enablePlayerAddressLogging, Servers servers, ForcedHosts forcedHosts,
       Advanced advanced, Query query, Metrics metrics, boolean forceKeyAuthentication) {
     this.bind = bind;
@@ -117,6 +120,7 @@ public class VelocityConfiguration implements ProxyConfig {
     this.forwardingSecret = forwardingSecret;
     this.onlineModeKickExistingPlayers = onlineModeKickExistingPlayers;
     this.pingPassthrough = pingPassthrough;
+    this.samplePlayersInPing = samplePlayersInPing;
     this.enablePlayerAddressLogging = enablePlayerAddressLogging;
     this.servers = servers;
     this.forcedHosts = forcedHosts;
@@ -371,6 +375,10 @@ public class VelocityConfiguration implements ProxyConfig {
     return pingPassthrough;
   }
 
+  public boolean getSamplePlayersInPing() {
+    return samplePlayersInPing;
+  }
+
   public boolean isPlayerAddressLoggingEnabled() {
     return enablePlayerAddressLogging;
   }
@@ -503,6 +511,8 @@ public class VelocityConfiguration implements ProxyConfig {
       final PingPassthroughMode pingPassthroughMode = config.getEnumOrElse("ping-passthrough",
               PingPassthroughMode.DISABLED);
 
+      final boolean samplePlayersInPing = config.getOrElse("sample-players-in-ping", false);
+
       final String bind = config.getOrElse("bind", "0.0.0.0:25565");
       final int maxPlayers = config.getIntOrElse("show-max-players", 500);
       final boolean onlineMode = config.getOrElse("online-mode", true);
@@ -533,6 +543,7 @@ public class VelocityConfiguration implements ProxyConfig {
               forwardingSecret,
               kickExisting,
               pingPassthroughMode,
+              samplePlayersInPing,
               enablePlayerAddressLogging,
               new Servers(serversConfig),
               new ForcedHosts(forcedHostsConfig),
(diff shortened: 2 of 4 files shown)


proxy/src/main/java/com/velocitypowered/proxy/connection/util/ServerListPingHandler.java
```diff
@@ -48,7 +48,9 @@ private ServerPing constructLocalPing(ProtocolVersion version) {
         new ServerPing.Version(version.getProtocol(),
             "Velocity " + ProtocolVersion.SUPPORTED_VERSION_STRING),
         new ServerPing.Players(server.getPlayerCount(), configuration.getShowMaxPlayers(),
-            ImmutableList.of()),
+            server.getAllPlayers().stream().map(p -> {
```
electronicboy: 1) This probably should respect the vanilla behavior of not sending the entire list of players, as well as actually randomising it
2) This needs to respect the name anonymisation setting added in recent versions

electronicboy (changes requested):

BBaoVanC:
Just realized that I was building on Java 18, and `Stream#toList()` was added in Java 16. So I'll fix that in a moment.

proxy/src/main/java/com/velocitypowered/proxy/connection/util/ServerListPingHandler.java
BBaoVanC: 1. Where can I see the vanilla behavior for that?
2. is that [this](https://github.com/PaperMC/Velocity/blob/86c65f3910a0c1433a1741d5b6408a1508738047/proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/ClientSettings.java#L38)?

electronicboy: 1) in the vanilla server code, inside of MinecraftServer#tickServer; but basically they just sublist the online player list from (accounting for sample size and online count, etc), doesn't need to follow this exact behaviour, just shouldn't really be sending the full list and should be shuffled for some level of randomness
2) yes
3) In general, this should probably have a config option, not sure what others think on that

BBaoVanC:
Just now tested it with 14 clients, the preview should be working properly. I updated the PR description with screenshots of it.

proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
```diff
@@ -525,6 +532,7 @@ public static VelocityConfiguration read(Path path) throws IOException {
         PlayerInfoForwarding.NONE);
     PingPassthroughMode pingPassthroughMode = config.getEnumOrElse("ping-passthrough",
         PingPassthroughMode.DISABLED);
+    boolean samplePlayersInPing = config.getOrElse("sample-players-in-ping", true);
```
hugmanrique: Imo the default should be `false` to maintain the current behavior; lots of users already use a plugin to override the player list and enabling this is wasteful since the plugin overrides the list anyways.

proxy/src/main/java/com/velocitypowered/proxy/connection/util/ServerListPingHandler.java
```diff
@@ -44,11 +47,29 @@ public ServerListPingHandler(VelocityServer server) {
...
   private ServerPing constructLocalPing(ProtocolVersion version) {
     VelocityConfiguration configuration = server.getConfiguration();
+    List<ServerPing.SamplePlayer> samplePlayers;
+    if (configuration.getSamplePlayersInPing()) {
+      List<ServerPing.SamplePlayer> unshuffledPlayers = server.getAllPlayers().stream()
+              .map(p -> {
+                if (p.getPlayerSettings().isClientListingAllowed()) {
+                  return new ServerPing.SamplePlayer(p.getUsername(), p.getUniqueId());
+                } else {
+                  return new ServerPing.SamplePlayer(
```
hugmanrique: We can cache this object in a `static final`, `ServerPing.SamplePlayer` is an immutable class.

proxy/src/main/java/com/velocitypowered/proxy/connection/util/ServerListPingHandler.java
```diff
@@ -44,11 +47,29 @@ public ServerListPingHandler(VelocityServer server) {
...
+                          "Anonymous Player",
+                          new UUID(0L, 0L)
+                  );
+                }
+              })
+              .collect(Collectors.toList());
+      Collections.shuffle(unshuffledPlayers);
+      samplePlayers = unshuffledPlayers.stream().limit(12).collect(Collectors.toList());
```
hugmanrique: There's no need for streams here, see [`List#subList`](https://docs.oracle.com/javase/8/docs/api/java/util/List.html#subList-int-int-). Ideally we would use Fisher-Yates in place (i.e. Durnstenfield) to avoid shuffling the entire list, but I couldn't find any implementation on the JDK (the `Collections.shuffle` javadoc hints at using this method, but doesn't allow stopping at a given index) nor Guava. We can evaluate the performance hit on larger servers and see if it's worth the extra method.

hugmanrique (changes requested):

proxy/src/main/java/com/velocitypowered/proxy/connection/util/ServerListPingHandler.java
BBaoVanC: Do you mean like holding a list of sample players statically as a variable in a class (maybe `VelocityServer`)? If that's the case then maybe I should change the current shuffling behavior so it either only shuffles once per join/leave event, or I could just have it insert each player join at a random point in the list.

What do you think?

hugmanrique: Sorry I wasn't clear enough, meant that we should cache the "unknown player" instance, not the entire list, i.e. there's no need to create a new "unknown player" `SamplePlayer` instance every time `p.getPlayerSettings().isClientListingAllowed()` returns `false`.

BBaoVanC: Is this what you meant? ac484c972f4e4c3e7312f629b31fb5e7d6f369ce

api/src/main/java/com/velocitypowered/api/proxy/server/ServerPing.java
```diff
@@ -412,6 +412,8 @@ public int hashCode() {
 
   public static final class SamplePlayer {
 
+    public static final SamplePlayer Anonymous = new SamplePlayer("Anonymous Player", new UUID(0L, 0L));
```
hugmanrique: The last minor nit: constants  are `UPPER_CASE`.

BBaoVanC: Fixed

hugmanrique (approved):

proxy/src/main/java/com/velocitypowered/proxy/connection/util/ServerListPingHandler.java
BBaoVanC: All three of these are fixed btw

BBaoVanC:
Rebased it, @electronicboy could you look at/review this?

4drian3d (approved):

4drian3d (approved):

Issue #812, opened 2022-08-03
Offline players cannot send messages


MagicalSheep:
Paper version: `git-Paper-104 (MC: 1.19.1) (Git: 7044a9c)`
Velocity version: `3.1.2-SNAPSHOT (git-e5b84ecf-b164)`

**Steps to reproduce:**

1. Run paper server using `java -jar paper-1.19.1-104.jar nogui`, and then stop it.
2. Set `online-mode=false` and `enforce-secure-profile=false` in `server.properties`.
3. Run paper server again.

So far, offline players can join the server and send messages successfully.

4. Run velocity server using `java -jar velocity-3.1.2-SNAPSHOT-164.jar`, and then stop it.
5. Set `online-mode = false` and `force-key-authentication = false` in `velocity.toml`.
6. Add `lobby = "127.0.0.1:25565"` into `servers` section and clear `forced-hosts` section.
7. Run velocity server again.

So far, offline players can join the server and send command but will get disconnected immediately when sending messages. Here is the error log from velocity server.

```
[16:02:21 INFO]: [connected player] sheep (/127.0.0.1:54538) has disconnected: An internal error occurred in your connection.
[16:02:21 ERROR]: [connected player] sheep (/127.0.0.1:54538): exception encountered in com.velocitypowered.proxy.connection.client.ClientPlaySessionHandler@55c364d3
com.velocitypowered.proxy.util.except.QuietRuntimeException: A packet did not decode successfully (invalid data). If you are a developer, launch Velocity with -Dvelocity.packet-decode-logging=true to see more.
[16:02:21 INFO]: [server connection] sheep -> lobby has disconnected
```

Since we keep the setting `player-info-forwarding-mode = "NONE"`, we can still join the paper server directly when velocity server running. The fact is that this problem only occurs when using velocity to join the server. For this reason, I can confirm that the problem is not related to the client or Paper, and it is caused by Velocity.

Some other people have the same issue: #806. This issue has closed but do not solve the problem.

DIDIRUS4:
> 
**+1**
Same error `com.velocitypowered.proxy.util.except.QuietRuntimeException: A packet did not decode successfully (invalid data)`

astei:
fixed with #814

Issue #806, opened 2022-07-31
Sending messages on 1.19.1 from offline player


Kapitowa:
Latest velocity, which support 1.19.1, without plugins.
Latest paper server without plugins.
`force-key-authentication = false` in velocity.toml
`enforce-secure-profile=false` in server.properties
Offline player can't send any message, it's just kicked.

Paper log is clear.
Error log from velocity on disconnect:
```
[17:37:30 INFO]: [connected player] Mr_White (/127.0.0.1:56789) has disconnected: An internal error occurred in your connection.
[17:37:30 ERROR]: [connected player] Mr_White (/127.0.0.1:56789): exception encountered in com.velocitypowered.proxy.connection.client.ClientPlaySessionHandler@7c8503e8
io.netty.handler.codec.CorruptedFrameException: Error decoding class com.velocitypowered.proxy.protocol.packet.chat.PlayerChat Direction SERVERBOUND Protocol 1.19.1 State PLAY ID 5
        at com.velocitypowered.proxy.protocol.netty.MinecraftDecoder.handleDecodeFailure(MinecraftDecoder.java:128) ~[velocity-3.1.2-SNAPSHOT-163.jar:3.1.2-SNAPSHOT (git-1a3fba42-b163)]
        at com.velocitypowered.proxy.protocol.netty.MinecraftDecoder.tryDecode(MinecraftDecoder.java:83) ~[velocity-3.1.2-SNAPSHOT-163.jar:3.1.2-SNAPSHOT (git-1a3fba42-b163)]
        at com.velocitypowered.proxy.protocol.netty.MinecraftDecoder.channelRead(MinecraftDecoder.java:58) ~[velocity-3.1.2-SNAPSHOT-163.jar:3.1.2-SNAPSHOT (git-1a3fba42-b163)]
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379) ~[velocity-3.1.2-SNAPSHOT-163.jar:3.1.2-SNAPSHOT (git-1a3fba42-b163)]
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365) ~[velocity-3.1.2-SNAPSHOT-163.jar:3.1.2-SNAPSHOT (git-1a3fba42-b163)]
[...]

PR #814, opened 2022-08-03
Fix decoding of unsigned messages on 1.19.1

JNNGL:
Some clients may send unsigned messages with a non-zero salt on 1.19.1 (#812, #806)

1.19.1:
image

1.19:
image


AVuTuVA:
Just checked and yes, it solved the problem!

astei (approved):

Issue #804, opened 2022-07-29
A plugin tried to cancel a signed chat message.


skerit:
So I'm preparing to upgrade my servers to 1.19.1, I decided to start with my network-wide-chat plugin and I immediately run into this problem:

> A plugin tried to cancel a signed chat message. This is no longer possible in 1.19.1 and newer. Disconnecting player

Added in this line:
https://github.com/PaperMC/Velocity/blob/c57fb489f3cc664fe05cdbdbb24fdca1347ef446/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientPlaySessionHandler.java#L200

Is there a technical reason why Velocity wants to turn this into an error?
Because if I simply disable that line and rebuild, my original plugin works fine.
Player chat messages are actually sent to all players across all servers as a simple system message.

electronicboy:
Due to changes in 1.19.1, cancelling chat has wider implications, i.e. the thing cancelling chat still needs to ensure that everything else gets the relevant updates, etc, this is especially an issue in multi proxy environments in which API will need to be developed for

astei:
I think it's very unlikely that we continue to support chat manipulation on the proxy side. It's an entire can of worms we don't really want to touch. In theory, there might be a path forward if the server simply doesn't verify messages, but we want Velocity to work on as many setups as possible.

A248:
If you choose to forsake chat handling, you'll have to accept that you have broken your API. Many plugins rely on this feature. It would be rather unfortunate for Velocity to stop handling chat entirely.

astei:
For the most part, I am extrapolating. Player chat manipulation was broken by Mojang and we certainly haven't seen what Mojang's ultimate plans are with this.

We can support chat manipulation if Velocity rewrites all chat messages to be unsigned - this is probably the _only_ path forward for allowing plugins to rewrite messages. Long term, however, Mojang may throw another wrench in the works (i.e. having Minecraft accounts that can _only_ join servers that verify profiles and have signed chat messages) and that will soon force a decision upon us. All this really does is kick the can down the road.

electronicboy:
Supporting chat on the proxy is doable, but it means deferring more behavior towards plugins, i.e. probably making them in control of updating the clients state, it creates many dozen issues because you need to keep the chain intact, otherwise you'll cause the client to DC whenever something uses a signed message, which is where the headaches come in

[...]

PR #816, opened 2022-08-03
Add option to disable authentication keys

roccodev:
As mentioned in #804 (which this PR is also a workaround for), Velocity's aim is to support as many setups as possible.

With the 1.19.1 update, it is no longer possible to cancel or edit the order of incoming messages, while still keeping support for signed chat and player reporting.  
However, Velocity currently lacks the option to forgo message signing (e.g., to freely cancel and transform everything into a system message) for 1.19.1 clients and servers.

This PR migrates the old `force-key-authentication` option into a three-variant enum (`key-authentication-policy`), with the default being the exact same behavior before this change. Additionally, if `force-key-authentication` was manually turned off, this falls back to the enabled state.

Disabling authentication keys is something the user must do manually via a config change, and it is discouraged in the option's comment.

### DIFF
--- a/proxy/src/main/resources/default-velocity.toml
+++ b/proxy/src/main/resources/default-velocity.toml
@@ -15,8 +15,18 @@ show-max-players = 500
 # Should we authenticate players with Mojang? By default, this is on.
 online-mode = true
 
-# Should the proxy enforce the new public key security standard? By default, this is on.
-force-key-authentication = true
+# Should the proxy support secure profiles (and signed chat) for 1.19.1+ clients and servers?
+# This option only affects 1.19.1+ clients and servers.
+#
+# Available options:
+# - "disabled":    Secure profiles are not checked, all chat messages will appear unsigned to backend
+#                  servers. This option is not recommended as it may break compatibility with 1.19.1
+#                  servers, and it will cause newer clients to show a warning upon login.
+# - "enabled":     The proxy will support secure profiles for 1.19.1+ clients, but will also allow
+#                  insecure profiles. 1.19.1+ backend servers might reject insecure logins.
+# - "enforced":    (default) The proxy will only allow secure profiles to log in. This has the same effect
+#                  as the old "force-key-authentication" option.
+key-authentication-policy = "ENFORCED"
 
 # If client's ISP/AS sent from this proxy is different from the one from Mojang's
 # authentication server, the player is kicked. This disallows some VPN and proxy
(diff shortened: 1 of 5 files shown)


4drian3d (approved):

proxy/src/main/resources/default-velocity.toml
```diff
@@ -15,8 +15,18 @@ show-max-players = 500
...
 
-# Should the proxy enforce the new public key security standard? By default, this is on.
-force-key-authentication = true
+# Should the proxy support secure profiles (and signed chat) for 1.19.1+ clients and servers?
+# This option only affects 1.19.1+ clients and servers.
+#
+# Available options:
+# - "disabled":    Secure profiles are not checked, all chat messages will appear unsigned to backend
+#                  servers. This option is not recommended as it may break compatibility with 1.19.1
+#                  servers, and it will cause newer clients to show a warning upon login.
```
BBaoVanC, line 24: Should this option allow cancelling signed chat messages on 1.19.1+? I tested this branch on a local server, but it still kicks me when sending any message on 1.19.1. This means that I can't convert everything to a system message (through a global chat plugin for example)

roccodev: What error are you getting? It should work if you set it to `disabled`.

BBaoVanC: Same error as before where it says canceling chat messages isn't supported on 1.19.1+, and it kicks the player. Maybe [this if statement](https://github.com/PaperMC/Velocity/blob/cae41c33a1a758f5ecd0a4361c04e8652dd67072/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientPlaySessionHandler.java#L213) needs to be modified

roccodev: When key auth is disabled, `player.getIdentifiedKey()` is null. I can cancel messages just fine, maybe you need to use `DISABLED`? (in caps)

BetTD (approved):

proxy/src/main/resources/default-velocity.toml
BetTD: Works completely fine on my end. When disabled, messages can be blocked. Otherwise, the killswitch gets triggered as the Velocity devs intended.

Xernium (changes requested):
While I agree with the motive of this PR this will need to be implemented differently to work around all kinds of unwanted scenarios introduced with #817 
Please understand that I cant pull this outright, but I will take inspiration while I work on this and I will pull parts of this to be in the final version of #817

roccodev:
Sounds good!

Strahilchu:
It should be encouraged and the default setting for sure!

gamingcodedev:
bump

Issue #770, opened 2022-06-22
sent message with invalid signature


IIMuyaII:
if I do "player-into-forwarding-mode =modern", this is a problem.

`! [Tue 18:53:11 WARN  Server/PlayerConnection] IIMuyaII sent message with invalid signature: 'ghfhgfhghfghgf'
! [Tue 18:53:17 WARN  Server/PlayerConnection] IIMuyaII sent message with invalid signature: 'fghfghfghfg'
! [Tue 18:53:38 WARN  Server/PlayerConnection] IIMuyaII sent message with invalid signature: 'a'

this problem only happens if you are logged in with client vanilla 1.19.

kyngs:
Does this appear on the server or the proxy? If possible attach the log of both.

IIMuyaII:
server, 1.19 this is reflected on my server console.

i can forward the log in the morning.
velocityversion:
> velocity version
[23:25:15 INFO]: Velocity 3.1.2-SNAPSHOT (git-0caff79b-b156)
[23:25:15 INFO]: Copyright 2018-2021 Velocity Contributors. Velocity is licensed under the terms of the GNU General Public License v3.
[23:25:15 INFO]: velocitypowered.com - GitHub

1.19 Server Version:
[23:26:32 INFO]: Checking version, please wait...
[23:26:33 INFO]: This server is running Paper version git-Paper-33 (MC: 1.19) (Implementing API version 1.19-R0.1-SNAPSHOT) (Git: 147a2b7)
You are running the latest version
Previous version: git-Paper-32 (MC: 1.19)

HarisAli3:
This same thing is happening with me , i am using latest velocity with 1.19 server . 
![image](https://user-images.githubusercontent.com/99520506/182888770-1d13d26c-2c68-418d-8842-aa374900a866.png)
This message appears on console and is not displayed in chat. Means no can see this message nor me .

NotMarra:
same 
![image](https://user-images.githubusercontent.com/70058346/184101195-6aaaeb2d-3979-4faf-b861-8aedb11bae40.png)

IIMuyaII:
I have updated to 1.19.2 paper version and velocity latest version this problem has been solved.

4drian3d:
This problem has most likely been fixed in current versions as mentioned, personally I have not been able to reproduce the bug either

PR #817, opened 2022-08-03
Chat sync V1

Xernium:
This is an attempt at allowing servers behind Velocity on 1.19.1 to use secure chat.    
Related to https://github.com/PaperMC/Paper/pull/8219

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/backend/BackendPlaySessionHandler.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/backend/BackendPlaySessionHandler.java
@@ -36,6 +36,7 @@ import com.velocitypowered.proxy.command.CommandGraphInjector;
 import com.velocitypowered.proxy.connection.MinecraftConnection;
 import com.velocitypowered.proxy.connection.MinecraftSessionHandler;
 import com.velocitypowered.proxy.connection.client.ClientPlaySessionHandler;
+import com.velocitypowered.proxy.connection.client.ConnectedPlayer;
 import com.velocitypowered.proxy.connection.player.VelocityResourcePackInfo;
 import com.velocitypowered.proxy.connection.util.ConnectionMessages;
 import com.velocitypowered.proxy.protocol.MinecraftPacket;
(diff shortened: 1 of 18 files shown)


Aurelien30000:
Hey! If you need me to test something about your fight against last seen chat messages, I have a setup of a velocity proxy and two backend paper servers locally (remotely if needed). Atm, same error as usual.

kyngs:
Would this also fix #770 ?

proxy/src/main/java/com/velocitypowered/proxy/connection/backend/BackendPlaySessionHandler.java
```diff
@@ -270,6 +272,16 @@ public boolean handle(AvailableCommands commands) {
     return true;
   }
 
+  @Override
+  public boolean handle(ServerPlayerChat packet) {
+    HeaderData chatHeader = packet.getHeaderData();
+    if (chatHeader != null) {
+      ConnectedPlayer who = this.serverConn.getPlayer();
+      who.pushHeader(chatHeader);
```
Kichura, line 280: This causes gradle to fail - Is there a library missing for it?

electronicboy: It was a push of a WIP set of changes is my understanding, stuff is still a WIP

Xernium: As cat said, this was me saving progress

davidmayr:
Whats the status of this?

D3it7i:
I have a suggestion, especially this pr seems to be difficult to complete, an optional option, when turned on, only verify the player chat security in the proxy server, so that the chat report on the client side works normally, and the backend server says the chat is not Safe, but the backend is all under a secure network, isn't it?

electronicboy:
Secure network is 100% irrelevant, you can't handle this stuff on just the proxy because the entire state of the chain needs to be maintained on the backend server. There is ongoing efforts to work on this stuff however it is a moving target as mojangs system is still also irrelevant, in part which should aid this issue

D3it7i:
> the entire state of the chain needs to be maintained on the backend server.

Why need back-end server maintenance, when you just want the back-end server to record an unsafe chat message. Isn't the server just responsible for verifying signatures and broadcasting chat signatures?

Xernium:
I am currently not active as developer but let me say: it’s not that simple.
This PR is heavily outdated due to me rebasing a ton of stuff before going into hiatus.
See https://github.com/xernium/Velocity/tree/work/wip/s02 if you want to see where I left of

PR #820, opened 2022-08-04
Clarify that plugin IDs must be lowercase

A248:


electronicboy (approved):

hugmanrique:
#751 took the alternative approach of accepting uppercase but normalizing all IDs (including those of dependencies).

Issue #555, opened 2021-08-04
SimpleCommand fails to forward commands with args


roccodev:
Might be related to #369.

When `hasPermission() == false`, `SimpleCommand` classes will only forward the command if no arguments are used. Adding any argument will make the proxy catch it, throwing an error.

Simple PoC:
```java
import com.velocitypowered.api.command.SimpleCommand;
import net.kyori.adventure.text.Component;
import net.kyori.adventure.text.format.NamedTextColor;

public class TestCommand implements SimpleCommand {
    @Override
    public void execute(Invocation invocation) {
        invocation.source().sendMessage(Component.text("Invoked!", NamedTextColor.GREEN));
    }

    @Override
    public boolean hasPermission(Invocation invocation) {
        // With this, the command should be forwarded to the backend server in any case.
        return false;
    }
}

// ...

commandManager.register("test", new TestCommand()); // same effect with CommandMeta
```

**Expected result**
`/test` -> Forwarded to backend
`/test abc` -> Forwarded to backend

**Actual result**
`/test` -> Forwarded to backend
`/test abc` -> `Incorrect argument for command at position 5: test <--[HERE]` (in red, Brigadier error)

When `hasPermission` is changed to return `true`, the command works as expected for `hasPermission() == true`. (printing `Invoked!` with and without args)

hugmanrique:
This is expected behavior (and as you say is essentially the same as the linked issue) and mimics Brigadier's parsing mechanism.

PR #824, opened 2022-08-06
Pass-through impermissible commands.

CoreyShupe:
**The Problem**
Currently if you have a setup such as:
```
proxy A declares command `/hello <string list>` under permission `example`
server A declares command `/hello [<literal x>] <string list>` under no permission

player without permission `example` joins proxy A and gets forwarded to server A

Expected:
the tab completion for `/hello` along with command execution should forward to the backend server

Actual:
Parsing fails at the location of `hello <==` and it refuses to pass anything including command execution along.
```

**Soltuion:**
We should be checking permissibility before running the dispatcher parse/executor - if we know the user doesn't have permission we can simply pass through the request. This also goes for command result executors and parsers.

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/command/CommandManager.java
+++ b/api/src/main/java/com/velocitypowered/api/command/CommandManager.java
@@ -125,4 +125,13 @@ public interface CommandManager {
    * @return true if the alias is registered; false otherwise
    */
   boolean hasCommand(String alias);
+
+  /**
+   * Determines whether a player has the command in their "permissible scope".
+   *
+   * @param alias the command alias to check
+   * @param source the source to check the permissibility against
+   * @return true if the alias is registered - and accessible by the source; false otherwise
+   */
+  boolean hasPermissibleCommand(String alias, CommandSource source);
 }
--- a/proxy/src/main/java/com/velocitypowered/proxy/command/VelocityCommandManager.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/command/VelocityCommandManager.java
@@ -22,10 +22,14 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
 import com.mojang.brigadier.CommandDispatcher;
 import com.mojang.brigadier.ParseResults;
+import com.mojang.brigadier.StringReader;
+import com.mojang.brigadier.context.CommandContextBuilder;
+import com.mojang.brigadier.context.StringRange;
 import com.mojang.brigadier.exceptions.CommandSyntaxException;
 import com.mojang.brigadier.suggestion.Suggestion;
 import com.mojang.brigadier.suggestion.Suggestions;
 import com.mojang.brigadier.tree.CommandNode;
+import com.mojang.brigadier.tree.LiteralCommandNode;
 import com.mojang.brigadier.tree.RootCommandNode;
 import com.velocitypowered.api.command.BrigadierCommand;
 import com.velocitypowered.api.command.Command;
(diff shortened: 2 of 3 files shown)


hugmanrique:
No, we document (and API users expect) that all the children of an alias node registered by the proxy **must** come from the proxy, see https://github.com/PaperMC/Velocity/blob/fee292bcc9b88638af94151286877296c1948f56/proxy/src/main/java/com/velocitypowered/proxy/command/CommandGraphInjector.java#L61. Thus, the backend server's nodes are discarded when constructing the command graph.

This change opens up lots of edge cases around redirects (I can elaborate in a following comment if you want) and does something (constructing context objects for permission checking) I specifically discouraged in https://github.com/PaperMC/Velocity/blob/fee292bcc9b88638af94151286877296c1948f56/proxy/src/main/java/com/velocitypowered/proxy/command/CommandGraphInjector.java#L107. This is for various reasons: Brigadier should be the only one creating these objects (during parsing), since it is the only one in charge of handling them. Maintaining the reader state along with the parsed node list is quite complicated. Furthermore, the current code doesn't check the context-aware permission predicate, used by the `InvocableCommand` implementations and API users that add their own `BrigadierCommand`s (i.e. this breaks API). Even if the current logic is correct, constructing the objects needed for the context-aware predicate is a greater hack and requires "inventing" a command execution by the player (I know because I've tried :P). Lastly, one of the main goals of the command system overhaul in the Velocity 1.1->2.0 transition was to avoid duplicate permission checking calls to plugins. This re-introduces the problem.

Some notes on the implementation, even if I disagree overall: these changes should go in `CommandGraphInjector` to avoid adding public API (not needed) and to leave the packet-handling code untouched (avoiding further entangling the networking and command-related portions of Velocity). Minor change, the methods in `CommandManager` need to lock to ensure thread-safety. Also, tests are needed for each command subinterface.

Ultimately proxy commands have their own unique aliases, and must not be shared with the backend server's aliases. If the same alias absolutely needs to be shared, the proxy plugin can register all involved nodes and forward execution to the backend servers by returning [`BrigadierCommand.FORWARD`](https://jd.papermc.io/velocity/3.0.0/com/velocitypowered/api/command/BrigadierCommand.html#FORWARD) from its `Command`. This behavior is both supported by `BrigadierCommand`s, and invocable commands (via the [hinting system](https://jd.papermc.io/velocity/3.0.0/com/velocitypowered/api/command/CommandMeta.Builder.html#hint(com.mojang.brigadier.tree.CommandNode))).

A similar issue was discussed in #555.

The command system is somewhat involved, and these changes need to be discussed before code gets written. I'm sorry for the wasted time, but I hope the explanation above serves as an introduction to further improve this system safely.

CoreyShupe:
> Ultimately proxy commands have their own unique aliases, and must not be shared with the backend server's aliases. If the same alias absolutely needs to be shared, the proxy plugin can register all involved nodes and forward execution to the backend servers by returning [`BrigadierCommand.FORWARD`](https://jd.papermc.io/velocity/3.0.0/com/velocitypowered/api/command/BrigadierCommand.html#FORWARD) from its `Command`. This behavior is both supported by `BrigadierCommand`s, and invocable commands (via the [hinting system](https://jd.papermc.io/velocity/3.0.0/com/velocitypowered/api/command/CommandMeta.Builder.html#hint(com.mojang.brigadier.tree.CommandNode))).

Not wasted time, we'll be running a version of this PR in production in my server so it's not necessarily something I'm worried about getting out fast.

We can discuss this thoroughly before it gets placed in the upstream though. I'm happy to do that.

I'd like this to be the PR where it's solved though - it's not proper behavior even if it's intended.

tony710521 (approved):

PR #829, opened 2022-08-08
Fix #828

### LINKED ISSUE #828: Attempting to register a packet with a protocol version same as lastValidProtocolVersion throws an Exception.
When I attempt to register a packet mapping with the same protocol version and last valid protocol version **while the proxy's max protocol version is higher than the mapping protocol version**, I get the following exception:
```
Next mapping version (1.19) should be lower then current (1.19)
```
This is related to this code part:
https://github.com/PaperMC/Velocity/blob/6496983ca3b36db56b50cd3a9589c254460f5cfc/proxy/src/main/java/com/velocitypowered/proxy/protocol/StateRegistry.java#L423-L426

Example scenario when this is unwanted:
Let's get back to when 1.19 just came out and changed the mapping for the MapData packet. I went and created the following mapping:
id = 0x24
protocolVersion = 759 (1.19)
lastValidProtocolVersion = (1.19)

This worked fine, however, when I downloaded Velocity with 1.19.1 support, this broke.
The obvious solution would be to add a mapping supporting 1.19.1, but, this will break when you install Velocity with 1.19.2 support and so on...

A long-term solution would probably be to delete the mentioned code block completely.

***I realize that I am using internal Velocity stuff and not the API, however, I have no other option***

kyngs:
Fixes #828, actually written by @Xernium

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/protocol/StateRegistry.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/protocol/StateRegistry.java
@@ -420,7 +420,9 @@ public enum StateRegistry {
         ProtocolVersion to = current == next ? lastValid != null
                 ? lastValid : getLast(SUPPORTED_VERSIONS) : next.protocolVersion;
 
-        if (from.compareTo(to) >= 0 && from != getLast(SUPPORTED_VERSIONS)) {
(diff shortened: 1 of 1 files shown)


kyngs:
Sorry for bumping this so early, but can we please get this merged? This is a pretty time-sensitive matter for my plugin.

astei (approved):
Please note that this is a volunteer project - there is no guarantee that "time sensitivity" will be followed

PR #839, opened 2022-08-15
Improving incorrect namespace/name error

IkeVoodoo:
Rather than getting the following:
`namespace is not valid`
or
`name is not valid`

You now get:

`namespace is not valid, must match: [a-z0-9/\\-_]* got your:namespace`
`name is not valid, must match: [a-z0-9/\\-_]* got your:name`

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/proxy/messages/MinecraftChannelIdentifier.java
+++ b/api/src/main/java/com/velocitypowered/api/proxy/messages/MinecraftChannelIdentifier.java
@@ -52,10 +52,10 @@ public final class MinecraftChannelIdentifier implements ChannelIdentifier {
     Preconditions.checkArgument(!Strings.isNullOrEmpty(namespace), "namespace is null or empty");
     Preconditions.checkArgument(name != null, "namespace is null or empty");
     Preconditions.checkArgument(VALID_IDENTIFIER_REGEX.matcher(namespace).matches(),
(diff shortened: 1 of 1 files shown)


api/src/main/java/com/velocitypowered/api/proxy/messages/MinecraftChannelIdentifier.java
```diff
@@ -52,9 +52,10 @@ public static MinecraftChannelIdentifier create(String namespace, String name) {
     Preconditions.checkArgument(!Strings.isNullOrEmpty(namespace), "namespace is null or empty");
     Preconditions.checkArgument(name != null, "namespace is null or empty");
     Preconditions.checkArgument(VALID_IDENTIFIER_REGEX.matcher(namespace).matches(),
-        "namespace is not valid");
+        String.format("namespace is not valid, must match: %s got %s", VALID_IDENTIFIER_REGEX.toString(), namespace));
```
hugmanrique: Instead of formatting manually on every check, use the `checkArgument` methods that take a message template, see https://guava.dev/releases/20.0/api/docs/com/google/common/base/Preconditions.html#checkArgument-boolean-java.lang.String-java.lang.Object-java.lang.Object-

(I also verified that `Pattern#toString` is a cheap operation, OpenJDK simply stores the string as returned by `Pattern#pattern`)

IkeVoodoo: Fixed the issue, I removed String.format

astei (approved):

PR #842, opened 2022-08-16
feat(plugin): auto updates

### LINKED ISSUE #809: Feature request: Add update folder mechanism similar to what Bukkit servers do
Before loading any plugins, check if an update folder has any Velocity plugins and replace plugin candidates with plugin ids that match the update files.

On my public server, we schedule proxy restarts to happen at 4 AM when players are less likely to be online. Since hot swapping plugins is not recommended, this would be the next best thing since it would allow me to drop the plugin updates in and forget about it.

AV3RG:
closes #809

TODO:

- [x] Rollback in case of a failure while updating
- [x] Let the users enable/disable auto updates using config or startup flags
- [x] Test changes


proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
```diff
@@ -330,18 +330,38 @@ private void loadPlugins() {
...
-          return;
-        }
+      }
+      if (!updatePath.toFile().exists()) {
+        Files.createDirectory(updatePath);
+      }
+      if (!outdatedPluginsPath.toFile().exists()) {
+        Files.createDirectory(outdatedPluginsPath);
+      }
+      if (!pluginPath.toFile().isDirectory()) {
```
4drian3d: Instead of using `toFile()` everywhere, I think it would be better to use methods of the [Files](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/file/Files.html) class since the type of object used is Path. You can replace `File#isDirectory` with [Files#isDirectory(Path)](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/file/Files.html#isDirectory(java.nio.file.Path,java.nio.file.LinkOption...)) and `File#exists` with [Files#exists(Path)](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/file/Files.html#exists(java.nio.file.Path,java.nio.file.LinkOption...))

proxy/src/main/resources/default-velocity.toml
```diff
@@ -141,6 +141,9 @@ log-command-executions = false
 # and disconnecting from the proxy.
 log-player-connections = true
 
+#Whether plugins should be auto updated/installed from the update folder
```
StealWonders: Missing space between # and comment

proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
```diff
@@ -330,18 +330,41 @@ private void loadPlugins() {
 
     try {
       Path pluginPath = Path.of("plugins");
+      Path updatePath = pluginPath.resolve("to-update");
+      Path outdatedPluginsPath = pluginPath.resolve("outdated");
 
-      if (!pluginPath.toFile().exists()) {
+      if (!Files.exists(pluginPath)) {
```
4drian3d: Maybe it would be better to use notExists instead of inverting the condition of the exists method https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/file/Files.html#notExists(java.nio.file.Path,java.nio.file.LinkOption...)

proxy/src/main/java/com/velocitypowered/proxy/plugin/VelocityPluginManager.java
```diff
@@ -101,8 +111,71 @@ public void loadPlugins(Path directory) throws IOException {
...
+            logger.error("Unable to load plugin candidate {}", path, e);
+          }
+        }
+      }
+
+      //match the to update plugin's id against already loaded plugins and replace them if match found
+      for (PluginDescription updatedDescription : updatesToApply) {
+        PluginDescription possibleMatch = found.get(updatedDescription.getId());
+        if (updatedDescription.getSource().isEmpty()) { //should not happen but just in case
+          logger.warn("");
```
4drian3d: Here you missed to put some warning message

proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
```diff
@@ -330,18 +330,41 @@ private void loadPlugins() {
...
-        if (!pluginPath.toFile().isDirectory()) {
-          logger.warn("Plugin location {} is not a directory, continuing without loading plugins",
-              pluginPath);
-          return;
-        }
+      }
+      if (!Files.exists(updatePath)) {
+        Files.createDirectory(updatePath);
+      }
+      if (!Files.exists(outdatedPluginsPath)) {
```
4drian3d: Same here

4drian3d (changes requested):

proxy/src/main/java/com/velocitypowered/proxy/plugin/VelocityPluginManager.java
AV3RG: oops

Noorquacker:
This would be a great feature :)

Joo200:
Is there any progress here?

The Merge conflicts can be resolved easily, i rebased the branch and fixed the checkstyles in this branch:
https://github.com/Joo200/Velocity/tree/feat/update-plugins

AV3RG:
Not too sure how I did not get a notification for pending conflicts in this pull request and how I passed the check-style last time

proxy/src/main/java/com/velocitypowered/proxy/plugin/VelocityPluginManager.java
```diff
@@ -105,8 +115,76 @@ public void loadPlugins(Path directory) throws IOException {
...
+        Path oldPluginPath = null;
+        if (possibleMatch != null) {
+          if (possibleMatch.getSource().isEmpty()) {
+            logger.warn("No source for plugin {} found, continuing without update.",
+                    possibleMatch.getId());
+            continue;
+          }
+          //move old plugin to outdated plugin directory to rollback in case of failure
+          oldPluginPath = possibleMatch.getSource().get();
+          Files.move(oldPluginPath, outdatedPluginDirectory.resolve(oldPluginPath.getFileName()));
```
Joo200: This can fail due to missing file permissions. It should be in a try-catch block.

Another issue with this move call: It doesn't overwrite existing files:
```
[23:05:24] [main/INFO] [com.velocitypowered.proxy.VelocityServer]: Loading plugins...
[23:05:24] [main/ERROR] [com.velocitypowered.proxy.VelocityServer]: Couldn't load plugins
java.nio.file.FileAlreadyExistsException: plugins/outdated/Punishments-Velocity.jar
	at sun.nio.fs.UnixCopyFile.move(UnixCopyFile.java:449) ~[?:?]
	at sun.nio.fs.UnixFileSystemProvider.move(UnixFileSystemProvider.java:266) ~[?:?]
	at java.nio.file.Files.move(Files.java:1432) ~[?:?]
	at com.velocitypowered.proxy.plugin.VelocityPluginManager.loadPlugins(VelocityPluginManager.java:156) ~[velocity-3.2.0-SNAPSHOT-2023-04-06.jar:3.2.0-SNAPSHOT (git-6c196c5c)]
	at com.velocitypowered.proxy.VelocityServer.loadPlugins(VelocityServer.java:368) ~[velocity-3.2.0-SNAPSHOT-2023-04-06.jar:3.2.0-SNAPSHOT (git-6c196c5c)]
	at com.velocitypowered.proxy.VelocityServer.start(VelocityServer.java:227) ~[velocity-3.2.0-SNAPSHOT-2023-04-06.jar:3.2.0-SNAPSHOT (git-6c196c5c)]
	at com.velocitypowered.proxy.Velocity.main(Velocity.java:69) ~[velocity-3.2.0-SNAPSHOT-2023-04-06.jar:3.2.0-SNAPSHOT (git-6c196c5c)]
[23:05:24] [main/INFO] [com.velocitypowered.proxy.VelocityServer]: Loaded 0 plugins
```
Using this should fix that issue:
```suggestion
          Files.move(oldPluginPath, outdatedPluginDirectory.resolve(oldPluginPath.getFileName()), java.nio.file.StandardCopyOption.REPLACE_EXISTING);
```

Noorquacker:
Hiiii,
Has any reviewer had a chance to look at the official code yet? It's been nearly a year in the making, but it's better late than never

Joo200:
I'm using this on my server and it works fine. The implementation and the code looks fine.

I'm not a reviewer here so I can't hit the approve button in github. But imo it should be ready to be merged.

Noorquacker:
I'm gonna do it:
@4drian3d @StealWonders can we get a review 👉🏾👈🏾

proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
```diff
@@ -332,18 +332,41 @@ private void loadPlugins() {
 
     try {
       Path pluginPath = Path.of("plugins");
+      Path updatePath = pluginPath.resolve("to-update");
```
4drian3d, line 417: Personally, I am not convinced by the name `to-update` for the plugins folder to be updated. Also, if it were up to me, I would integrate the folder inside the plugins folder, since having 3 folders in the main Velocity directory dedicated to contain plugins would be very strange

proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
```diff
@@ -332,18 +332,41 @@ private void loadPlugins() {
...
+        Files.createDirectory(updatePath);
+      }
+      if (Files.notExists(outdatedPluginsPath)) {
+        Files.createDirectory(outdatedPluginsPath);
+      }
+      if (!Files.isDirectory(pluginPath)) {
+        logger.warn("Plugin location {} is not a directory, continuing without loading plugins",
+                pluginPath);
+        return;
+      }
```
4drian3d, line 435: I would move this check just after the plugins folder creation, to avoid creating secondary plugins folders if the main folder is not available

proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
```diff
@@ -332,18 +332,41 @@ private void loadPlugins() {
...
-              pluginPath);
-          return;
-        }
+      }
+      if (Files.notExists(updatePath)) {
+        Files.createDirectory(updatePath);
+      }
+      if (Files.notExists(outdatedPluginsPath)) {
+        Files.createDirectory(outdatedPluginsPath);
+      }
```
4drian3d: The creation of plugin update folders could be moved inside the [`VelocityConfiguration#isAutoUpdate`](https://github.com/PaperMC/Velocity/pull/842/files#diff-1217e64751ab6522195a858620cceeb4f2d2b6506c36dd4e396726ad3e7ef0cbR354) check, to avoid creating these folders unnecessarily in case someone doesn't want this function

proxy/src/main/java/com/velocitypowered/proxy/plugin/VelocityPluginManager.java
```diff
@@ -105,8 +116,83 @@ public void loadPlugins(Path directory) throws IOException {
...
+            try {
+              PluginDescription rolledBackDescription = loader.loadCandidate(oldPluginPath);
+              found.put(rolledBackDescription.getId(), rolledBackDescription);
+            } catch (Exception ex) {
+              logger.error("Unable to load rollback plugin candidate {}", oldPluginPath, ex);
+            }
+          }
+        }
+      }
+    }
```
4drian3d: I think this block of code could be moved into a separate private method, to avoid overloading this method even more

proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
AV3RG: Im confused
Its already inside the plugin folder, I am doing `pluginPath.resolve()`

Noorquacker:
Erm
Any update?

proxy/src/main/java/com/velocitypowered/proxy/plugin/VelocityPluginManager.java
AV3RG: Should be resolved

proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
AV3RG: Should be resolved

proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
AV3RG: Done

AV3RG:
Kind of conflicted on the usage of `List` instead of `HashMap` for loaded candidates. Using HashMap would help in matching of plugin ids when trying to search for the possible updates here
`PluginDescription possibleMatch = found.get(updatedDescription.getId());`

Do we wanna use streams here or should we revert back to `HashMap`

4drian3d:
You have conflicts to resolve before requesting another review
image

carlodrift:
Updated from upstream, resolved conflicts, and tested the changes—everything seems to be working fine: https://github.com/AV3RG/Velocity/pull/1
Should I create a new PR?

AV3RG:
Thanks a lot @carlodrift. Was putting this off for a while.

carlodrift:
Heya @4drian3d @StealWonders!

We've resolved the conflicts and tested the changes—everything seems to be working fine. Could you take a look and let us know if anything else needs fixing? Appreciate your time!

proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
```diff
@@ -414,18 +414,50 @@ private void loadPlugins() {
 
     try {
       Path pluginPath = Path.of("plugins");
+      Path updatePath = pluginPath.resolve("to-update");
```
OakLoaf, line 417: I personally think that this should be a configurable directory name similar to how it is in Bukkit. For me this would allow naming to `.update` which would ensure it's always at the top of directories and easily distinguishable from other plugin folders.

OakLoaf:
I almost feel like a config option for this is somewhat obsolete? Surely if someone doesn't want the feature they just wouldn't put the jars in the update folder?

I may be missing something obvious regarding this though.

Noorquacker:
Agreed with @OakLoaf, unless there's some really strange circumstance, it seems unusual to want to disable the updater.
On top of this, why not follow the same convention as Paper and use `plugins/update` as the update folder? If it's going to be a config option then I'll just yield to that, but it seems a little unexpected to change the folder to `to-update` or something like that by default

Phoenix616:
> I almost feel like a config option for this is somewhat obsolete? Surely if someone doesn't want the feature they just wouldn't put the jars in the update folder?
> 
> I may be missing something obvious regarding this though.

I personally think a config option is good and that this functionality could even be disabled by default, as plugins might download files into that update folder without user-consent. Sure they could also just download files directly if they are malicious but I'm thinking about cases where someone needs to use an older version for some reason (e.g. compatibility or bugs in a new version) but the plugin tries to always auto-update. (Ideally the config would even allow turning it off on a per-plugin basis)

PR #843, opened 2022-08-16
Change max char size from 4 to 3

Outfluencer:
BungeeCord reference: https://github.com/SpigotMC/BungeeCord/pull/3377/commits/b5365dcba75f9350684cc30e87ef3a83f89d1326

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/protocol/ProtocolUtils.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/protocol/ProtocolUtils.java
@@ -185,9 +185,9 @@ public enum ProtocolUtils {
   private static String readString(ByteBuf buf, int cap, int length) {
     checkFrame(length >= 0, "Got a negative-length string (%s)", length);
     // `cap` is interpreted as a UTF-8 character length. To cover the full Unicode plane, we must
-    // consider the length of a UTF-8 character, which can be up to 4 bytes. We do an initial
+    // consider the length of a UTF-8 character, which can be up to 3 bytes. We do an initial
     // sanity check and then check again to make sure our optimistic guess was good.
-    checkFrame(length <= cap * 4, "Bad string size (got %s, maximum is %s)", length, cap);
+    checkFrame(length <= cap * 3, "Bad string size (got %s, maximum is %s)", length, cap);
     checkFrame(buf.isReadable(length),
         "Trying to read a string that is too long (wanted %s, only have %s)", length,
         buf.readableBytes());


proxy/src/main/java/com/velocitypowered/proxy/protocol/ProtocolUtils.java
```diff
@@ -185,9 +185,9 @@ public static String readString(ByteBuf buf, int cap) {
   private static String readString(ByteBuf buf, int cap, int length) {
     checkFrame(length >= 0, "Got a negative-length string (%s)", length);
     // `cap` is interpreted as a UTF-8 character length. To cover the full Unicode plane, we must
-    // consider the length of a UTF-8 character, which can be up to 4 bytes. We do an initial
+    // consider the length of a UTF-8 character, which can be up to 3 bytes. We do an initial
     // sanity check and then check again to make sure our optimistic guess was good.
-    checkFrame(length <= cap * 4, "Bad string size (got %s, maximum is %s)", length, cap);
+    checkFrame(length <= cap * 3, "Bad string size (got %s, maximum is %s)", length, cap);
```
Xernium, line 190: Deferred. Older versions still respect the old limit, this needs to respect that.

Xernium (changes requested):

proxy/src/main/java/com/velocitypowered/proxy/protocol/ProtocolUtils.java
Outfluencer: Yes, but you are wrong.

In java a utf8 char has a maximum limit of 3 bytes, so in old clients 4 bytes is used as limit but this limit is impossible to reach so we can decrease the limit to 3 bytes

For example if you use the highest valued char in a string and the length of the string is 1 the byte[] len of the string will be 3 or less. In every case! Otherwise it would be 2 chars and than the rule would be 2*3 = 6 so max 6 bytes.

You can test it

hugmanrique: [The Unicode Standard](https://www.unicode.org/versions/Unicode14.0.0/ch03.pdf#G7404) is quite clear in that UTF-8 encodes a scalar value as a byte sequence of 1 to 4 bytes of length (see point D92). Table 3-7 lists the well-formed UTF-8 byte sequences, among which those corresponding to code points U+10000..U+3FFFF, U+40000..U+FFFFF and U+100000..U+10FFFF are 4 bytes long. Even if the vanilla client doesn't produce code unit sequences of this length (which I haven't checked), a Unicode 8-bit string (as defined in point D81) of length greater than `3*#code points` units/bytes is totally valid. I would even consider this a bug in Mojang's implementation.

Outfluencer: No mojangs impl is right its just not possible in java. Instead of arguing with docs you should just test it lmao

electronicboy: Mojang uses the read/write utf8 methods for any varint length encoded strings, which checks the String#length(), and the length of getBytes(); given that the max size of a char is 3 bytes, this means that a string which is 255 in length, cannot be over 765, meanwhile we currently allow 1020; 

The whole UTF8 thing gets into a whole 'nother nightmare around terminology and handling of UTF8 in general just being a plague due to the variable length nature; what this gets to in java is that a codepoint which is 32 wide, is represented as 2 characters, for 5 bytes, to my understanding; (at least 5 bytes is the largest "valid" codepoint that I saw when screwing around)

IF the read/write calls weren't also backed behind a String#length check in the protocol, this would be a change which could actually break stuff, however, because it's impossible in java for a character to be over 3 bytes in length when encoded, this is basically a none breaking change

electronicboy (approved):

proxy/src/main/java/com/velocitypowered/proxy/protocol/ProtocolUtils.java
Outfluencer: Thank you

hugmanrique: I was just noting that this method is correct only because Java uses sixteen-bit UTF-16 code units (that is, sequences of `char`s) to encode strings, where supplementary characters (those outside of the BMP) are implemented using surrogate pairs. The `buf.toString(buf.readerIndex(), length, StandardCharsets.UTF_8)` call will happily read a Unicode string with more than `3*#code points` bytes, if it is valid. Adding this note to the javadoc wouldn't hurt.

I agree that there doesn't exist a non-Java client with enough popularity to warrant support for longer byte sequences.

hugmanrique: I was just noting that this method is no longer reading UTF-8 encoded strings, but strings encoded using Java's _[modified UTF-8 encoding](https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/io/DataInput.html)_, where only the 1,2 and 3-byte variants are used (and supplementary characters are encoded using surrogate pairs). Adding this to the javadoc wouldn't hurt: indeed, the `buf.toString(buf.readerIndex(), length, StandardCharsets.UTF_8)` call will happily read a UTF-8 string with more than `3*#codepoints` bytes, if it is valid.

I agree that there doesn't exist a non-Java client with enough popularity to warrant keeping the 4-byte limit, but as it stands, the Javadoc is technically incorrect.

PR #860, opened 2022-09-09
Resolve the closest locale only for Velocity translations

Yeregorix:
Currently, all translators registered in the `GlobalTranslator` resolve with a locale different from the player's actual locale. Velocity is modifying the locale so it matches one of the locales available for its own translations. Velocity is doing this globally which interfere with other translators which may want to know the original locale. This PR fixes this by resolving the closest locale only inside Velocity's translator.


proxy/src/main/java/com/velocitypowered/proxy/console/VelocityConsole.java
```diff
@@ -56,8 +55,7 @@ public final class VelocityConsole extends SimpleTerminalConsole implements Cons
   private PermissionFunction permissionFunction = ALWAYS_TRUE;
   private final @NotNull Pointers pointers = ConsoleCommandSource.super.pointers().toBuilder()
           .withDynamic(PermissionChecker.POINTER, this::getPermissionChecker)
-          .withDynamic(Identity.LOCALE, () -> ClosestLocaleMatcher.INSTANCE
-              .lookupClosest(Locale.getDefault()))
+          .withDynamic(Identity.LOCALE, Locale::getDefault)
```
4drian3d: I think it would have been ideal to keep this, since the default language and the language in which messages are sent to the console could vary, while the language defined in the JVM could always be checked with a `Locale.getDefault()`

4drian3d (approved):

proxy/src/main/java/com/velocitypowered/proxy/console/VelocityConsole.java
Yeregorix: Do you mean just in the `pointers` or in the `sendMessage` method too?

In the first case, I would find this weird to have a mismatch between these two.

In the second case, I think it the same problem as for players, translators will have no way to know the original locale of the console. Sure it is `Locale.getDefault()` but translators don't know if they are translating something for the console or for a player.

4drian3d: I am referring to the console pointers, since anyone can use `Locale.getDefault()`, but the language in which the messages are sent to the console may vary.
But it is something without much importance and it can be left as you have done it

4drian3d (approved):

PR #868, opened 2022-09-16
Use vanilla illegal_characters message

Pantera07:
Use vanilla illegal_characters message

```
"multiplayer.disconnect.illegal_characters": "Illegal characters in chat"
```


electronicboy:
I think that the general notion is to use velocities messages, as we can't guarantee that just because something speaks the MC protocol that such messages will exist

Pantera07:
You're right. Minecraft versions lower than 1.12 do not have this message.

I think the `multiplayer.disconnect.authservers_down` message also has this issue.
https://github.com/PaperMC/Velocity/blob/213f3f8864203cb1c001757621e407ecd3858d28/proxy/src/main/java/com/velocitypowered/proxy/connection/client/InitialLoginSessionHandler.java#L257-L261

Astralchroma:
Well the proxy knows what version the client is using, so why not use the vanilla translated message normally, and use velocities message when the version is too old?

PR #869, opened 2022-09-16
Bump netty to 4.1.82.Final

Pantera07:
Bump netty to 4.1.82.Final

https://netty.io/news/2022/09/08/4-1-81-Final.html
https://netty.io/news/2022/09/13/4-1-82-Final.html


electronicboy (approved):
Don't think that there is anything of care in here outside of maybe 1 or 2 small improvements, but, might as well

PR #871, opened 2022-09-20
Update Every Dam thingy

JT122406:


kyngs:
You've updated configurate from 3, to 4 which introduces API breaking changes.

EDIT: considering that velocity doesn't build at all, there most likely have been multiple breaking changes

PR #873, opened 2022-09-20
Update Dependencies to Latest

JT122406:
Yes I verified this builds on my system and shouldn't have any breaking changes mainly just security changes

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -12,7 +12,7 @@ allprojects {
         // dependency versions
         adventureVersion = '4.11.0'
         junitVersion = '5.9.0'
-        slf4jVersion = '2.0.1'
+        slf4jVersion = '2.0.2'
         log4jVersion = '2.19.0'
         nettyVersion = '4.1.82.Final'
         guavaVersion = '31.1-jre'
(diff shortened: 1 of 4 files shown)


JT122406:
This one builds on System and should be fine in WorkFlow

build.gradle
```diff
@@ -10,13 +10,13 @@ allprojects {
 
     ext {
         // dependency versions
-        adventureVersion = '4.10.1'
-        junitVersion = '5.7.0'
-        slf4jVersion = '1.7.30'
-        log4jVersion = '2.17.1'
+        adventureVersion = '4.11.0'
+        junitVersion = '5.9.0'
+        slf4jVersion = '2.0.1'
```
Kichura: SLF4J 2.0.2 is available by the way.

JT122406: ah ok thanks

astei:
A good chunk of these are breaking. I took only the ones not likely to break the public API and put them in 7d77bfb5

A248:
By the way, Slf4j 2.0 is binary compatible with respect to client code using Slf4j 1.7. Only the logging framework would need to be updated -- that is, `log4j-slf4j-impl` needs to support Slf4j 2.0. The Slf4j docs describe the changes in 2.0.x: https://www.slf4j.org/faq.html#changesInVersion200.

Something to consider perhaps for another time.

Issue #870, opened 2022-09-17
Optimize API/Events for TabList


Joo200:
### API Changes

Currently the changes from the backend servers are passed to the client without any event or api to interrupt/change stuff.
E.g. it's not possible to change the display name without hacks. It would be nice to have an event to modify the added or changed tablist item and to prevent the proxy from sending those changes.

### Another reason to change stuff there

Currently modifications at the player list will not affect the sent package. This may be a problem:
In VelocityTabList#processBackendPacket the key should be replaced by the public key when no key is present from the backend server. I don't know how this should happen but with the current implementation the key of the player will only be replaced on the server, the package is sent without any modification.

PR #874, opened 2022-09-21
Added ServerUpdateTablistEvent to allow modifications of the player list

Joo200:
This event is called when the backend server modifies the tablist entries. The entries can be modified but not removed or changed.

This merge request should fix my issue #870 .

Example code to use that event:
```java
@Subscribe
public void onUpdate(ServerUpdateTablistEvent event) {
  switch (event.getAction()) {
    case UPDATE_DISPLAY_NAME, ADD_PLAYER -> {
      event.getEntries().forEach(entry -> 
          entry.setDisplayName(Component.text(entry.getProfile().getName(), NamedTextColor.YELLOW)));
  }
}
```

It was necessary to change the logic and to not send the backend server's packet directly to the client. This should be no breaking change and it works with the enabled and disabled signing feature.

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/event/player/ServerUpdateTablistEvent.java
+++ b/api/src/main/java/com/velocitypowered/api/event/player/ServerUpdateTablistEvent.java
@@ -1,0 +1,101 @@
+/*
+ * Copyright (C) 2018 Velocity Contributors
+ *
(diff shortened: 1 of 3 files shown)


Joo200:
I converted that PR to a draft. With the recent 1.19.3 changes this is completly incompatible.

I'll update that PR soon.

justEli:
Would love to see this added! Is there any update on this?

PR #890, opened 2022-11-27
Pin libdeflate version

romanalexander:
Libdeflate changed to cmake which prevents compile-linux.sh from running as is. Instead of compiling against head, let's compile against a tag at least.


astei:
I think it'll probably be easier to just update the script to use the new build system...

PR #895, opened 2022-12-07
regression: fix legacy chat cancelling on < 1.19

roccodev:
After the 1.19.3 update, cancelling chat events would no longer work for legacy chat (messages would just go through).  
Additionally, uncancelled messages would appear twice.

Oddly enough, this fixes both issues.


CoreyShupe:
LGTM; must've missed that boolean flip; thank you

PR #897, opened 2022-12-07
chat: fix encoding for changed unsigned messages

roccodev:
I think you forgot to initialise `lastSeenMessages` when converting unsigned "session" (the new format from 1.19.3) chat messages for backend servers.  
Looking at the code flow, it appears that a NPE should always be triggered for unsigned messages that were changed.

I could reproduce it locally, but I apologise if this is something that can never happen in a real scenario, as I'm using a fork with chat signing disabled for legacy support.


CoreyShupe:
Your fix looks accurate to me.

PR #900, opened 2022-12-08
Fix command packet

### LINKED ISSUE #898: `forwardToServer` in the `CommandExecuteEvent` not executing commands for 1.19.3 clients
Whenever I try set the result to use CommandResult#forwardServer in the CommandExecuteEvent, it does not execute the user's command on a minecraft server by using the following code;
```
@Subscribe()
public void onCommand(CommandExecuteEvent event) {
    event.setResult(CommandExecuteEvent.CommandResult.forwardToServer(event.getCommand()));
}
```

This is because players from 1.19.3 and other players have different command handlers to handle their incoming commands.

Players from 1.19.3 get assigned to the SessionCommandHandler, which [uses](https://github.com/PaperMC/Velocity/blob/dev/3.0.0/proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/chat/session/SessionCommandHandler.java#L45) `if (!result.isAllowed())`
This checks if the result is not allowed, which forwarding commands are not, thus this command will never be executed.

All other command handlers [use](https://github.com/PaperMC/Velocity/blob/dev/3.0.0/proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/chat/keyed/KeyedCommandHandler.java#L48) `if (result == CommandExecuteEvent.CommandResult.denied())`
In this example, it checks if it matches the denied property itself, and since forwarding commands dont match these, they will be executed safely.

This probaby isn't intended behavior as it might break commands entirely for servers using this method.

Edit: fixed the redirects
Edit: This issue also affects the [LegacyCommandHandler](https://github.com/PaperMC/Velocity/blob/dev/3.0.0/proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/chat/legacy/LegacyCommandHandler.java#L46)

james58899:
Fix command packet wrong dropped `lastSeenMessages` causing chat chain break, also fix #898.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/chat/session/SessionCommandHandler.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/chat/session/SessionCommandHandler.java
@@ -43,7 +43,7 @@ public class SessionCommandHandler implements CommandHandler<SessionPlayerComman
     queueCommandResult(this.server, this.player, event -> {
       CommandExecuteEvent.CommandResult result = event.getResult();
       if (!result.isAllowed()) {
-        if (!packet.argumentSignatures.isEmpty()) {
+        if (packet.isSigned()) {
           logger.fatal("A plugin tried to deny a command with signable component(s). " + "This is not supported. "
               + "Disconnecting player " + player.getUsername());
           player.disconnect(Component.text(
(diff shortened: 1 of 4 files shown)


proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/chat/session/SessionCommandHandler.java
```diff
@@ -42,8 +42,8 @@ public Class<SessionPlayerCommand> packetClass() {
   public void handlePlayerCommandInternal(SessionPlayerCommand packet) {
     queueCommandResult(this.server, this.player, event -> {
       CommandExecuteEvent.CommandResult result = event.getResult();
-      if (!result.isAllowed()) {
-        if (!packet.argumentSignatures.isEmpty()) {
+      if (result == CommandExecuteEvent.CommandResult.denied()) {
```
electronicboy, line 45: Pretty sure that this check needs to stay as-is, at least in regards to 1.19.x existing for now

james58899: Forward but modifying the command has been handled below, error message is also deny.

electronicboy: ah, I erm... left my glasses somewhere around here

electronicboy (approved):

PR #901, opened 2022-12-09
Add ResourcePackResponse packet id for 1.19.3

JNNGL:


CoreyShupe:
https://github.com/ViaVersion/ViaVersion/blob/master/common/src/main/java/com/viaversion/viaversion/protocols/protocol1_19_3to1_19_1/ServerboundPackets1_19_3.java#L60

Looks like PackStatus is still 0x24; look at the missing comment for chat session & add 1.

Is there something happening which makes you think it's not 0x24?

PR #902, opened 2022-12-09
Reduce Spam from the TabList by not sending every package multiple times

Joo200:
VelocityTabList#processUpsert called entry.setX which will create a package and send it to the client. 

BackendPlaySessionHandler doesn't return true for those packages, therefore the package for tab list updates will be send two times.


CoreyShupe:
This all looks good to me

PR #921, opened 2023-01-03
Update Plugins and Repo

JT122406:


astei:
The repo issue is already known and I'm working with the Paper team to resolve that.

I'll do a pass to update the Gradle plugins later, but they're far from the biggest fish to fry.

PR #929, opened 2023-01-10
Handle deprecated custom bungeecord subchannels

EpicPlayerA10:
Some old plugins are still using `BungeeCord` channel for comunicating with proxy <-> subserver. I wanted to support these plugins on velocity but they can't be really updated to use their own channel.


astei:
Upstream BungeeCord has not supported this in a long while either. Closing since we're not interested in supporting ancient plugins.

astei:
We're generally not interested in bug-for-bug compatibility with BungeeCord

PR #930, opened 2023-01-13
Use TCP_NODELAY only when netty don't enable it by default

UserNugget:
By default netty already enable TCP_NODELAY, so enabling it twice leads to unnecessary syscall
https://github.com/netty/netty/blob/netty-4.1.86.Final/transport-classes-epoll/src/main/java/io/netty/channel/epoll/EpollSocketChannelConfig.java#L50-L52
https://github.com/netty/netty/blob/netty-4.1.86.Final/transport/src/main/java/io/netty/channel/socket/DefaultSocketChannelConfig.java#L50-L57


astei:
`PlatformDependent` is an internal class and I don't see the benefit of avoiding an additional system call here

PR #938, opened 2023-01-24
feat: Implement RegisteredServer#ping(PingOptions)

### LINKED ISSUE #937: RegisteredServer#ping does not support HEX colors
I was making a plugin to show to the admins of a server the motd of their servers, which uses the `RegisteredServer#ping` method, but when the ping is shown, it shows the colors as if they were shown to a Minecraft 1.15.2 or lower client.

This is because in the `#ping()` method the [ProtocolVersion.UNKNOWN is provided instead of any modern version](https://github.com/PaperMC/Velocity/blob/dev/3.0.0/proxy/src/main/java/com/velocitypowered/proxy/server/VelocityRegisteredServer.java#L88), which, when looking for the appropriate GsonSerializer for that version, the [Serializer for Legacy versions is used](https://github.com/PaperMC/Velocity/blob/f744b37ad5c7b6f00ca01681f26b75cad2584b7e/proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java#L750).

4drian3d:
Resolves #937 by implementing a new method to ping a server with a specified options

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/proxy/server/PingOptions.java
+++ b/api/src/main/java/com/velocitypowered/api/proxy/server/PingOptions.java
@@ -1,0 +1,170 @@
+/*
+ * Copyright (C) 2018-2023 Velocity Contributors
+ *
+ * The Velocity API is licensed under the terms of the MIT License. For more details,
+ * reference the LICENSE file in the api top-level directory.
+ */
+
+package com.velocitypowered.api.proxy.server;
+
+import com.velocitypowered.api.network.ProtocolVersion;
+import java.time.Duration;
+import java.util.Objects;
+import java.util.concurrent.TimeUnit;
+import net.kyori.adventure.builder.AbstractBuilder;
+import org.checkerframework.checker.nullness.qual.Nullable;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Represents the characteristics with which a ping to a Minecraft server is simulated.
+ * This class is immutable.
+ *
+ * @since 3.2.0
+ */
+public final class PingOptions {
+  private final ProtocolVersion protocolVersion;
+  private final long timeout;
+
+  private PingOptions(final Builder builder) {
+    this.protocolVersion = builder.protocolVersion;
+    this.timeout = builder.timeout.toMillis();
+  }
+
+  /**
+   * The version of Minecraft from which ping will be emulated.
+   *
+   * @return the emulated Minecraft version
+   */
+  public ProtocolVersion getProtocolVersion() {
+    return this.protocolVersion;
+  }
+
+  /**
+   * The maximum timeout duration for receiving the {@link ServerPing}.
+   *
+   * @return the server ping timeout in milliseconds
+   */
+  public long getTimeout() {
+    return this.timeout;
+  }
+
+  /**
+   * Create a new builder to assign values to a new PingOptions.
+   *
+   * @return a new {@link PingOptions.Builder}
+   */
+  public static Builder builder() {
+    return new Builder();
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (o == null) {
+      return false;
+    }
+    if (!(o instanceof PingOptions)) {
+      return false;
+    }
+    final PingOptions other = (PingOptions) o;
+    return Objects.equals(this.protocolVersion, other.protocolVersion)
+            && Objects.equals(this.timeout, other.timeout);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(this.protocolVersion, this.timeout);
+  }
+
+  @Override
+  public String toString() {
+    return "PingOptions{"
+            + "protocolVersion=" + protocolVersion
+            + ", timeout=" + timeout
+            + '}';
+  }
+
+  /**
+   * A builder for {@link PingOptions} objects.
+   *
+   * @since 3.2.0
+   */
+  public static final class Builder implements AbstractBuilder<PingOptions> {
+    private ProtocolVersion protocolVersion;
+    private Duration timeout;
+
+    private Builder() {
+    }
+
+    /**
+     * The protocol version used to ping the server
+     * corresponding to this Builder.
+     *
+     * @return the protocol version corresponding to this Builder
+     */
+    public @Nullable ProtocolVersion getProtocolVersion() {
+      return this.protocolVersion;
+    }
+
+    /**
+     * The maximum timeout duration to wait for the ping
+     * corresponding to this Builder.
+     *
+     * @return the timeout corresponding to this Builder
+     */
+    public @Nullable Duration getTimeout() {
+      return this.timeout;
+    }
+
+    /**
+     * Sets the protocol with which the server is to be pinged.
+     *
+     * @param protocolVersion the specified protocol
+     * @return this builder
+     */
+    public Builder version(final ProtocolVersion protocolVersion) {
+      this.protocolVersion = protocolVersion;
+      return this;
+    }
+
+    /**
+     * Sets the maximum time to wait to get the required {@link ServerPing}.
+     *
+     * @param timeout the timeout duration
+     * @return this builder
+     */
+    public Builder timeout(final @NotNull Duration timeout) {
+      this.timeout = timeout;
+      return this;
+    }
+
+    /**
+     * Sets the maximum time to wait to get the required {@link ServerPing}.
+     *
+     * @param time the timeout duration
+     * @param unit the unit of time to be used to provide the timeout duration
+     * @return this builder
+     */
+    public Builder timeout(final long time, final @NotNull TimeUnit unit) {
+      this.timeout = Duration.of(time, unit.toChronoUnit());
+      return this;
+    }
+
+    /**
+     * Create a new {@link PingOptions} with the values of this Builder.
+     *
+     * @return a new PingOptions object
+     * @throws NullPointerException if the timeout is not assigned
+     *     until this method is executed
+     */
+    @Override
+    public @NotNull PingOptions build() {
+      if (timeout == null) {
+        throw new NullPointerException("TimeOut cannot be null");
+      }
+      if (protocolVersion == null) {
+        protocolVersion = ProtocolVersion.UNKNOWN;
+      }
+      return new PingOptions(this);
+    }
+  }
+}
--- a/api/src/main/java/com/velocitypowered/api/proxy/server/PingOptions.java
+++ b/api/src/main/java/com/velocitypowered/api/proxy/server/PingOptions.java
@@ -7,7 +7,6 @@
 
 package com.velocitypowered.api.proxy.server;
 
-import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.velocitypowered.api.network.ProtocolVersion;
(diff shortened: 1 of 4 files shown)


astei:
I'm not a huge fan of this - there's a lot of variables that can be tweaked for pinging a Minecraft server. I'd rather see a builder here, perhaps having an overload of `ping()` that takes in a `ServerPingOptions` that's constructed with a builder?

4drian3d:
> I'm not a huge fan of this - there's a lot of variables that can be tweaked for pinging a Minecraft server. I'd rather see a builder here, perhaps having an overload of `ping()` that takes in a `ServerPingOptions` that's constructed with a builder?

Personally I consider that `PingOptions` would be better because `ServerPingOptions` may generate confusion about the connection with `ServerPing` and its possible need to introduce this new class within the ServerPing class, although I am not sure of this name either.
The only additional field that I found useful to add was the timeout, otherwise, I don't know what other variable could be added.

On the topic of javadocs, I do not speak English natively, if someone can give a suggestion of any change, it would be perfect.

api/src/main/java/com/velocitypowered/api/proxy/server/PingOptions.java
```diff
@@ -0,0 +1,170 @@
...
+import com.velocitypowered.api.network.ProtocolVersion;
+import java.time.Duration;
+import java.util.Objects;
+import java.util.concurrent.TimeUnit;
+import net.kyori.adventure.builder.AbstractBuilder;
+import org.checkerframework.checker.nullness.qual.Nullable;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Represents the characteristics with which a ping to a Minecraft server is simulated.
```
hugmanrique: "Contains the parameters used to ping a `RegisteredServer`". A `@see` tag to the `RegisteredServer#ping` method would be useful.

api/src/main/java/com/velocitypowered/api/proxy/server/PingOptions.java
```diff
@@ -0,0 +1,170 @@
...
+  private final ProtocolVersion protocolVersion;
+  private final long timeout;
+
+  private PingOptions(final Builder builder) {
+    this.protocolVersion = builder.protocolVersion;
+    this.timeout = builder.timeout.toMillis();
+  }
+
+  /**
+   * The version of Minecraft from which ping will be emulated.
```
hugmanrique: "The protocol version used to ping the server", I would avoid mentioning emulation.

api/src/main/java/com/velocitypowered/api/proxy/server/PingOptions.java
```diff
@@ -0,0 +1,170 @@
...
+   *
+   * @return the emulated Minecraft version
+   */
+  public ProtocolVersion getProtocolVersion() {
+    return this.protocolVersion;
+  }
+
+  /**
+   * The maximum timeout duration for receiving the {@link ServerPing}.
```
hugmanrique: "The maximum period of time to wait for a response from the remote server"

api/src/main/java/com/velocitypowered/api/proxy/server/PingOptions.java
```diff
@@ -0,0 +1,170 @@
...
+    private Builder() {
+    }
+
+    /**
+     * The protocol version used to ping the server
+     * corresponding to this Builder.
+     *
+     * @return the protocol version corresponding to this Builder
+     */
+    public @Nullable ProtocolVersion getProtocolVersion() {
```
hugmanrique: Imo builder classes don't need getter methods

api/src/main/java/com/velocitypowered/api/proxy/server/PingOptions.java
```diff
@@ -0,0 +1,170 @@
...
+    /**
+     * Create a new {@link PingOptions} with the values of this Builder.
+     *
+     * @return a new PingOptions object
+     * @throws NullPointerException if the timeout is not assigned
+     *     until this method is executed
+     */
+    @Override
+    public @NotNull PingOptions build() {
+      if (timeout == null) {
```
hugmanrique: We need to provide a default timeout value instead of throwing here. We can use 0 (see below).

api/src/main/java/com/velocitypowered/api/proxy/server/PingOptions.java
```diff
@@ -0,0 +1,170 @@
...
+
+    /**
+     * Sets the maximum time to wait to get the required {@link ServerPing}.
+     *
+     * @param time the timeout duration
+     * @param unit the unit of time to be used to provide the timeout duration
+     * @return this builder
+     */
+    public Builder timeout(final long time, final @NotNull TimeUnit unit) {
+      this.timeout = Duration.of(time, unit.toChronoUnit());
```
hugmanrique: Check `time` is positive or `-1` for no timeout; durations can be negative. Perhaps we can store the duration as milliseconds directly in the builder as well to simplify the `-1` case (there's no concept of an unlimited `Duration`).

proxy/src/main/java/com/velocitypowered/proxy/server/VelocityRegisteredServer.java
```diff
@@ -83,20 +84,27 @@ public Collection<Player> getPlayersConnected() {
...
    *
    * @param loop    the event loop to use
-   * @param version the version to report
+   * @param timeout the timeout respective
+   * @param protocolVersion the version to report
    * @return the server list ping response
    */
-  public CompletableFuture<ServerPing> ping(@Nullable EventLoop loop, ProtocolVersion version) {
+  public CompletableFuture<ServerPing> ping(
+          @Nullable EventLoop loop, long timeout, ProtocolVersion protocolVersion) {
```
hugmanrique: To clean this a bit, we can make the no-arguments `ping` method pass a default `PingOptions` instance (which can be cached and be exposed as an static member called `DEFAULT` in that class) to this method, and remove the method with all options below (which won't "scale" as the option set grows in the future).

proxy/src/main/java/com/velocitypowered/proxy/server/VelocityRegisteredServer.java
```diff
@@ -108,7 +116,8 @@ protected void initChannel(Channel ch) throws Exception {
             ch.pipeline()
                 .addLast(FRAME_DECODER, new MinecraftVarintFrameDecoder())
                 .addLast(READ_TIMEOUT,
-                    new ReadTimeoutHandler(server.getConfiguration().getReadTimeout(),
+                    new ReadTimeoutHandler(timeout < 0
```
hugmanrique: -1 usually means "no timeout", could we use 0 for the default value in the `PingOptions` class?

api/src/main/java/com/velocitypowered/api/proxy/server/PingOptions.java
```diff
@@ -0,0 +1,170 @@
...
+
+  /**
+   * The maximum timeout duration for receiving the {@link ServerPing}.
+   *
+   * @return the server ping timeout in milliseconds
+   */
+  public long getTimeout() {
+    return this.timeout;
+  }
+
```
hugmanrique, line 56: Some users prefer the convention `PingOptions.builder().[...].build().ping(server)` instead of `server.ping(PingOptions.builder().[...].build())`. Can we add that helper method that forwards the call to the existing ping member in `RegisteredServer`? I must say this is a big nit, however.

hugmanrique (commented):
Made some suggestions regarding the javadocs and some methods. They are pretty minor, everything else LGTM.

api/src/main/java/com/velocitypowered/api/proxy/server/PingOptions.java
4drian3d: Initially I added them because I was implementing the `PingOptions#asBuilder` method, _similar to `ServerPing#asBuilder`_. But I think in this class it won't be necessary for the moment.

api/src/main/java/com/velocitypowered/api/proxy/server/PingOptions.java
4drian3d: I'm not so sure about adding that method, I don't see much need to add it, besides that in Velocity there is no class that is built by builder that behaves that way

hugmanrique: I agree :) Let's ignore it then.

proxy/src/main/java/com/velocitypowered/proxy/server/VelocityRegisteredServer.java
4drian3d: I'm not sure how good it would be to leave the option of no timeout, although I'm not an expert on the subject to be able to give an opinion on it

api/src/main/java/com/velocitypowered/api/proxy/server/PingOptions.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    /**
+     * Sets the protocol with which the server is to be pinged.
+     *
+     * @param protocolVersion the specified protocol
+     * @return this builder
+     */
+    public Builder version(final ProtocolVersion protocolVersion) {
+      this.protocolVersion = protocolVersion;
```
hugmanrique, line 112: Check for null here and initialize the version class field to `UNKNOWN`; that way we don't need the null check in the `build` method.

api/src/main/java/com/velocitypowered/api/proxy/server/PingOptions.java
```diff
@@ -0,0 +1,157 @@
...
+ * Contains the parameters used to ping a {@link RegisteredServer}.
+ * This class is immutable.
+ *
+ * @since 3.2.0
+ * @see RegisteredServer#ping(PingOptions)
+ */
+public final class PingOptions {
+  /**
+   * Default PingOptions.
+   * Uses a {@link ProtocolVersion#UNKNOWN} version and a timeout of 3000.
```
hugmanrique: Let's not document the specific default timeout, as it may be changed by the user.

hugmanrique (commented):
Some last changes, thank you.

api/src/main/java/com/velocitypowered/api/proxy/server/PingOptions.java
hugmanrique: I agree, let's keep it this way :)

proxy/src/main/java/com/velocitypowered/proxy/server/VelocityRegisteredServer.java
hugmanrique: In the end, some effective timeout is placed by the network equipment. I was noting that the value -1 is usually used to mean "no limit", whereas 0 sometimes means "use a sane default".

4drian3d: Ok, I removed the limits for the timeout and documented the meaning of a value of 0 and a negative value

api/src/main/java/com/velocitypowered/api/proxy/server/PingOptions.java
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    /**
+     * Sets the protocol with which the server is to be pinged.
+     *
+     * @param protocolVersion the specified protocol
+     * @return this builder
+     */
+    public Builder version(final @NotNull ProtocolVersion protocolVersion) {
+      checkNotNull(protocolVersion, "protocolVersion cannot be negative");
```
hugmanrique: This message needs to be updated, everything else LGTM! 😀

4drian3d: Fixed 😅

hugmanrique:
For future reference (and curiosity), Netty's `ReadTimeoutHandler` uses [`IdleStateHandler`](https://github.com/netty/netty/blob/4.1/handler/src/main/java/io/netty/handler/timeout/IdleStateHandler.java) internally. While not documented, passing a negative value to the timeout handler thus disables the timeout logic (see the constructors of the latter class, which while their javadoc only specifies this is the case for zero, apply the same logic for negative values). This is why we can use a negative integer as the "no timeout" value in `PingOptions`, and handle the zero case separately.

hugmanrique (approved):

astei (approved):

PR #943, opened 2023-02-03
Implement a command and chat rate limit

zPirroZ3007:
## Abstract

In the last weeks [this exploit](https://github.com/LuckPerms/LuckPerms/issues/3546) affected a lot of proxies with great suffering from a lot of Minecraft server owners.

The exploit just works by creating a custom client that floods "heavy" commands in order to flood the proxy with tasks to crash it.

## The solution

I've implemented a simple command + chat rate limiter that works just like the Minecraft server does. If a user sends more than 10 commands in less than 4 seconds, gets kicked.

This fix will keep sure that things like this will never happena again

Clear and simple.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
@@ -550,23 +550,27 @@ public class ConnectedPlayer implements MinecraftConnectionAssociation, Player, 
   }
 
   @Override
-  public boolean chatRateLimit() {
-    boolean allowed = limiter.attempt(this);
-
-    if (!allowed) {
-      disconnect(Component.translatable("disconnect.spam", NamedTextColor.RED));
-    }
-
-    return allowed;
-  }
-
-  @Override
   public void disconnect(Component reason) {
     if (connection.eventLoop().inEventLoop()) {
       disconnect0(reason, false);
     } else {
       connection.eventLoop().execute(() -> disconnect0(reason, false));
     }
+  }
+
+  /**
+   * Checks and kicks the player when chat rate limit threshold has been reached.
+   *
+   * @return whether the message should be processed or not
+   */
+  public boolean isChatRateLimited() {
+    boolean allowed = limiter.attempt(this);
+
+    if (!allowed) {
+      disconnect(Component.translatable("disconnect.spam", NamedTextColor.RED));
+    }
+
+    return !allowed;
   }
 
   /**
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
@@ -78,6 +78,8 @@ import com.velocitypowered.proxy.tablist.VelocityTabListLegacy;
 import com.velocitypowered.proxy.util.ClosestLocaleMatcher;
 import com.velocitypowered.proxy.util.DurationUtils;
 import com.velocitypowered.proxy.util.collect.CappedSet;
+import com.velocitypowered.proxy.util.ratelimit.Ratelimiter;
+import com.velocitypowered.proxy.util.ratelimit.Ratelimiters;
 import io.netty.buffer.ByteBufUtil;
 import io.netty.buffer.Unpooled;
 import java.net.InetSocketAddress;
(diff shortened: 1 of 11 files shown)


proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
```diff
@@ -135,6 +137,8 @@ public class ConnectedPlayer implements MinecraftConnectionAssociation, Player,
 
   private static final Logger logger = LogManager.getLogger(ConnectedPlayer.class);
 
+  private static final Ratelimiter<Player> limiter = Ratelimiters.createWithMaxHitLimit(4000, 10);
```
4drian3d, line 140: Personally I think that putting 4 seconds would break some uses. In my experience, for example, in servers with protections you can get to use several commands quickly when claiming terrain or configuring some things. For that kind of cases, I think it would be better to add an argument when starting velocity to be able to customize this ratelimit or a new configuration option

zPirroZ3007: It's 10 messages/commands in 4 seconds. The Minecraft server uses this same limit. I will work on adding a config option. I'm also testing this feature in production with around 300 online users

xism4: > 

Players with lag will always be thrown off by this feature, I think it is better that the native velocity commands (either /velocity) be cached.

zPirroZ3007: > > 
> 
> Players with lag will always be thrown off by this feature, I think it is better that the native velocity commands (either /velocity) be cached.

@xIsm4 I'm testing this feature on my personal server, `play.tecnocraft.net` with around 300 online users averagely. I haven't noticed any complaint yet.

Also, what do you mean by caching commands?

Emibergo02:
@astei this is crucial. fixes a huge vunerability

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
astei: _This is an obvious memory leak_. For that reason alone, I'm not merging this straight away.

I'd also want to see the ability to have a customizable rate limit.

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
```diff
@@ -554,6 +558,21 @@ public void disconnect(Component reason) {
...
+   */
+  public boolean isChatRateLimited() {
+    boolean allowed = limiter.attempt(this);
+
+    if (!allowed) {
+      disconnect(Component.translatable("disconnect.spam", NamedTextColor.RED));
+    }
+
+    return !allowed;
+  }
```
astei, line 574: This is extremely deceptive - a function that has a non-obvious side effect when called. I'd rather have the disconnect moved into the sites where `isChatRateLimited()` is called.

proxy/src/main/java/com/velocitypowered/proxy/util/ratelimit/CaffeineCacheRatelimiter.java
```diff
@@ -22,15 +22,14 @@
...
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
-import java.net.InetAddress;
 import java.util.concurrent.TimeUnit;
 
 /**
  * A simple rate-limiter based on a Caffeine {@link Cache}.
  */
-public class CaffeineCacheRatelimiter implements Ratelimiter {
+public class CaffeineCacheRatelimiter<T> implements Ratelimiter<T> {
```
astei, line 30: Isn't this the same as `MaxHitRatelimiter` with `maxLimit = 1`? Why not combine the two?

proxy/src/main/java/com/velocitypowered/proxy/util/ratelimit/MaxHitRatelimiter.java
```diff
@@ -0,0 +1,66 @@
...
+import com.google.common.base.Preconditions;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * A simple rate-limiter based on a Caffeine {@link Cache}.
+ * This particular version limits the number of actions
+ * that can be accomplished in a limited amount of time
+ */
+public class MaxHitRatelimiter<T> implements Ratelimiter<T> {
```
astei, line 33: I'd like to see a unit test for this with various scenarios.

astei (changes requested):

4drian3d:
> updates on this?

There is no need to bump a pull request if there is no activity. If you are having any related problems, update to the latest version of Velocity, a change has been implemented that may help solve this problem

zPirroZ3007:
Closing this because has already been fixed with better methods by the Velocity team.

PR #945, opened 2023-02-04
Added API support for modifying command suggestions

Aseeef:
This pull request adds a new event called 'CommandTabCompleteEvent'. Currently, velocity does not support modifying tab complete suggestions for command arguments.

For example, if you attempt to Tab complete the following command, no event will be fired:
`/ban Asee`

**Motivation**
I have a multi-proxy setup using RedisBungee. Naturally, most plugins do not support such setups. Tab completions for commands for plugins on velocity do not return the full list of players on the network. Particularly, if I am attempting to ban a player 'Aseef' on the server, but, the player is on a different proxy than me, then the tab completion for the command `/ban Asee` will not autocomplete `Asee` to `Aseef`.

This issue creates the need to allow me to modify command tab completions for other plugins via my own plugins. And that precisely is what this new 'CommandTabCompleteEvent' attempts to solve.

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/event/player/CommandTabCompleteEvent.java
+++ b/api/src/main/java/com/velocitypowered/api/event/player/CommandTabCompleteEvent.java
@@ -1,0 +1,108 @@
+package com.velocitypowered.api.event.player;
+
+import com.velocitypowered.api.event.annotation.AwaitingEvent;
+import com.velocitypowered.api.proxy.Player;
+import net.kyori.adventure.text.Component;
+
+import java.util.List;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * This event is fired after a command tab complete response is sent by the remote server. The
+ * plugin that owns this command then suggests tab completions. Here, this event provides you the
+ * opportunity to modify the command tab completions suggested by the plugin. Velocity will wait
+ * for this event to finish firing before sending the tab complete results to the client. Be sure
+ * to be as fast as possible, since the client will freeze while it waits for the tab complete
+ * results.
+ */
+@AwaitingEvent
+public class CommandTabCompleteEvent {
+    private final Player player;
+    private final String command;
+    private List<CommandSuggestion> suggestions;
+
+    /**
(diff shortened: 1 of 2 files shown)


Xernium:
There’s a good reason we don’t have this. Command suggestions stopped working this way in 1.13 and thus if you need to provide completions you must register them correctly in your commands. Anything else is horrifically bad practice and should not be supported. If you need to do this please register a command that overwrites/overloads the command you’re trying to modify instead.

hugmanrique:
Or, arguably less of a hack (but still a hack), modify the command tree using the [`PlayerAvailableCommandsEvent`](https://jd.papermc.io/velocity/3.0.0/com/velocitypowered/api/event/command/PlayerAvailableCommandsEvent.html) to insert "dummy" nodes that provide the desired suggestions while redirecting execution to the original command. See the hinting system in Velocity (specifically https://github.com/PaperMC/Velocity/blob/8761d02def69d9c2ed8e026ba348a7468a4c6d7b/proxy/src/main/java/com/velocitypowered/proxy/command/VelocityCommandMeta.java#L103 and https://github.com/PaperMC/Velocity/blob/8761d02def69d9c2ed8e026ba348a7468a4c6d7b/proxy/src/main/java/com/velocitypowered/proxy/command/registrar/InvocableCommandRegistrar.java#L118) for inspiration. If you follow the same structure, you can fake the insertion of hinting nodes; but be advised you're depending on implementation details.

Aseeef:
> There’s a good reason we don’t have this. Command suggestions stopped working this way in 1.13 and thus if you need to provide completions you must register them correctly in your commands. Anything else is horrifically bad practice and should not be supported. If you need to do this please register a command that overwrites/overloads the command you’re trying to modify instead.

I actually tested this on a 1.16.4 and command suggestions were working perfectly fine.

Also, in my case overloading is not an option because the source plugin is obfuscated.

I'll likely just maintain my own fork of velocity.

hugmanrique:
What Five meant is that `TabCompleteEvent` is only fired for specific argument nodes on 1.16. For example, the client will not notify the server when providing suggestions for the partial input `/give `.

PR #948, opened 2023-02-14
Fix for edgecase FML|HS packet desynchronization

DEv0on:
Fixes edgecase when JoinGame packet arrives to Velocity before FMLHandshakeClientState.COMPLETE arrives from client, leading to infinite state of LegacyForgeHandshakeClientPhase.PENDING_COMPLETE at ConnectedPlayer#connectionPhase and queuing packets from client.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientPlaySessionHandler.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientPlaySessionHandler.java
@@ -277,7 +277,11 @@ public class ClientPlaySessionHandler implements MinecraftSessionHandler {
 
   @Override
   public boolean handle(PluginMessage packet) {
-    VelocityServerConnection serverConn = player.getConnectedServer();
+    // Handling edgecase when packet with FML client handshake (state COMPLETE)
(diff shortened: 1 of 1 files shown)


proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientPlaySessionHandler.java
```diff
@@ -277,7 +277,11 @@ public boolean handle(TabCompleteRequest packet) {
 
   @Override
   public boolean handle(PluginMessage packet) {
-    VelocityServerConnection serverConn = player.getConnectedServer();
+    // Handling edgecase when packet with FML client handshake (state COMPLETE)
+    // arrives after JoinGame packet from destination server
+    VelocityServerConnection serverConn = player.getConnectedServer() != null
+            ? player.getConnectedServer() : player.getConnectionInFlight();
```
astei: I'm not a big fan of this approach. We should limit this to if we're handling a Forge message specifically, or we're handling a Forge client (as server switch failures with Forge are considered fatal once we send the reset packet to the client).

astei (changes requested):

astei (approved):

PR #957, opened 2023-02-27
Allow underscore in hostname

ohAleee:


Xernium:
Only letters, digits and hyphens are allowed under [RFC1035](https://www.rfc-editor.org/rfc/rfc1035) in public domain names, thus this feature won’t be considered.

ohAleee:
it is useful with docker...

Xernium:
That doesn't change the fact that you're not only breaking api spec but also Java spec with this. Plus this opens up the possibility of resource abuse from the outside.

clrxbl:
> it is useful with docker...

There are so many workarounds to this and it is unlikely to actually be useful with Docker, name your containers properly and avoid underscores.

PR #972, opened 2023-03-11
Expose Caffeine in the API module

### LINKED ISSUE #971: Feature Request - Use Caffeine instead of Google Guava
Caffeine has extensions for features like JCache and some more features Guava has, also you get a much less memory overhead.
https://github.com/ben-manes/caffeine

4drian3d:
closes #971 

Currently Caffeine is used internally instead of Guava in the Cache topic, but it is not exposed in the API.

As of commit d72d707b1c0dd6215a4984956faf4b2ec4bdbd97 the API module is compiled with Java 11, so Caffeine can now be exposed in the API as an alternative to Guava with better performance


4drian3d:
The build failed because the SLF4J javadocs were down

kashike (approved):

PR #974, opened 2023-03-13
Igzip native compressor

JNNGL:
Replaces libdeflate native compression with igzip which is almost 3x faster (tested on i9-9900k: libdeflate - 102 mbps | igzip - 303 mbps)


rafalohaki:
588315715app_icon

SkyWatcher2019:
On my server  too.

astei:
Comments to the effect of “+1, please merge” are not going to get this merged any faster.

I will look at this when I have the time to do so.

ebiggers:
Which compression level is being used with libdeflate, and which level with igzip?  AFAICS, `VelocityConfiguration.java` has `compressionLevel = -1` which eventually gets translated into level 6 for libdeflate, and level 2 for igzip.  It doesn't make sense to do a comparison with these compression levels, as they are targeting vastly different time/space tradeoffs.

igzip's levels are also scaled a bit differently than libdeflate's (and zlib's) levels.  All the igzip levels target fast compression with a poor compression ratio.  They should all be compared to libdeflate level 1.  In my tests, libdeflate level 1 compresses more than any igzip level, and its speed is usually in between that of igzip level 2 and 3.

I believe the only use case for igzip here would be is if libdeflate **level 1** is not fast enough, in which case igzip level 1 or 2 could be used.  But it is not clear that libdeflate level 1 has been evaluated here.

astei:
Quoting the official documentation about the ISA-L deflate levels:

> New compression levels added. ISA-L fast deflate now has more levels to balance speed vs. target compression level. Level 0, 1 are as in previous generations. New levels 2 & 3 target higher compression roughly comparable to zlib levels 2-3.

This maps levels 0-1 to level 0, levels 2-6 to level 1, level 7 to level 2, and levels 8-10 to level 3. One bug in this PR is that there are no levels for 11 and 12, which libdeflate (and Velocity) supports.

Furthermore, if your goal is to reduce CPU usage at the cost of additional bandwidth, the compression level and minimum packet compression threshold can already be modified in Velocity. The default `compression-threshold` is 256 - I have long believed 1024 is a more sensible setting (ideally the setting should be the minimum MTU of your clients, less some, but 1024 will get you a lot of mileage, and it's a nice round power of 2). Finally, the `compression-level` can be tweaked in `velocity.toml`. Setting it to 1 should provide the fastest possible compression.

(Incidentally, @ebiggers, I know about a configuration where the developers are very happy with level 12 compression, in what is essentially a near-realtime compression scenario! That's not something I would have expected.)

PR #977, opened 2023-03-15
Bump netty to 4.1.90.Final

Pantera07:
https://netty.io/news/2023/03/14/4-1-90-Final.html
> We are happy to announce the release of netty 4.1.90.Final. This is a bug-fix release, but also includes some performance improvements. Please consider upgrading as soon as possible.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -10,7 +10,7 @@ guiceVersion=5.1.0
 
 # Proxy dependencies.
 log4jVersion=2.19.0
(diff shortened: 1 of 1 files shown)


4drian3d (approved):

astei:
Rolled up into 7e9a2520.

In the future, I would suggest not updating Netty each time a new release comes out. Most Netty releases don't have much value except for the lowest-level changes, and these are often fairly minor changes that don't move the needle.

PR #983, opened 2023-03-26
Move dependencies to gradle version catalog and clean up buildscripts

alexstaeding:
As a follow-up to #918 (and #518), I've opened this pull request to continue cleaning up the gradle scripts.

I've first primarily targeted the dependency declarations, opting to move them to a central version catalog. There are also some other minor fixes/improvements I made.

In summary:
- Moved dependencies to a central version catalog and referenced the type-safe accessors in each buildscript instead.
- Cleaned up some whitespace to be in line with the official Kotlin coding conventions. Most notably the continuation indent (see https://kotlinlang.org/docs/code-style-migration-guide.html#differences-between-kotlin-coding-conventions-and-intellij-idea-default-code-style)
- Moved the version declaration to a separate "version" file. This may be a bit of a personal preference of mine, but it has two advantages:
	- It is more obvious where the version is stored, instead of having it in build logic
	- It is slightly more git-friendly, as each release requires two commits (release, and bump) that end up pushing the entire file twice into version control (https://github.blog/2020-12-17-commits-are-snapshots-not-diffs/)
	
I'm open to feedback. My next target would be migrating `buildSrc` to `build-logic` - I can do it either here on in a separate PR.


build.gradle.kts
```diff
@@ -6,11 +6,9 @@ plugins {
     `java-library`
 }
 
-val junitVersion: String by project.extra
-
 allprojects {
     group = "com.velocitypowered"
-    version = "3.2.0-SNAPSHOT"
+    version = rootProject.file("version").readLines().first()
```
4drian3d: Instead of declaring the version in a separate file, I think it would be better to declare it in the `gradle.properties` file, that way you avoid creating a separate file and you also avoid this declaration of allprojects because by placing the version in the `gradle.properties` file, these values are shared throughout the project when using `project.version`

alexstaeding: Resolved in 39fc2805fd396d7a66bf0aa3444137714c7e7c7b

kashike:
It would also be good to fully use [indra](https://github.com/KyoriPowered/indra/wiki), and not just the publishing plugin - something you might want to do for another PR?

alexstaeding:
Yes, that's something I could revisit in a future PR. Just want to avoid getting roadblocked by smaller things when there are other fixes (such as this) that can already be merged.

4drian3d (approved):

kashike (approved):

PR #984, opened 2023-03-26
feat: added event when a command is executed from velocity

### LINKED ISSUE #954: No event for executed commands on proxy
I want to create a plugin to log commands usage. The plugin is a multi module project so commands on backend servers will be logged on the backend server:

- Command /foo: executed from a plugin on velocity -> my velocity plugin will log that execution
- Command /bar: executed on the backend -> my bukkit plugin will log (with coordinates)

I can't log every command on the proxy because I need current coordinates and worlds from the backend.

# Issue

It's currently not possible to listen to an event for that.

## Event CommandExecuteEvent:

This event is currently fired when a player is sending a command. `CommandExecuteEvent.CommandResult` has a method to check for server forwarding (isForwardingToServer()) but that's only true when some other plugin is forcing this command to the backend server.

After the event call Velocity will try to execute the command and forward it to the backend when no command could be found. That's not covered by any event and can't be changed.

For my problem I would have to check for existing commands before logging the issued command. Imo this is not really clean and not helpful.

# Solution

Fix the existing event and make `isForwardingToServer()` reliable for that. (preset the value before creating the event)

Joo200:
This PR will add a new event to allow plugins to monitor executed events from Velocity.

Based on the current implementation there is no easy way to make the CommandResult of CommandExecuteEvent reliable.

This fixes https://github.com/PaperMC/Velocity/issues/954


proxy/src/main/java/com/velocitypowered/proxy/command/VelocityCommandManager.java
```diff
@@ -210,7 +211,11 @@ private boolean executeImmediately0(final CommandSource source, final String cmd
     try {
       // The parse can fail if the requirement predicates throw
       final ParseResults<CommandSource> parse = this.parse(normalizedInput, source);
-      return dispatcher.execute(parse) != BrigadierCommand.FORWARD;
+      if (dispatcher.execute(parse) != BrigadierCommand.FORWARD) {
+        eventManager.fireAndForget(new CommandExecutedEvent(source, cmdLine));
```
hugmanrique: Instead of simply saving the raw (string) command, could we save the Brigadier `ParseResults` in the event? The raw command can still be retrieved through `parse.getReader().getString()` (feel free to provide a getter in the event class itself; `getCommandLine()` maybe?).

hugmanrique (commented):
Requested one minor change. I'm still dubious about the need for this event, though. I will let others weigh in.

api/src/main/java/com/velocitypowered/api/event/command/CommandExecutedEvent.java
```diff
@@ -0,0 +1,60 @@
...
+import com.google.common.base.Preconditions;
+import com.velocitypowered.api.command.CommandSource;
+import com.velocitypowered.api.event.annotation.AwaitingEvent;
+
+/**
+ * This event is fired when velocity executed a command. This event is called after the event
+ * is handled. It's not fired for forwarded commands or commands with syntax errors.
+ */
+@AwaitingEvent
+public final class CommandExecutedEvent {
```
4drian3d: I think the event needs a better name, anyone can get confused between `CommandExecuteEvent` and `CommandExecutedEvent`, especially if you autocomplete the event with an IDE

Joo200: Maybe `ServerCommandExecutedEvent` or `PostCommandExecuteEvent`?

4drian3d: > Maybe `ServerCommandExecutedEvent` or `PostCommandExecuteEvent`?

I like the `PostCommandExecuteEvent` more, but I am not totally convinced because the event is not executed after each command, but, after each command successfully executed from Velocity

Joo200: We can change the event to cover all result types (Forward, Syntax error and execution), then PostCommandExecuteEvent would be the correct name

Joo200: I renamed it to `PostCommandExecuteEvent` and refactored the code a little bit.

This way the result can catch an exception and syntax errors too.

proxy/src/main/java/com/velocitypowered/proxy/command/VelocityCommandManager.java
Joo200: With my most recent changes this is not possible anymore.
Not sure if the ParseResults should be exposed through the API, that implementation doesn't look too stable.

Refactoring the whole command process is a bigger deal and I don't want to change too much. I need the executed commands for my plugin but I'm happy if I can cover other use cases too.

api/src/main/java/com/velocitypowered/api/event/command/PostCommandExecuteEvent.java
```diff
@@ -0,0 +1,81 @@
...
+            + "commandSource=" + commandSource
+            + ", command=" + command
+            + '}';
+  }
+
+  /**
+   * The result of the executed command.
+   */
+  public enum ExecuteResult {
+    EXECUTED,
```
hugmanrique: I like the generality of this new approach (indeed, the next major version will likely replace the `boolean` return of the `execute*` family of methods by an enum similar to this one). For this reason, I would move this enum to its own file.

I agree with 4drian3d; even with this change I still don't like the name. As an API user I would expect this event to be fired only in the case that a command was actually executed. If the command input has incorrect syntax, then a command clearly wasn't executed. Similarly, if an exception occurs during command execution, the command wasn't fully executed and the `Post` part of the event/code flow wasn't reached. (The javadoc needs an update btw.)

I understand the need for an event to listen to everything command-related. However, this use case is very niche (I can only think about logging plugins that upload exceptions to e.g. services like Sentry---even that should be handled by a SLFJ integration). One approach would be to have one event type (`CommandSyntaxErrorEvent`, `CommandForwardedEvent`, etc.) per case, but this could be too "heavy"/"enterprisey". Another way to partially satisfy your use case follows. A mostly unknown feature provided by Brigadier is setting an execution callback (https://github.com/Mojang/brigadier/blob/a6c4baa1b639da62f7f0df62aa61de713e2136c8/src/main/java/com/mojang/brigadier/CommandDispatcher.java#L107 ). Brigadier calls this function with the results of a command execution, or on error. I say "partially" because your plugin wouldn't be notified of syntax errors, only of execution success, forward (by checking the `int` return code--this is API), or error.

In short, we need to come up with a better name for this event now that it is fired in response to more cases than execution success on the proxy. The `ExecuteResult` also needs a new name: I propose simply `CommandResult`, inspired by other Minecraft APIs. If we cannot come up with something better, then I suggest sticking to the Brigadier-callback approach above. We could then re-evaluate the situation on a future major version of Velocity, and modify the contract of command execution methods (for example, maybe it was a mistake to fire the original `CommandExecuteEvent` if the given input is for a command known by the proxy but the input has invalid syntax?).

Regarding not exposing the `ParseResults`: this is a public type provided by Brigadier, and in case the parse fails it contains data on why that is the case. Under this new approach, we could provide a nullable getter.

hugmanrique (changes requested):

proxy/src/main/java/com/velocitypowered/proxy/command/VelocityCommandManager.java
hugmanrique: The command process cannot be refactored (as hinted by the multiple todo comments, heh) due to API requirements, unfortunately.

See the comment below for why it's safe and useful to expose `ParseResult` objects.

api/src/main/java/com/velocitypowered/api/event/command/PostCommandExecuteEvent.java
Joo200: > A mostly unknown feature provided by Brigadier is setting an execution callback 

This looks nice and for me the right thing, however this is not exposed in the Velocity API. It's possible to set **one** Consumer but not multiple so exposing it only makes sense as an event.

After your hint this should be the better idea. Instead of implementing the event like my suggestion simply use the consumer callback, call the event there and it's done.

Then I don't need to add more logic to VelocityCommandManager and it's done in a few lines of code.

hugmanrique: The current implementation is mostly fine, the only issue is the naming of the event.

Velocity is published with Brigadier, so plugins can (and actually do) use its API freely (similarly to how you use Brigadier classes to construct your commands, you can use other Brigadier methods and utilities). Velocity does not and will most likely not need to use the callback feature in the future (we maintain our own API-compatible branch of Brigadier). Plugins are advised to set their own callback, but respecting others that might have set it before: call the already set callback, if any, inside your callback.

I understand this callback-driven API is a bit restrictive, but the implementation in this PR needs no more changes. Instead, please propose alternative names for the event.

Joo200: Just a note about your Brigadier suggestion:

Velocity does not expose the `CommandDispatcher` instance through the api. That's why a plugin can't use that method. And beside of that: `CommandDispatcher` is API but doesn't expose already active Consumers.

Plugins can't use that callback without reflection and can't call already set callbacks.

Using an event instead should be much more reliable. What's about `ProxyCommandExecuteEvent`, the current implementation of this event is only called when the proxy tries to execute the Command and it was not forced to the backend server in `CommandExecuteEvent`.

Feel free to suggest any class names.

api/src/main/java/com/velocitypowered/api/command/CommandResult.java
```diff
@@ -0,0 +1,30 @@
...
+ * Copyright (C) 2020-2021 Velocity Contributors
+ *
+ * The Velocity API is licensed under the terms of the MIT License. For more details,
+ * reference the LICENSE file in the api top-level directory.
+ */
+
+package com.velocitypowered.api.command;
+
+/**
+ * The result of an executed command.
```
hugmanrique: Replace "an executed command" by "a command invocation attempt"

api/src/main/java/com/velocitypowered/api/command/CommandResult.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+public enum CommandResult {
+  /**
+   * The command was successfully executed by velocity.
```
hugmanrique: s/velocity/the proxy (or capitalize Velocity).

api/src/main/java/com/velocitypowered/api/command/CommandResult.java
```diff
@@ -0,0 +1,30 @@
...
+   */
+  EXECUTED,
+  /**
+   * The command was forwarded to the backend.
```
hugmanrique: "The command invocation was forwarded to the backend server. The command may be successfully executed or not."

api/src/main/java/com/velocitypowered/api/command/CommandResult.java
```diff
@@ -0,0 +1,30 @@
...
+   */
+  FORWARDED,
+  /**
+   * A syntax error occurred while executing the command.
```
hugmanrique: "The provided command input contained syntax errors."

api/src/main/java/com/velocitypowered/api/command/CommandResult.java
```diff
@@ -0,0 +1,30 @@
...
+   */
+  SYNTAX_ERROR,
+  /**
+   * A unexpected exception was thrown while executing the command.
```
hugmanrique: An unexpected exception occurred while executing the command in the proxy.

api/src/main/java/com/velocitypowered/api/event/command/ProxyCommandExecuteEvent.java
```diff
@@ -8,17 +8,21 @@
...
 
 import com.google.common.base.Preconditions;
+import com.velocitypowered.api.command.CommandResult;
 import com.velocitypowered.api.command.CommandSource;
 
 /**
  * This event is fired when velocity executed a command. This event is called after the event
  * is handled. It's not fired for forwarded commands or commands with syntax errors.
+ *
+ * <p>Commands can be cancelled or forced to backend servers in {@link CommandExecuteEvent}
```
hugmanrique: If by _forced_ we mean _forwarded_, then the new event is expected to be fired, right? This would correspond to the `CommandResult.FORWARDED` value.

I agree that we shouldn't fire the new event if the execution is outright cancelled in `CommandExecuteEvent`.

proxy/src/main/java/com/velocitypowered/proxy/command/VelocityCommandManager.java
```diff
@@ -242,7 +241,7 @@ public CompletableFuture<Boolean> executeAsync(final CommandSource source, final
     Preconditions.checkNotNull(cmdLine, "cmdLine");
 
     return callCommandEvent(source, cmdLine).thenApplyAsync(event -> {
-      CommandResult commandResult = event.getResult();
+      CommandExecuteEvent.CommandResult commandResult = event.getResult();
```
hugmanrique, line 257: This is an unfortunate name clash, but nothing we can do yet. We could merge both enums in the next major version.

api/src/main/java/com/velocitypowered/api/event/command/ProxyCommandExecuteEvent.java
```diff
@@ -8,17 +8,21 @@
...
+ * this event will not be caled in those cases.</p>
  */
-public final class PostCommandExecuteEvent {
+public final class ProxyCommandExecuteEvent {
```
hugmanrique: One name idea: `PostCommandInvocationEvent`. It's long, but it follows the same naming as the already present [`CommandInvocation`](https://github.com/PaperMC/Velocity/blob/36cc2bde23e7d50545c04b4060a52f3e6b461368/api/src/main/java/com/velocitypowered/api/command/CommandInvocation.java#L15) class, which "provides information related to the **possible execution** of a command".

hugmanrique (changes requested):
I'm happy with the current API and implementation. I provided some minor documentation changes, and a name suggestion.

Thanks for the contribution :smile:

kennytv (approved):

api/src/main/java/com/velocitypowered/api/event/command/PostCommandInvocationEvent.java
```diff
@@ -0,0 +1,75 @@
...
+
+package com.velocitypowered.api.event.command;
+
+import com.google.common.base.Preconditions;
+import com.velocitypowered.api.command.CommandResult;
+import com.velocitypowered.api.command.CommandSource;
+
+/**
+ * This event is fired when velocity executed a command. This event is called after the event
+ * is handled. It's not fired for forwarded commands or commands with syntax errors.
```
hugmanrique: It _is_ fired for forwarded commands or input with syntax errors.

api/src/main/java/com/velocitypowered/api/event/command/PostCommandInvocationEvent.java
```diff
@@ -0,0 +1,75 @@
...
+ * This event will not be called in those cases.</p>
+ */
+public final class PostCommandInvocationEvent {
+
+  private final CommandSource commandSource;
+  private final String command;
+  private final CommandResult result;
+
+  /**
+   * Constructs a CommandExecuteEvent.
```
hugmanrique: Update the class name.

api/src/main/java/com/velocitypowered/api/event/command/PostCommandInvocationEvent.java
```diff
@@ -0,0 +1,75 @@
...
+   *
+   * @return the execution result
+   */
+  public CommandResult getResult() {
+    return result;
+  }
+
+  @Override
+  public String toString() {
+    return "CommandExecuteEvent{"
```
hugmanrique: Update the class name.

hugmanrique (approved):

Joo200:
**Push**

rebased and resolved merge conflicts.

4drian3d (approved):

PR #987, opened 2023-03-29
feat: add TabList#addEntries

### LINKED ISSUE #958: Tablist addEntries(List<TabListEntry> entries)
This is possible to add addEntries(List<TabListEntry> entries) in function of tablist?

powercasgamer:
Closes #958

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/proxy/player/TabList.java
+++ b/api/src/main/java/com/velocitypowered/api/proxy/player/TabList.java
@@ -44,6 +44,28 @@ public interface TabList {
   void addEntry(TabListEntry entry);
 
   /**
+   * Adds a {@link Iterable} of {@link TabListEntry}'s to the {@link Player}'s tab list.
+   *
+   * @param entries to add to the tab list
+   */
+  default void addEntries(Iterable<TabListEntry> entries) {
+    for (TabListEntry entry : entries) {
+      addEntry(entry);
+    }
+  }
+
+  /**
+   * Adds an array of {@link TabListEntry}'s to the {@link Player}'s tab list.
+   *
+   * @param entries to add to the tab list
+   */
+  default void addEntries(TabListEntry... entries) {
+    for (TabListEntry entry : entries) {
+      addEntry(entry);
+    }
+  }
+
+  /**
    * Removes the {@link TabListEntry} from the tab list with the {@link GameProfile} identified with
    * the specified {@link UUID}.
    *


api/src/main/java/com/velocitypowered/api/proxy/player/TabList.java
```diff
@@ -43,6 +44,13 @@ public interface TabList {
    */
   void addEntry(TabListEntry entry);
 
+  /**
+   * Adds a {@link List} of {@link TabListEntry}'s to the {@link Player}'s tab list.
+   *
+   * @param entries to add to the tab list
+   */
+  void addEntries(List<TabListEntry> entries);
```
4drian3d: It is not totally necessary to use a `List` here, you could simply use a `Collection` to allow any collection to be used. It would also be nice to add another method that accepts an array `addEntries(TabListEntry... entries)`

api/src/main/java/com/velocitypowered/api/proxy/player/TabList.java
```diff
@@ -45,11 +44,18 @@ public interface TabList {
...
-   * Adds a {@link List} of {@link TabListEntry}'s to the {@link Player}'s tab list.
+   * Adds a {@link Collection} of {@link TabListEntry}'s to the {@link Player}'s tab list.
    *
    * @param entries to add to the tab list
    */
-  void addEntries(List<TabListEntry> entries);
+  void addEntries(Collection<TabListEntry> entries);
+
+  /**
+   * Adds an array of {@link TabListEntry}'s to the {@link Player}'s tab list.
```
powercasgamer, line 58: Don't really like the word "array" here but couldn't come up with something better

proxy/src/main/java/com/velocitypowered/proxy/tablist/KeyedVelocityTabList.java
```diff
@@ -90,6 +90,20 @@ public void addEntry(TabListEntry entry) {
...
+      addEntry(entry);
+    }
+  }
+
+  @Override
+  public void addEntries(TabListEntry... entries) {
+    for (TabListEntry entry : entries) {
+      addEntry(entry);
+    }
+  }
```
4drian3d: Instead of making an `@Override` in each implementation, it would be better to implement these methods as a default method in the TabList interface

4drian3d (approved):

api/src/main/java/com/velocitypowered/api/proxy/player/TabList.java
alexstaeding: In my experience, most methods such as this one even use Iterable instead of Collection or List - wouldn't that make more sense here?

4drian3d: > In my experience, most methods such as this one even use Iterable instead of Collection or List - wouldn't that make more sense here?

Personally, I feel that `Collection` is the most appropriate and consistent for Velocity, since other methods also use Collection as a parameter or as the type of the object to return

[`Collection<TabListEntry>`](https://jd.papermc.io/velocity/3.0.0/com/velocitypowered/api/proxy/player/TabList.html#getEntries())
[`Collection<Player>`](https://jd.papermc.io/velocity/3.0.0/com/velocitypowered/api/proxy/ProxyServer.html#getAllPlayers())
[`Collection<RegisteredServer>`](https://jd.papermc.io/velocity/3.0.0/com/velocitypowered/api/proxy/ProxyServer.html#getAllServers())

api/src/main/java/com/velocitypowered/api/proxy/player/TabList.java
```diff
@@ -43,6 +43,28 @@ public interface TabList {
    */
   void addEntry(TabListEntry entry);
 
+  /**
+   * Adds a {@link Collection} of {@link TabListEntry}'s to the {@link Player}'s tab list.
+   *
+   * @param entries to add to the tab list
+   */
+  default void addEntries(Collection<TabListEntry> entries) {
```
kashike:
```suggestion
  default void addEntries(Iterable<TabListEntry> entries) {
```

kashike (changes requested):

kashike (approved):

PR #988, opened 2023-03-29
[ci skip] add velocity.toml & forwarding secret file to gitignore

powercasgamer:
before:
image


astei (approved):

astei:
Looks like this change was already applied at some point, so I'll close this out as redundant

PR #989, opened 2023-03-30
feat: implement click callback

powercasgamer:
also updates adventure to 4.13.1

javaw_rFhgS2cem2
test.gif

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/command/builtin/VelocityCommand.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/command/builtin/VelocityCommand.java
@@ -31,6 +31,7 @@ import com.velocitypowered.api.proxy.ProxyServer;
 import com.velocitypowered.api.proxy.server.RegisteredServer;
 import com.velocitypowered.api.util.ProxyVersion;
 import com.velocitypowered.proxy.VelocityServer;
+import com.velocitypowered.proxy.provider.ClickCallbackProviderImpl;
 import com.velocitypowered.proxy.util.InformationUtils;
 import java.io.BufferedWriter;
 import java.io.IOException;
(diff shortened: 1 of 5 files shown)


proxy/velocity.toml
kashike, line 1: Remove this file from your PR.

kashike (changes requested):

proxy/src/main/java/com/velocitypowered/proxy/command/builtin/VelocityCommand.java
```diff
@@ -485,4 +488,33 @@ public boolean hasPermission(CommandSource source, String @NonNull [] args) {
     }
 
   }
+
+  /**
+   * Callback SubCommand.
+   */
+  public static class Callback implements SubCommand {
+    private static final Logger logger = LogManager.getLogger(Callback.class);
```
4drian3d: This Logger creation is unnecessary, it is not used anywhere.

4drian3d (approved):

proxy/src/main/java/com/velocitypowered/proxy/provider/ClickCallbackProviderImpl.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    public boolean valid() {
+      return hasRemainingUses() && !expired();
+    }
+
+    public UUID id() {
+      return this.id;
+    }
+  }
```
astei, line 138: I'm really not a big fan of this approach.

* We already have an expiring cache, Caffeine. Why have an eviction task running every 5 milliseconds?
* The `HashMap` used here is not thread safe. Velocity is a multi-threaded environment. This kind of implementation would make sense with Paper, with a single server thread, but it won't fly here.

For this reason I am not willing to sign off on this implementation unless these issues were resolved.

astei (changes requested):

powercasgamer:
closing in favor of https://github.com/PaperMC/Velocity/tree/feat/click-callback-provider ?

PR #991, opened 2023-04-02
Added Unix Domain Socket support.

0xIO32:
It adds Unix Domain Socket support, so that it don't require a tcp port to be binded to work.

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/proxy/server/ServerInfo.java
+++ b/api/src/main/java/com/velocitypowered/api/proxy/server/ServerInfo.java
@@ -8,7 +8,7 @@
 package com.velocitypowered.api.proxy.server;
 
 import com.google.common.base.Preconditions;
-import java.net.InetSocketAddress;
+import java.net.SocketAddress;
 import java.util.Objects;
 import org.checkerframework.checker.nullness.qual.Nullable;
 
@@ -19,15 +19,15 @@ import org.checkerframework.checker.nullness.qual.Nullable;
 public final class ServerInfo implements Comparable<ServerInfo> {
 
   private final String name;
-  private final InetSocketAddress address;
-
+  private final SocketAddress address;
+  
   /**
    * Creates a new ServerInfo object.
    *
    * @param name the name for the server
    * @param address the address of the server to connect to
    */
-  public ServerInfo(String name, InetSocketAddress address) {
+  public ServerInfo(String name, SocketAddress address) {
     this.name = Preconditions.checkNotNull(name, "name");
     this.address = Preconditions.checkNotNull(address, "address");
   }
@@ -35,8 +35,8 @@ public final class ServerInfo implements Comparable<ServerInfo> {
   public final String getName() {
     return name;
   }
-
-  public final InetSocketAddress getAddress() {
+  
+  public final SocketAddress getAddress() {
     return address;
   }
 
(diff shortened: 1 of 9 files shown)


MarkusTieger:
As Information, i created for Paper also an Pull Request which adds Unix Domain Socket support:
https://github.com/PaperMC/Paper/pull/9084

astei:
This is a pretty clear API breakage (not to mention the fact it requires us to expose internal implementation details to plugins), and while it's been planned for the next major revision of Velocity, it's not been a high priority for a few reasons (mainly due to lack of time on my part).

0xIO32:
I think it shouldn't break the api. In the only class in the api that i modified, i only added methods or modified private fields and added constructors, but removed or modified no existing components (except for the "ServerInfo" constructor to use the SocketAddress type as parameter and not the InetSocketAddress type, which shouldn't break something, because the type InetSocketAddress extends from it)

astei:
Modifying the existing 2-ary constructor of `ServerInfo` constitutes API breakage. The class specifically asks for `InetSocketAddress`. While this is _source code compatible_ (I can compile against this change), it's not _binary compatible_.

In addition, `getAddress()` is now broken - it's now possible that a `ClassCastException` can be thrown by the JVM, since you cannot cast `DomainSocketAddress` to an `InetSocketAddress`.

0xIO32:
Yes i could add a third constructor with this type and for the `getAddress()` method, i could add a change, that the `InetAddress.getLoopbackAddress()` ( "127.0.0.1" ), will be used when it is a `DomainSocketAddress`. That would than break nothing?

api/src/main/java/com/velocitypowered/api/proxy/server/ServerInfo.java
```diff
@@ -19,24 +21,55 @@
...
+   * Returns the Address of the server to connect to
+   * or if it isn't a InetSocketAddress it returns a loopback address.
+   * }
+   */
+  @Deprecated
   public final InetSocketAddress getAddress() {
+    if (address instanceof InetSocketAddress) {
+      return (InetSocketAddress) address;
+    }
+    return InetSocketAddress.createUnresolved("127.0.0.1", 25565);
```
hugmanrique: This is not an acceptable default, especially because it can create unexpected duplicates in e.g. maps used by plugins.

api/src/main/java/com/velocitypowered/api/proxy/server/AddressType.java
```diff
@@ -0,0 +1,17 @@
...
+ * The Velocity API is licensed under the terms of the MIT License. For more details,
+ * reference the LICENSE file in the api top-level directory.
+ */
+
+package com.velocitypowered.api.proxy.server;
+
+/**
+ * Enumerates the supported Address types.
+ */
+public enum AddressType {
```
hugmanrique: I don't see much point in having this enum, since the addresses can already be distinguished by their class. I prefer the API from the original domain sockets commit in https://github.com/PaperMC/Velocity/commit/b00389029f4c11ad34898b5298625f4fcfc03de5

Reverting this addition would also simplify the other `InetSocketAddress -> Pair<SocketAddress, AddressType>` changes.

hugmanrique (changes requested):
I don't see how to avoid the `getAddress` breakage sensibly. We will have to wait a bit, iirc I've seen some talks about a near-future major release launch in Discord.

I would prefer to see the commit mentioned in the other comment backported, since astei already tested those changes (the address parsing portion of this PR is a perfect candidate for some unit tests).

0xIO32:
Now i removed the `getAddress` method, the `AddressType` enum and the constructor which was only used for not breaking the api.

PR #995, opened 2023-04-04
feat: player displaynames

### LINKED ISSUE #993: Add Player.getDisplayName()
Hi there! I'm looking to enhance the Player interface by adding two new features. 

The first one is a method called Player#.getDisplayName() which will return the username of the player by default. 
The second one is Player#.setDisplayName(Component text), which will allow me to set a custom prefix based on the player's rank more easily. 

Thank you!

powercasgamer:
Closes #993 

Something to consider, should `/glist` use the player's name or displayname?

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/proxy/Player.java
+++ b/api/src/main/java/com/velocitypowered/api/proxy/Player.java
@@ -51,6 +51,20 @@ public interface Player extends
   String getUsername();
 
   /**
+   * Returns the player's display name.
+   *
+   * @return the display name
+   */
+  Component getDisplayName();
+
+  /**
+   * Sets the player's display name.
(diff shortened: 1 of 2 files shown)


Xernium (changes requested):
This feature has my veto as a team member. We‘ve had a lengthy discussion back in the day around Velocity 1.1.x about the viability and use of „display names“. The consensus was that it‘s a confusing feature that has no real use on Velocity.
That said, we have considered and are still evaluation the ability to add dynamic link objects to something like the Player-linked data structures. Using that you could add a „display name“ to a player object; It would roughly result in the same thing.
This is, and was, mainly a feature to eliminate plugins using Map<Player, linked object…> maps

PR #996, opened 2023-04-07
Add kqueue support

JNNGL:


astei (approved):
Approving, but this seems to be of dubious value. bStats reports that 0.1% of users actually use macOS to run Velocity:
image

PR #1003, opened 2023-04-21
Switch to using a ConcurrentMap for storing TabListEntries

### LINKED ISSUE #1002: VelocityTabList uses a non-thread safe map for storing player entries
Adding, removing and iterating on entries to the tab list concurrently across multiple threads can lead to various race conditions as the internal map used to store entries isn't thread safe

https://github.com/PaperMC/Velocity/blob/e364e2c7d1918ec7c20986fb640f3f6a64127bb0/proxy/src/main/java/com/velocitypowered/proxy/tablist/VelocityTabList.java#L55-L65

In my plugin, for instance, I have a Task that updates and replaces player entries on a VelocityTabList. This task runs concurrently. When players join/leave while this task is running, however, the lack of concurrency on this map causes issues when updating offline players. I think this map ought to be made a ConcurrentHashMap to better support concurrent operations on it.

WiIIiam278:
Fixes #1002 

Uses [`Maps#newConcurrentMap()`](https://guava.dev/releases/19.0/api/docs/src-html/com/google/common/collect/Maps.html#line.283) which creates an implementing ConcurrentMap through MapMaker, which fixes issues adding/removing entries from VelocityTabLists concurrently.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/tablist/VelocityTabList.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/tablist/VelocityTabList.java
@@ -62,7 +62,7 @@ public class VelocityTabList implements InternalTabList {
   public VelocityTabList(ConnectedPlayer player) {
     this.player = player;
     this.connection = player.getConnection();
-    this.entries = Maps.newHashMap();
+    this.entries = Maps.newConcurrentMap();
   }
 
   @Override


electronicboy:
Just changing that to a CHM won't alleviate any actual concurrency concerns here

WiIIiam278:
My current understanding through reading VelocityTabList is that thread safety isn't guaranteed when iterating through/adding/removing entries; the specific issue I've encountered is inconsistencies with what I'm able to iterate over when simultaneously updating the collection of entries while reading its contents. But perhaps you're talking about other concurrency issues with regard to how it sends the update tab list packets. 

Also worth mentioning that LegacyVelocityTabList is backed by a CHM here, though not sure that's necessarily relevant

PR #1009, opened 2023-05-11
Expose connecting player's UUID in the PreLoginEvent

kezz:
Unsure if docs need to be more specific, or if the old event constructor should be deprecated/removed.


0utplay:
I think using `uniqueId` as field, parameter and getter name is more appropriate as the API usally uses that for any uuids.

https://github.com/PaperMC/Velocity/blob/7f776abf550001186a1f11f4e5ff760dd66e3b04/api/src/main/java/com/velocitypowered/api/proxy/Player.java#L75

proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/ServerLogin.java
```diff
@@ -26,6 +26,7 @@
 import com.velocitypowered.proxy.protocol.ProtocolUtils.Direction;
 import com.velocitypowered.proxy.util.except.QuietDecoderException;
 import io.netty.buffer.ByteBuf;
+import java.util.Objects;
```
astei: Unused import

api/src/main/java/com/velocitypowered/api/event/connection/PreLoginEvent.java
```diff
@@ -33,17 +34,30 @@ public final class PreLoginEvent implements ResultedEvent<PreLoginEvent.PreLogin
...
   /**
-   * Creates a new instance.
+   * Creates a new instance, without an associated UUID.
    *
    * @param connection the connection logging into the proxy
    * @param username the player's username
    */
   public PreLoginEvent(InboundConnection connection, String username) {
+    this(connection, username, null);
+  }
```
astei: let's deprecate this

astei (changes requested):

api/src/main/java/com/velocitypowered/api/event/connection/PreLoginEvent.java
```diff
@@ -55,6 +69,15 @@ public String getUsername() {
     return username;
   }
 
+  /**
+   * Returns the UUID of the connecting player. This value is {@code null} on 1.19.1 and lower.
+   *
+   * @return the uuid
+   */
+  public @Nullable UUID getUuid() {
```
astei: Change this to `getUniqueId()`

astei (changes requested):

kyngs:
Isn't the GameProfileRequestEvent called after the PreLoginEvent? If yes, what's the point of exposing the UUID even though it still cannot be used for DB purposes?

proxy/src/main/java/com/velocitypowered/proxy/connection/client/InitialLoginSessionHandler.java
```diff
@@ -119,7 +119,7 @@ public boolean handle(ServerLogin packet) {
     inbound.setPlayerKey(playerKey);
     this.login = packet;
 
-    PreLoginEvent event = new PreLoginEvent(inbound, login.getUsername());
+    PreLoginEvent event = new PreLoginEvent(inbound, login.getUsername(), login.getHolderUuid());
```
456dev: is this the uuid sent by the client?

4drian3d: Yes

4drian3d (approved):

PR #1025, opened 2023-06-18
Hide ip address in initial and legacy connections

PancakeTAS:
Turns out cancelling an initial connection still logged the ip address in the console and so does a legacy connection. When specified in velocity.toml `enable-player-address-logging = false` all ip addresses should be hidden.
image

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/client/HandshakeSessionHandler.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/client/HandshakeSessionHandler.java
@@ -235,7 +235,12 @@ public class HandshakeSessionHandler implements MinecraftSessionHandler {
 
     @Override
     public String toString() {
-      return "[legacy connection] " + this.getRemoteAddress().toString();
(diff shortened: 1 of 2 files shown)


electronicboy (approved):

hugmanrique:
Can we extract the address formatting logic into a utility method and also use in the modern variants?

PancakeTAS:
Are you sure we need to extract those 3 occurences into a util? Where would I put it?
Also do you mean by "modern variants"? Keep in mind I don't really know much of the codebase  @hugmanrique

hugmanrique:
You're right, I thought the non-legacy connection code also included ≈3 more occurrences of this code. Turns out it's only one, https://github.com/search?q=repo%3APaperMC%2FVelocity%20%22ip%20address%20withheld%22&type=code

electronicboy (approved):

theFoxley:
I don't really understand the usefulness of this change. It's just a displayed IP that is not linked to any other player information...
And it's the only information printed, so the log doesn't mean anything now

electronicboy:
Some people don't want to record this form of information due to stuff like the GDPR

PR #1027, opened 2023-06-19
Add server-id hash to LoginEvent

roccodev:
This adds the `serverId` hash (that was sent to Mojang during authentication) to a login event for plugins to use.

The reason for this is that while `prevent-client-proxy-connections` is currently working (previously, a Mojang bug prevented it from working), there is no way to distinguish a proxy block from a regular offline-mode user.

Based on production experience, the proxy block system might prevent legitimate players from joining, especially those in
countries that employ network filtering.

This solution acts as a workaround to allow plugins to manually check for proxies (by sending more `hasJoined` requests to Mojang), by giving them access to the server ID hash that was generated during authentication.

There are definitely more ways to expose this information, such as:

* Storing the hash in `ConnectedPlayer`, however this would bloat the `Player` API with something that's only relevant during login. Additionally, Mojang starts rejecting the same hash after a while.
* Creating a different event, like `AuthenticationEvent`, dedicated to auth data like the server ID hash.

I think `LoginEvent` is currently the best compromise to store this data without introducing a new event, let me know what you think.


api/src/main/java/com/velocitypowered/api/event/connection/LoginEvent.java
```diff
@@ -22,17 +22,34 @@
...
 
   public Player getPlayer() {
     return player;
   }
 
+  /**
+   * Returns the server ID hash that was sent to Mojang to authenticate the player.
+   * If the connection was in offline-mode, this returns {@code null}.
+   */
+  public String getServerIdHash() {
```
kennytv: missing a nullable annotation

roccodev:
Hey, thanks for reviewing. I just wanted to check, are you satisfied with the design?

It's quite an old PR so I wanted to make sure it's still the right fit before I resolve the conflicts and make the minor change.

For what it's worth, I've since been running this patch in production, and it's worked quite well.

kennytv:
Yeah should be alright

4drian3d (approved):

api/src/main/java/com/velocitypowered/api/event/connection/LoginEvent.java
```diff
@@ -22,17 +23,34 @@
 public final class LoginEvent implements ResultedEvent<ResultedEvent.ComponentResult> {
 
   private final Player player;
+  private final String serverIdHash;
   private ComponentResult result;
 
-  public LoginEvent(Player player) {
```
Emilxyz, line 32: Do we consider event constructors to be part of the public API? If so, we might want to deprecate the old constructor instead of removing

kennytv (approved):

PR #1030, opened 2023-06-25
clarify what file has to be created

powercasgamer:


electronicboy:
I don't think that this really clarifies too much, it says that the file
doesn't exist, but, most people who hit this don't understand the nature of
this file

Honestly, my plan is to rewrite the configuration stuff and add a migration
that moves the option over if it already exists, or make sure that it can
actually Regen the file automatically; maybe only if the setting is
unset/the default, as this situation shouldn't occur, and generally only
occurs when running in an environment that manually creates a partial
config, shamefully low down on the Todo list as I don't really have
hardware able to work this stuff ATM

PR #1036, opened 2023-07-01
Log exceptions instead of just kicking the player (ClientPlaySessionHandler)

ytendx:
This commit is build-able using the script. 

While debugging a connection problem i noticed that these errors do not get printed.
This would really help my workflow.


ytendx:
Please don't mind me, errors still get logged elsewhere

Issue #909, opened 2022-12-17
Player get kicked due to out-of-order chat when sending messages rapidly


Glucy-2:
Velocity version: `3.1.2-SNAPSHOT-201`
* No Velocity plugin installed
* No matter whether the value of `force-key-authentication` is true or false, the player will be kicked
* I've tested `modern` and `none` player info forwarding mode

Server: 1.19.3, fabric with mods: `fabric-api`, `fabric-carpet`, `carpet-tis-addition`, `fabricproxy-lite`
Velocity log:
````py
[16:02:00 INFO]: [connected player] AnnaNo (/113.x.x.x:56524) has connected
[16:02:00 INFO]: [server connection] AnnaNo -> mirror has connected
[16:02:38 INFO]: [connected player] AnnaNo (/113.x.x.x:56524) -> executed command /gamerule sendCommandFeedback
[16:02:38 INFO]: [connected player] AnnaNo (/113.x.x.x:56524) -> executed command /fill 9 -32 256 10 -32 256 minecraft:iron_block replace air
[16:02:38 INFO]: [connected player] AnnaNo (/113.x.x.x:56524) -> executed command /setblock 9 -32 257 minecraft:observer[facing=south,powered=false]
[16:02:38 INFO]: [connected player] AnnaNo (/113.x.x.x:56524) -> executed command /fill 9 -29 256 9 -29 257 minecraft:redstone_wire[east=none,north=side,power=0,south=side,west=none] replace air
...
[16:02:38 INFO]: [connected player] AnnaNo (/113.x.x.x:56524) -> executed command /setblock 11 -29 256 minecraft:white_stained_glass
[16:02:38 INFO]: [connected player] AnnaNo (/113.x.x.x:56524): kicked from server mirror: multiplayer.disconnect.out_of_order_chat
[16:02:38 INFO]: [server connection] AnnaNo -> mirror has disconnected
[16:02:38 INFO]: [connected player] AnnaNo (/113.x.x.x:56524) has disconnected: You were kicked from mirror: multiplayer.disconnect.out_of_order_chat
````
fabric server log:
````py
[16:02:00] [Server thread/INFO]: AnnaNo[/113.x.x.x:49340] logged in with entity id 142 at (-0.4497645856701771, -40.0, 248.48987400782113)
[16:02:00] [Server thread/INFO]: AnnaNo joined the game
[16:02:01] [Server thread/INFO]: Player AnnaNo joined with a matching carpet client
[...]

PR #399, opened 2020-12-11
Spelling issue

jkcoxson:
aserver -> a server

PR #1042, opened 2023-07-11
Fix chat race condition

EpicPlayerA10:
# The issue
Chat system in velocity has a race condition. When player is sending messages very fast, then a backend server will kick him for `out-of-order chat`.

You can test this by installing [litematica mod](https://www.curseforge.com/minecraft/mc-mods/litematica) and use the feature called `Paste schematic into world`. Then litematica will send commands very fast and then you get kicked from the proxy. This issue only happens when player is connected to the proxy. When you are connected to the minecraft server directly then the issue disappears.

The race condition looks like this:
<details>
<summary>Flowchart before</summary>

Flowchart before
</details>

You can see here that velocity doesn't wait for packet to be sent. The chat gets out of sync and thats why players are getting kicked from backend server.

# Fix
This pull request fixes it by waiting for packets to be sent. Now the flowchart looks like this:
<details>
<summary>Flowchart after</summary>

Flowchart after
</details>

## Related issues
- #909


electronicboy:
Don't have hardware to look into this properly, but on the surface, this looks much closer to how I said this stuff should probably work in the first place

intersecato:
Hi @EpicPlayerA10 @electronicboy,
we were able to test the fix in production with various users. The problem seems to persists but more rarely.
We will try to investigate the problem further.

image

EpicPlayerA10:
> The problem seems to persists but more rarely.

Okay, now the latest commit will ensure that chat packets are fully sent. This should get rid of this problem.

intersecato:
> 
> 
> Okay, now the latest commit will ensure that chat packets are fully sent. This should get rid of this problem.

The problem is still there. A player reported me about a kick again today.

EpicPlayerA10:
> The problem is still there. A player reported me about a kick again today.

Hmm. I don't know why this is still happening. I have no idea. I think that i did everything correct.

CatTeaA:
Is there any progress? Looking forward to it

electronicboy:
I recall looking over this a few times and it generally looked fine, finally have hardware to test, just time; if this can be rebased, I'll try and get to this on the weekend or sometime next week

EpicPlayerA10:
Rebased

electronicboy (approved):

Incluuu:
I receive a lot of reports from player, I use version #399 velocity.

sent out-of-order chat: 'spawn': 1718714687 > 1718714673
lost connection: Out-of-order chat packet received. Did your system time change?

sent out-of-order chat: 'tpaccept': 1718715264 > 1718715261
lost connection: Out-of-order chat packet received. Did your system time change?

sent out-of-order chat: 'spawn': 1718715542 > 1718715530
lost connection: Out-of-order chat packet received. Did your system time change?

sent out-of-order chat: 'shop': 1718715930 > 1718715916
lost connection: Out-of-order chat packet received. Did your system time change?

demxn013:
this issue still exists on 1.20 paper backend server on client version 1.21. Im using the most recent version of velocity. 
It only happens with commands though.
image

7wOv6ySCjo:
The bug is still occurring in version 26.1.2
We get the error: ‘keepalive: time out (out of order!)’
Several disconnections a day for various players.

I'm not entirely sure whether it's an ‘out of order’ error with keepalive packets or with the chat specifically. The strange thing is that you don't even need to type commands or send messages in the chat; just exploring the world for a few minutes is enough to trigger this ‘timeout’ error.

EpicPlayerA10:
> The bug is still occurring in version 26.1.2 We get the error: ‘keepalive: time out (out of order!)’ Several disconnections a day for various players.
> 
> I'm not entirely sure whether it's an ‘out of order’ error with keepalive packets or with the chat specifically. The strange thing is that you don't even need to type commands or send messages in the chat; just exploring the world for a few minutes is enough to trigger this ‘timeout’ error.

Can you send a screenshot of this error?

WouterGritter:
@7wOv6ySCjo could you check and confirm whether #1782 solves this issue for you or not?

7wOv6ySCjo:
> > The bug is still occurring in version 26.1.2 We get the error: ‘keepalive: time out (out of order!)’ Several disconnections a day for various players.
> > I'm not entirely sure whether it's an ‘out of order’ error with keepalive packets or with the chat specifically. The strange thing is that you don't even need to type commands or send messages in the chat; just exploring the world for a few minutes is enough to trigger this ‘timeout’ error.
> 
> Can you send a screenshot of this error?

Captura de pantalla (8136)

This usually happens with entity ID number 1, as it doesn’t happen as often with subsequent entities ID, but it still disconnects the user.

I investigated the code and the line causing the disconnection is in: net.minecraft.server.network.ServerCommonPacketListenerImpl

                LOGGER.info("Disconnecting {} for sending keepalive response ({}) out-of-order!", this.playerProfile().name(), packet.getId());
                //this.disconnectAsync(TIMEOUT_DISCONNECTION_MESSAGE, io.papermc.paper.connection.DisconnectionReason.TIMEOUT);
                
                and
                
        LOGGER.info("Disconnecting {} for sending keepalive response ({}) without matching challenge!", this.playerProfile().name(), packet.getId());
        //this.disconnectAsync(TIMEOUT_DISCONNECTION_MESSAGE, io.papermc.paper.connection.DisconnectionReason.TIMEOUT);
                
I simply commented out those two disconnection lines and that was it; I no longer lose users due to arbitrary disconnections, although I don’t know what the consequences of preventing this disconnection might be.

PR #1047, opened 2023-07-29
Update snakeyaml to version 2.0 in api/build.gradle.kts

Xemii16:
I was getting error java.lang.NoSuchMethodError: 'void org.yaml.snakeyaml.LoaderOptions.setCodePointLimit(int)' using snakeyaml 2.0 in my plugin. I learned that the Velocity core contains the snakeyaml version 1.33 library. I suggest replacing the version with 2.0 because of the large number of vulnerabilities in the older version

### DIFF
--- a/api/build.gradle.kts
+++ b/api/build.gradle.kts
@@ -30,6 +30,8 @@ dependencies {
     api("net.kyori:adventure-text-serializer-plain")
     api("net.kyori:adventure-text-minimessage")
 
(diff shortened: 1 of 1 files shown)


pascalgutsche:
I have the same issue. It would be great to update to version 2.0 or even 2.2!

astei:
SnakeYAML 2.x is a major API breakage. The best we can do is version 1.33.

PR #1057, opened 2023-08-13
Replaced weired i with i

Plugrol:
In this little patch I replaced an i which caused my build process to crash with an i


electronicboy:
I mean, his should be merged but, you should probably work out why your environment is broken

electronicboy (approved):

PR #1060, opened 2023-08-15
Set project encoding to UTF-8 

Plugrol:
This commit updates the build script to specifically set the project encoding to UTF-8 for JavaCompile, Javadoc, and ProcessResources tasks to ensure consistent handling of characters across different environments.


Nacioszeczek:
Is there a specific issue you're encountering that this is necessary?

Plugrol:
Yes, after cloning this repo on my arch linux (KDE) system, IntelliJ has set the wrong charset whichfor my build process crashed. However I was able to fix it with the addition I made.

alexstaeding:
I don't think setting the release version to Java 11 is necessary considering the toolchain is already configured for Java 11.
Also, configuring tasks should probably be done in the task block at the bottom so that the build script is a bit more sensibly organized.

As a side note, I use the same environment as you (Arch Linux, KDE) and do not have any charset issues. Are you sure you set it correctly with `localectl`? Verify that the output of `locale` shows the correct UTF-8 encoding.

Plugrol:
My encoding seems to be correctly set to UTF-8. However I agree on the part about the java version and maybe also the placement (even I must say that this placement also is pretty structured and clearly does it job). I'll adjust this later on as I am going to be gone for a week.

kyngs:
The charset issues usually occur when you have misconfigured locales. Or at least that's when it happened in my case. In general, I don't think that forcing utf-8 will have any bad consequences.

PR #1064, opened 2023-08-21
Expose player info forwarding in the API ( Fix #1063 )

### LINKED ISSUE #1063: Expose forwarding related configuration data in ProxyConfig.
Currently, the ProxyConfig interface does not expose anything related to player info forwarding. This is quite unfortunate if you need to access the forwarding mode (and secret) from your plugin. (For example, to send the secret to a built-in limbo).

The workaround to this issue is quite simple, import the velocity API implementation, cast the configuration to VelocityConfiguration, and then you can get the forwarding config. But, I believe that it is quite unfortunate to need to use the velocity API implementation directly.

kyngs:
Fix #1063


hugmanrique (approved):
Thanks! 😀 (The CI fails due to an incorrect order of imports)

kyngs:
Hang on, gotta make checkstyle happy with the import order

Xernium (changes requested):
This requires approval by @astei as he made a very deliberate effort in ensuring this does **not** get exposed to the API.

kyngs:
> This requires approval by @astei as he made a very deliberate effort in ensuring this does **not** get exposed to the API.

May I ask the reason?

electronicboy:
The intention is to avoid exposing implementation details to plugins, especially given that things like Configuration are not API

I do wonder the use-case for this sorta thing, the forwarding mode should be effectively transparent to plugins, and have no real impacts elsewhere

kyngs:
> I do wonder the use-case for this sorta thing, the forwarding mode should be effectively transparent to plugins, and have no real impacts elsewhere

My practical use case is setting up nano limbo on-demand from the proxy. To do so, I need to forward the forwarding configuration to it.

kyngs:
Any update?

PR #1073, opened 2023-08-28
Add send command

Lunna5:
This pr adds a command to send players to others servers from console or in game with the permission `velocity.command.send`


kyngs:
I'm pretty sure that such a command is intentionally not included.

carlosleondev:
whaos

Lunna5:
> I'm pretty sure that such a command is intentionally not included.

I know Velocity isn't trying to be all-in-one, but I think it's a feature that is present in most servers that run under velocity.
Installing a plugin just for that feature seems a bit nonsensical to me, especially when it doesn't affect performance because in the end it's just a command that if you don't want to use it, just don't give permission and it would be like it wasn't there.
I fully understand that you do not want to merge it may seem stupid to you but take it more as a suggestion with the work done

electronicboy:
it's the general "do we want to maintain low effort 'it does a thing' commands in the proxy", especially when plugins exist and are generally much easier to work on than the proxy itself in terms of providing variance and fulfilling needs.

The lack of any command feedback on this at all is generally a non-starter to me, but, what would that even look like

Nacioszeczek:
afaik Bungee doesn't provide those commands in the proxy itself, instead there are plugins bundled with it that add them.
Would probably be better to - with permissions - bundle already existing plugins and not do redundant stuff.

Dreeam-qwq:
use https://github.com/Elytrium/VelocityTools

astei:
I'm going to be the bad guy and say this is fine to include.

PR #1078, opened 2023-09-07
Pass compressed packets directly to the backend server/client

JNNGL:
This passes compressed packets that are not present in the packet registry to the backend server/client without decompressing/recompressing them whenever possible; only the first 5 bytes are decompressed to get the packet id.


astei:
This has been proposed many times. By all means, this is not a bad idea. My main concern is that it could introduce overhead from decompressing to check the packet ID. zlib decompression is actually _really_ cheap, the actual CPU-intensive part is compressing the packet. Personally, I would rather have Mojang pull out the packet ID from compressed packets, but the window to do that was in 2013 and that's well past us now.

JNNGL:
> My main concern is that it could introduce overhead from decompressing to check the packet ID.

Wouldn't that still be faster than recompressing every packet?

astei:
> Wouldn't that still be faster than recompressing every packet?

Compressing small packets is still pretty cheap.

Could we have a carveout for packets with a compressed size lower than, say, 2048 bytes? Larger packets will benefit from this much more than smaller packets.

ytnoos:
Maybe make a configurable amount

PR #1082, opened 2023-09-17
Fix unsigned commands detected as signed

Joo200:
# The problem

Using the proxy to intercept commands is sometimes required and should work with the api and the CommandExecuteEvent.

Currently every denied command will result into a disconnect with the message of a invalid protocol state. Even if the command is not signed and only has none or unsigned arguments.

# The solution

To solve this problem I debugged the issue and found the reason: The salt is set for unsigned commands even if the command itself is not signed. This will cause a wrong return value for `SessionPlayerCommand#isSigned` and the `SessionCommandHandler` will not cancel the event correctly.

I changed the implmentation of `isSigned`and added some useful debug information in case of denied signed messages.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/chat/LastSeenMessages.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/chat/LastSeenMessages.java
@@ -49,4 +49,12 @@ public class LastSeenMessages {
   public boolean isEmpty() {
     return acknowledged.isEmpty();
   }
+
+  @Override
+  public String toString() {
+    return "LastSeenMessages{" +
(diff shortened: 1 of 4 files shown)


electronicboy:
I'm dubious on the nature of this given the current issues where having where cancelling anything ends up busting the chain more or less, even proxy commands appear to be doing so, yet I thought we had that handled just fine...

Joo200:
Yes, the current implementation has more issues and chat signing and the chain doesn't work at all when I'm using velocity as a proxy. I would prefer some proper fix to allow chat and command cancellation but until someone had the time this workaround should work fine.

electronicboy (approved):

PR #1085, opened 2023-09-19
feature: Cancellable connection establish event

kezz:
Don't know if it makes more sense to do something else that isn't blocking on this event firing - couldn't really think of a better way to handle it. Potential for this to be expanded in the future to distinguish between ping vs. connection.


api/src/main/java/com/velocitypowered/api/event/connection/ConnectionEstablishEvent.java
```diff
@@ -0,0 +1,46 @@
...
+    return this.connection;
+  }
+
+  @Override
+  public GenericResult getResult() {
+    return this.result;
+  }
+
+  @Override
+  public void setResult(final @NotNull GenericResult result) {
```
Dominik48N:
```suggestion
  public void setResult(@NotNull GenericResult result) {
```
In other event classes, the method arguments are not declared final. In order to ensure a uniform code style, I would remove the final here.

api/src/main/java/com/velocitypowered/api/event/connection/ConnectionEstablishEvent.java
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Event called when a connection is initially established with the proxy.
+ * Velocity will wait for this event to fire before continuing with the connection.
+ */
+@AwaitingEvent
+public class ConnectionEstablishEvent implements ResultedEvent<ResultedEvent.GenericResult> {
+  private final InboundConnection connection;
+  private GenericResult result = GenericResult.allowed();
+
+  public ConnectionEstablishEvent(final @NotNull InboundConnection connection) {
```
Dominik48N:
```suggestion
  public ConnectionEstablishEvent(@NotNull InboundConnection connection) {
```
In other event classes, the constructor arguments are not declared final. In order to ensure a uniform code style, I would remove the final here.

Dominik48N (changes requested):
Good pull request. A feature that you need, in my opinion.

Gerrygames:
The event should include the intention of the connection.
Also instead of blocking you should disabled auto-reading while the event is being fired ([like here for example](https://github.com/PaperMC/Velocity/blob/2bd2c692ea0294aaa77b8e5653d9d3344c30e487/proxy/src/main/java/com/velocitypowered/proxy/connection/backend/TransitionSessionHandler.java#L111-L161)).

kezz:
> In other event classes, the method arguments are not declared final. In order to ensure a uniform code style, I would remove the final here.

I'm just going to leave this one, it's a better code style that is less error prone. Other event classes, for example, do not have nullness annotations, we shouldn't follow that just because they use that style.

> The event should include the intention of the connection. Also instead of blocking you should disabled auto-reading while the event is being fired ([like here for example](https://github.com/PaperMC/Velocity/blob/2bd2c692ea0294aaa77b8e5653d9d3344c30e487/proxy/src/main/java/com/velocitypowered/proxy/connection/backend/TransitionSessionHandler.java#L111-L161)).

Thanks, edited this and it should be good now.

proxy/src/main/java/com/velocitypowered/proxy/connection/client/HandshakeSessionHandler.java
```diff
@@ -85,31 +86,57 @@ public boolean handle(LegacyHandshake packet) {
...
-          // If you get this, it's a bug in Velocity.
-          throw new AssertionError("getStateForProtocol provided invalid state!");
-      }
-    }
+
+    // Handle connection establish event.
+    connection.setAutoReading(false);
+    server.getEventManager()
+            .fire(new ConnectionEstablishEvent(
+                    ic, getIntentionForStatus(handshake.getNextStatus())))
```
astei: If `getIntentionForStatus` returns null, I would prefer if we just closed the connection right then and there. But I can understand why we might want to check for an invalid status.

4drian3d (approved):

PR #1087, opened 2023-09-24
Include forwarding secret file name in error message

Warriorrrr:
Includes the configured forwarding secret file name in parentheses in the error message when the file doesn't exist


Warriorrrr:
Apparently another PR already does this, so closing this

PR #1096, opened 2023-10-11
Support plugin message in configuration stage

### LINKED ISSUE #1095: Joining back-end Fabric server without content-adding mods yields "incomplete set of tags" disconnect message
When connecting to a back-end server running Fabric with only server-side mods, clients connecting through Velocity are being disconnected with the message: `Unable to connect to survival: Incomplete set of tags received from server. Please contact server operator.`

When the back-end server is made directly accessible, vanilla clients can connect, confirming that none of the server's mods add content to the game that can cause the 'incomplete tags' disconnect.

# Setup
- Proxy running Velocity 3.2.0-SNAPSHOT-269, no plugins.
- Backend server running Fabric Loader 0.14.23, with mods: `cloth-config`, `fabric-api`, `ferritecore`, `krypton`, `lazydfu`, `lithium`, `luckperms`, `starlight` and `vanilla-permissions`.

james58899:
Fabric-API will send **PluginMessage** in configuration stage and use **PingIdentify** to confirm the client is ready, so Velocity must transfer packets to the backend server.

Fix #1095

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientConfigSessionHandler.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientConfigSessionHandler.java
@@ -17,6 +17,7 @@
 
 package com.velocitypowered.proxy.connection.client;
 
+import com.velocitypowered.api.event.player.PlayerClientBrandEvent;
 import com.velocitypowered.proxy.VelocityServer;
 import com.velocitypowered.proxy.connection.MinecraftConnection;
 import com.velocitypowered.proxy.connection.MinecraftSessionHandler;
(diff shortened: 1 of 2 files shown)


proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientConfigSessionHandler.java
```diff
@@ -105,6 +108,31 @@ public boolean handle(FinishedUpdate packet) {
...
 
+  @Override
+  public boolean handle(PluginMessage packet) {
+    VelocityServerConnection serverConn = player.getConnectionInFlight();
+    if (serverConn != null) {
+      if (PluginMessageUtil.isMcBrand(packet)) {
+        String brand = PluginMessageUtil.readBrandMessage(packet.content());
+        server.getEventManager().fireAndForget(new PlayerClientBrandEvent(player, brand));
+        player.setClientBrand(brand);
+        // Client sends `minecraft:brand` packet immediately after Login,
```
electronicboy, line 119: We should probably look into ensuring that this is at least sent somewhere if we can't just send it right now; I need to spend some time looking over this, ideally we should race hard to get the connection to the server into configuration so we can deal with syncing stuff, but, that is probably a wider headache to deal with someday (hopefully soon™)

james58899: The client also won't send minecraft:brand when it re-enters the configuration stage after being connected, so I think it is safe to discard this packet.

In the design of Fabric-API, the client will only respond after receiving a packet from the server, so it should not lose any other packets.

ghost:
Bump, would really like to see this fixed

Tetricz:
I too would like to see this fixed.

james58899:
@electronicboy Can it be merged at the current state? At least this allows 1.20.2 with FabricAPI to connect.

Before this PR, the velocity would discard all PluginMessages, and the vanilla server didn't handle PluginMessages at all. So, I don't think this PR will break anything unless it's already broken.

4drian3d (approved):

DeathTrooperr:
Would love to see this fixed as well. Waiting to start a small Minecraft server network for my friends and I and this is the last thing missing.

astei (approved):

PR #1097, opened 2023-10-11
Improve/1.20.2

Gerrygames:


kennytv (approved):

api/src/main/java/com/velocitypowered/api/proxy/Player.java
```diff
@@ -148,9 +148,7 @@ public interface Player extends
   /**
    * Clears the tab list header and footer for the player.
    *
-   * @deprecated Use {@link TabList#clearHeaderAndFooter()}.
    */
-  @Deprecated
```
kashike, line 153: Can we introduce a new method instead of un-deprecating this one? `clearPlayerListHeaderAndFooter`

kashike (changes requested):

proxy/src/main/java/com/velocitypowered/proxy/connection/backend/TransitionSessionHandler.java
```diff
@@ -143,7 +144,9 @@ public boolean handle(JoinGame packet) {
           // Now set the connected server.
           serverConn.getPlayer().setConnectedServer(serverConn);
 
-          if (player.getClientSettingsPacket() != null) {
+          // Send client settings. In 1.20.2+ this is done in the config state.
```
VeVeVeVel, line 147: For 1.20.2, shouldn't ServerPostConnectEvent also be handled in the config state?

```
         // Send client settings. In 1.20.2+ this is done in the config state.
          if (smc.getProtocolVersion().compareTo(ProtocolVersion.MINECRAFT_1_20_2) < 0
              && player.getClientSettingsPacket() != null) {
            serverConn.ensureConnected().write(player.getClientSettingsPacket());
          }

          // We're done! :)
          server.getEventManager().fireAndForget(new ServerPostConnectEvent(player,
              previousServer));
```

i

electronicboy: I think that there was some debates over what "connected" is, but, I'd be inclined to generally agree, we should be considered connected as soon as we enter the point in which we should be transitioning the connection over for the client, for 1.20.1 and below, that is when it enters play, for 1.20.2+, config

kashike (approved):

proxy/src/main/java/com/velocitypowered/proxy/connection/backend/TransitionSessionHandler.java
Gerrygames: We would need to call the ServerConnectedEvent in the config state as well. However I wouldn't consider the config state as being fully connected to the server.
Ideally I would like to handle the config state similar to the login state. Don't switch the client into config state right away. Keep them connected to the old server. Only if a resource pack is sent by the backend server or set by API (would need some kind of new event). If the registry data of the old and new backend server is the same we wouldn't switch the client into config state at all.
But that's quite a lot of work.

electronicboy: > Ideally I would like to handle the config state similar to the login state.

this is long term just a suicide run, we have no sane way to diff the registry to say if something has changed, and given the changes there is a potential that we'd just induce further headaches as there is a potential we'd need to store info during the configuration process to say that "oh, something changed, we better replay that", which creates many headaches with stuff wanting to syncronise stuff, i.e. mods

it just sounds like a huge dance in which you start introducing many dozens of potential conditional flows that we'd endure and have to expect plugins to endure and ensure that we exposed this and all caveats it comes with in the API. It makes more sense just to adopt what is expected here and deal with one general set of expected state flows instead of trying to be overly smart about it and shooting ourselves in the foot in the process.

electronicboy: I think ideally we need to look into devising a nicer flow chain here, we've already had talks about letting the client and server connect to one another earlier for the sake of mods, given mojangs changes also leaning in that direction, it generally looks like, now is the time to do that, and hope that if modding environments follow, we get free compat, at least here

Gerrygames: Yeah I guess you are right. But I guess this is out of the scope of this PR.

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientPlaySessionHandler.java
```diff
@@ -482,12 +482,13 @@ public CompletableFuture<Void> doSwitch() {
       // Send keep alive to try to avoid timeouts
       player.sendKeepAlive();
 
-      // Reset Tablist header and footer to prevent desync
-      player.clearHeaderAndFooter();
+      // Config state clears everything in the client. No need to clear later.
```
electronicboy, line 485: just to check, this is not assumative towards 1.20.2 behavior? i.e. this isn't going to have a negagtive impact for older versions?

Gerrygames: Correct, doSwitch is only called for 1.20.2 players.

electronicboy: image

electronicboy (approved):

PR #1100, opened 2023-10-12
acknowledge seen messages to server when running proxy commands

electronicboy:
This fixes aspects like running /server from causing clients to disconnect the next time they interface with the server, as well as restores the ability to cancel signed chat messages on current versions. This will apply to at least 1.20, but, need to adjust the protocol checks and do some further checking

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/chat/session/SessionCommandHandler.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/chat/session/SessionCommandHandler.java
@@ -55,7 +55,7 @@ public class SessionCommandHandler implements CommandHandler<SessionPlayerComman
                   + "Contact your network administrator."));
         }
         // We seemingly can't actually do this if signed args exist, if not, we can probs keep stuff happy
-        if (player.getProtocolVersion().getProtocol() == ProtocolVersion.MINECRAFT_1_20_2.getProtocol()) {
+        if (player.getProtocolVersion().compareTo(ProtocolVersion.MINECRAFT_1_20_2) >= 0) {
           return CompletableFuture.completedFuture(new ChatAcknowledgement(packet.lastSeenMessages.getOffset()));
         }
         return CompletableFuture.completedFuture(null);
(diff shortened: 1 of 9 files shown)


electronicboy:
So, I have command handling working on the proxy for 1.20.2, cancelling chat doesn't work, but, I think the issue is obvious, I need to store the old offset and use that rather than grabbing the one in the packet

electronicboy:
So, it seems that we can only use this for preventing cases where we're not actually seeing signed stuff, I am probably going to merge this (soon) as is in order to fix command handling on the proxy, might be nice for somebody else to test this, need to add support for some older versions, forgot what ver mojang added the packet we're using here in

proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/chat/session/SessionCommandHandler.java
```diff
@@ -52,6 +54,10 @@ public void handlePlayerCommandInternal(SessionPlayerCommand packet) {
               "A proxy plugin caused an illegal protocol state. "
                   + "Contact your network administrator."));
         }
+        // We seemingly can't actually do this if signed args exist, if not, we can probs keep stuff happy
+        if (player.getProtocolVersion().getProtocol() == ProtocolVersion.MINECRAFT_1_20_2.getProtocol()) {
```
Joo200: Be optimistic that the packet is there in 1.20.3 too.
```suggestion
        if (player.getProtocolVersion().compareTo(ProtocolVersion.MINECRAFT_1_20_2) >= 0) {
```

proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/chat/session/SessionCommandHandler.java
```diff
@@ -102,6 +108,9 @@ public void handlePlayerCommandInternal(SessionPlayerCommand packet) {
                 .toServer();
           }
         }
+        if (player.getProtocolVersion().getProtocol() == ProtocolVersion.MINECRAFT_1_20_2.getProtocol()) {
```
Joo200: Be optimistic that the packet is there in 1.20.3 too.
```suggestion
        if (player.getProtocolVersion().compareTo(ProtocolVersion.MINECRAFT_1_20_2) >= 0) {
```

PR #1103, opened 2023-10-12
Fix tablist header and footer desync

### LINKED ISSUE #1101: Tablist header and footer are not reset when a player using client 1.20.1 or lower is kicked from the backend.
Tablist header and footer are not reset when a player using client 1.20.1 or lower is kicked from the backend.

VeVeVeVel:
Resolves #1101


VeVeVeVel:
This happens because 'existingConnection' is null when the player is kicked from the backend.
If there is a better solution than this, please apply it.

electronicboy (approved):

RealBauHD:
This pr causes the problem that the tablist of plugins that set it in the "PostLoginEvent" is immediately reset again.

electronicboy:
PostLoginEvent was never a place that we'd expect anybody to try setting that stuff given that the client is not even in a state to recieve such data

RealBauHD:
Oh yeah that's actually true, did it in an old plugin before the config phase was invented

PR #1022, opened 2023-06-12
Implement ComponentLogger

4drian3d:

PR #1105, opened 2023-10-12
Removed console legacy color support

4drian3d:
Now that there is a stable way to display components in the console #1022 , it is no longer necessary to support the legacy format


proxy/src/main/resources/log4j2.xml
```diff
@@ -22,18 +22,18 @@
...
+            pattern="%highlightError{[%d{HH:mm:ss} %level]: %msg%n%xEx}"/>
         </LoggerNamePatternSelector>
       </PatternLayout>
     </TerminalConsole>
     <RollingRandomAccessFile name="File" fileName="logs/latest.log"
       filePattern="logs/%d{yyyy-MM-dd}-%i.log.gz"
       immediateFlush="false">
       <PatternLayout
-        pattern="[%d{HH:mm:ss}] [%t/%level] [%logger]: %minecraftFormatting{%msg}{strip}%n"/>
+        pattern="[%d{HH:mm:ss}] [%t/%level] [%logger]: %msg%n"/>
```
kezz: What did the `{strip}` do here?

4drian3d: Removes ansi color formatting when saving logs. https://github.com/Minecrell/TerminalConsoleAppender/blob/master/src/main/java/net/minecrell/terminalconsole/MinecraftFormattingConverter.java#L126

rymiel: The [corresponding paper patch](https://github.com/PaperMC/Paper/pull/9313/files#diff-ef193170056b3e41ea067f8e473d2c7e675b85ef9ee65f2f8cd94b65152f6ba9L61-R61) added a [`stripAnsi`](https://github.com/PaperMC/Paper/pull/9313/files#diff-6c1396d60730e7053f0b761bdb487752467c9bc0444a4aac41908376b38a56bdR198-R233) processor. Is that not required here?

4drian3d: Of course, yesterday after 1 year I was trying again to make a Log4j2 plugin for Velocity but this time to remove the ansi codes, but it is still being impossible because it is not being recognized as it was last year https://github.com/PaperMC/Velocity/pull/718#issuecomment-1226019613

This processor is necessary even without this pull request because the minecraftFormatting plugin only avoids adding the codes that were converted from the legacy format, it does not remove the ones that have already been implemented. I will keep trying to fix that, but anyone is free to implement a workaround if they can.

rymiel: Thanks for the clarification, I wasn't aware of your earlier struggles, but good luck

4drian3d: implemented

electronicboy (approved):

electronicboy (approved):

PR #1106, opened 2023-10-12
Draft: Use PluginMessage to allow Velocity to cancel signed chat and commands

Joo200:
Add some logic to send cancelled signed messages/commands to the backend server.

This is a draft implementation and could be further improved, e.g. by a config option.

There is already an open PR in Paper for this feature: https://github.com/PaperMC/Paper/pull/9828

Note: I removed the version checks in SessionCommandHandler because that handler is only created for versions >= 1.19.3 (see ClientPlaySessionHandler)


Joo200:
I marked this as Ready. I'm using it on my server in production and it works perfectly fine.

PR #1130, opened 2023-10-29
Fix incorrectly merged buildscript changes from dev/3.0.0

alexstaeding:
It seems some changes from #983 were missed when merging into dev/5.0.0

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,2 +1,2 @@
 group=com.velocitypowered
-version=3.2.0-SNAPSHOT
+version=5.0.0-SNAPSHOT
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,26 +1,2 @@
-# API dependencies.
-gsonVersion=2.10.1
-junitVersion=5.9.0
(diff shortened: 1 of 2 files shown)


gradle.properties
```diff
@@ -1,26 +1,2 @@
...
-nightConfigVersion=3.6.6
-completableFuturesVersion=0.3.5
-adventureFacetVersion=4.2.0
-fastutilVersion=8.5.12
-disruptorVersion=3.4.4
-jlineVersion=3.23.0
-terminalConsoleAppenderVersion=1.3.0
-joptSimpleVersion=5.0.4
+group=com.velocitypowered
+version=3.2.0-SNAPSHOT
```
powercasgamer:
```suggestion
version=5.0.0-SNAPSHOT
```

powercasgamer (changes requested):

gradle.properties
alexstaeding: Oh right, don't know how I missed that

powercasgamer (approved):

Issue #1134, opened 2023-10-29
MEDIUM: Use of AccessController.doPrivileged() in Public Method of Public Class


DawmosTomie:
The code will execute with system privileges and should be manually checked with great care to ensure no vulnerabilities are present.

Line: 539 - Filename: Velocity\proxy\src\main\java\com\velocitypowered\proxy\VelocityServer.java

`AccessController.doPrivileged(new PrivilegedAction<Void>() {`

electronicboy:
Not an issue, this code is in the shutdown path as is expected, only fix here would basically be to just appeasing overzelous scanning tools

PR #1135, opened 2023-10-29
MEDIUM: Use of AccessController.doPrivileged() in Public Method of Public Class

DawmosTomie:
issue：https://github.com/PaperMC/Velocity/issues/1134#issue-1967025499

In this change, I created a new method for private exitWithPrivilege (), and the AccessController. The doPrivileged () call mobile to this method. The new method is then called in shutdownProcess. So you can avoid the common methods of directly using the AccessController. The doPrivileged (), so as to improve the security of the code.


electronicboy:
This generally changes nothing outside of appeasing overzealous automated analysis tooling

PR #1136, opened 2023-11-05
Add -haproxy-protocol startup parameter

### LINKED ISSUE #1024: Add -haproxy-protocol startup parameter
I know that there is a startup parameter of -port, but in some cases, multiple velocities may be synchronized with git, some of which will enable haproxy-protocol, and some will not.

This is a small update, thanks.

BingYanchi:
resolves https://github.com/PaperMC/Velocity/issues/1024


proxy/src/main/java/com/velocitypowered/proxy/ProxyOptions.java
```diff
@@ -34,6 +34,7 @@ public final class ProxyOptions {
   private static final Logger logger = LogManager.getLogger(ProxyOptions.class);
   private final boolean help;
   private final @Nullable Integer port;
+  private final Boolean haproxy;
```
powercasgamer:
```suggestion
  private final boolean haproxy;
```

proxy/src/main/java/com/velocitypowered/proxy/ProxyOptions.java
```diff
@@ -64,4 +71,8 @@ boolean isHelp() {
   public @Nullable Integer getPort() {
     return this.port;
   }
+
+  public Boolean isHaproxy() {
```
powercasgamer:
```suggestion
  public boolean isHaproxy() {
```

powercasgamer (changes requested):

proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
```diff
@@ -243,6 +243,12 @@ void start() {
       this.cm.bind(configuration.getBind());
     }
 
+    final Boolean haproxy = this.options.isHaproxy();
+    if (haproxy != null) {
```
electronicboy, line 261: This is always true, as you declared the method return to be a primative

4drian3d (approved):

BingYanchi:
```
Exception in thread "main" java.lang.NullPointerException: Cannot read the array length because "<local2>" is null
        at joptsimple.ArgumentAcceptingOptionSpec.defaultsTo(ArgumentAcceptingOptionSpec.java:225)
        at com.velocitypowered.proxy.ProxyOptions.<init>(ProxyOptions.java:51)
        at com.velocitypowered.proxy.Velocity.main(Velocity.java:62)
```
There is a problem with this PR and it needs to be rolled back.

PR #1137, opened 2023-11-05
Log console command executions

Joo200:
Currently command executions are only logged for players. Imo this config option should include the console because not every command has a sender feedback and it could be nice to see it in the logs.


novitpw:
And what commands may not have a feedback the sender? Don't you think this is already a plugin problem?

Joo200:
> And what commands may not have a feedback the sender? Don't you think this is already a plugin problem?

Yes, plugins should provide feedback but depending on the command there is none or not enough feedback to know which command was executed. Why should a plugin repeat the executed command, the executor already know the command.

So no, that's not directly a plugin issue.

4drian3d (approved):

electronicboy (approved):

electronicboy:
Some considersations that maybe this should be a separate config option for the console though; rewriting all the config stuff is a WIP which I'll get back to when I'm not too dizzy

PR #1147, opened 2023-11-29
fix backend not receiving brands from 1.20.2 clients

### LINKED ISSUE #1131: Cannot get client bran name behind proxy
### Expected behavior

Able to get client brand name

### Observed/Actual behavior

Client brand name is null

### Steps/models to reproduce

Run paper server with 
- player.getClientBrand() will return what client player use

When run paper server behind velocity
- player.getClientBrand() will return null

When use the plugin to run on Velocity
- player.getClientBrand() will return what client player use

plugin is just
    @EventHandler
    public void onJoin(PlayerJoinEvent event) {
        Bukkit.getScheduler().runTaskLater(this, () -> {
            System.out.println(event.getPlayer().getClientBrandName());
        }, 100L);
       
i dont know if this issue is on velocity or paper or in between.

### Plugin and Datapack List

[23:39:12 INFO]: Server Plugins (1):
[23:39:12 INFO]: Bukkit Plugins:
[23:39:12 INFO]:  - Test_Plugin

Just plugin for getting client brand

### Paper version

This server is running Paper version git-Paper-248 (MC: 1.20.2) (Implementing API version 1.20.2-R0.1-SNAPSHOT) (Git: b974a6b)

### Other

-

AoElite:
Is this the correct way to do this? I have no idea. Does it work? Yes.
If this isn't acceptable then let me know how to do this properly.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientConfigSessionHandler.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientConfigSessionHandler.java
@@ -32,6 +32,7 @@ import com.velocitypowered.proxy.protocol.packet.ResourcePackResponse;
 import com.velocitypowered.proxy.protocol.packet.config.FinishedUpdate;
 import com.velocitypowered.proxy.protocol.util.PluginMessageUtil;
 import io.netty.buffer.ByteBuf;
+import io.netty.buffer.Unpooled;
 import java.util.concurrent.CompletableFuture;
 import net.kyori.adventure.text.Component;
 import net.kyori.adventure.text.format.NamedTextColor;
@@ -46,7 +47,7 @@ public class ClientConfigSessionHandler implements MinecraftSessionHandler {
   private static final Logger logger = LogManager.getLogger(ClientConfigSessionHandler.class);
   private final VelocityServer server;
   private final ConnectedPlayer player;
-  private PluginMessage brandPacket = null;
+  private String brandChannel = null;
 
   private CompletableFuture<Void> configSwitchFuture;
 
@@ -117,9 +118,9 @@ public class ClientConfigSessionHandler implements MinecraftSessionHandler {
         String brand = PluginMessageUtil.readBrandMessage(packet.content());
         server.getEventManager().fireAndForget(new PlayerClientBrandEvent(player, brand));
         player.setClientBrand(brand);
+        brandChannel = packet.getChannel();
         // Client sends `minecraft:brand` packet immediately after Login,
         // but at this time the backend server may not be ready
-        brandPacket = packet.retain();
       } else {
         serverConn.ensureConnected().write(packet.retain());
       }
@@ -184,11 +185,16 @@ public class ClientConfigSessionHandler implements MinecraftSessionHandler {
    * @return a future that completes when the config stage is finished
    */
   public CompletableFuture<Void> handleBackendFinishUpdate(VelocityServerConnection serverConn) {
-    if (brandPacket != null) {
+    String brand = serverConn.getPlayer().getClientBrand();
+    if (brand != null && brandChannel != null) {
+      PluginMessage brandPacket = new PluginMessage(
+              brandChannel, Unpooled.wrappedBuffer(brand.getBytes()));
+
       serverConn.ensureConnected().write(
-                  PluginMessageUtil.rewriteMinecraftBrand(brandPacket, server.getVersion(),
-                          player.getProtocolVersion()));
+              PluginMessageUtil.rewriteMinecraftBrand(brandPacket.retain(), server.getVersion(),
+                      player.getProtocolVersion()));
     }
+
     player.getConnection().write(new FinishedUpdate());
     serverConn.ensureConnected().write(new FinishedUpdate());
     return configSwitchFuture;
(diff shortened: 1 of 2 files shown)


electronicboy (approved):

electronicboy:
a "proper" fix is more likely going to involve rewriting the entire process between the client and the server, so, in the meantime...

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientConfigSessionHandler.java
```diff
@@ -117,7 +118,8 @@ public boolean handle(PluginMessage packet) {
         server.getEventManager().fireAndForget(new PlayerClientBrandEvent(player, brand));
         player.setClientBrand(brand);
         // Client sends `minecraft:brand` packet immediately after Login,
-        // but at this time the backend server may not be ready, just discard it.
+        // but at this time the backend server may not be ready
+        brandPacket = packet.retain();
```
electronicboy: Actually, if we retain, we should probably release this inside of the handlers inactive method

electronicboy (changes requested):

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientConfigSessionHandler.java
AoElite: Which method are you referring to exactly?

electronicboy: I think that the better solution would be not to try to retain this, but to just send a new one, otherwise, would need to be inside of deactivate and handleDIsconnect

AoElite: Should I recreate the packet entirely or can I just .copy() it?

AoElite:
```java
  public CompletableFuture<Void> handleBackendFinishUpdate(VelocityServerConnection serverConn) {
    String brand = serverConn.getPlayer().getClientBrand();
    if (brand != null) {
      PluginMessage brandPacket = new PluginMessage("minecraft:brand", Unpooled.wrappedBuffer(brand.getBytes()));

      serverConn.ensureConnected().write(
              PluginMessageUtil.rewriteMinecraftBrand(brandPacket.retain(), server.getVersion(),
                      player.getProtocolVersion()));
    }

    player.getConnection().write(new FinishedUpdate());
    serverConn.ensureConnected().write(new FinishedUpdate());
    return configSwitchFuture;
  }
  ```
  How about this? I'd probably have to keep track of what channel they used as well for version compatibility.

Comment (electronicboy) [reply proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientConfigSessionHandler.java]: Yes, that generally looks better, we don't need to worry about ensuring that we don't leak the brand packet we got

Comment (AoElite) [reply proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientConfigSessionHandler.java]: Pushed changes, I'm not sure if I need to use .retain() or not here though, and if it's ok to store the player's channel brand in that class.

Comment (electronicboy) [reply proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientConfigSessionHandler.java]: I don't think that calling retain is needed, but, I'm not 100% sure; storing the chanel name should be fine, we should only really see the minecraft:brand one in there, but 🤷

Comment (AoElite) [reply proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientConfigSessionHandler.java]: I've removed it. Doesn't seem to be necessary after testing.

**Review by electronicboy — APPROVED** [review]

**Review by electronicboy — CHANGES_REQUESTED** [review]
actually, one derp, the brand message itself is written as length prefixed (i.e. should use `writeUtf`), also, this shouldn't call rewriteMinecraftBrand, that's for S>C in order to inject the velocity brand into the thing

**Comment by AoElite** [comment]
@electronicboy On the current version of velocity player minecraft brands sent to the servers are rewritten with (Velocity) infront of them. It's always done this as long as I remember. Are you saying this is actually a bug?

**Comment by electronicboy** [comment]
Imho, yes, that would be a bug

**Comment by AoElite** [comment]
I thought it was intentional as it's been like this for years. You sure I should change that in this PR?

**Comment by electronicboy** [comment]
@astei 
Imho, it makes 0 sense that the C>S should have this branding on it, the entire point is that post proxy specific stuff, i.e. forwarding info, the nature of the proxy should generally be transparent, it shouldn't impact stuff like this

**Comment by AoElite** [comment]
@electronicboy I've made a commit to fix that issue. I've also tested it to make sure it doesn't affect the server's brand.

File: proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientConfigSessionHandler.java [hunk 2]
```diff
@@ -182,6 +185,13 @@ public void exception(Throwable throwable) {
    * @return a future that completes when the config stage is finished
    */
   public CompletableFuture<Void> handleBackendFinishUpdate(VelocityServerConnection serverConn) {
+    String brand = serverConn.getPlayer().getClientBrand();
+    if (brand != null && brandChannel != null) {
+      PluginMessage brandPacket = new PluginMessage(
+              brandChannel, Unpooled.wrappedBuffer(brand.getBytes()));
```
electronicboy: This should not be the raw bytes, it should be written as a utf string to the buffer

electronicboy (changes requested):

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientConfigSessionHandler.java
AoElite: Alright done

electronicboy (approved):

PR #1151, opened 2023-12-07
switch from Collections and Guava Collections to Javas

RealBauHD:


4drian3d (approved):

astei:
I do not see the benefit in doing this. Guava's immutable collections aren't going to go away despite JEP 269.

PR #1163, opened 2023-12-18
chore: update adventure

RealBauHD:


RealBauHD:
The ByteBuf Holder in 'Registry Sync' can now be replaced by nameless Compound reading, right?

RealBauHD:
I wait for some feedback regarding which way I should go with resource packs.

EpicPlayerA10:
> I wait for some feedback regarding which way I should go with resource packs.

At first seems good to me. I was thinking of moving the whole resourcepack logic to some kind of `ResourcepackManager` because the code, that is handling resourcepacks, grows very quickly.

electronicboy:
I believe that theres been a few attempts to try to deal with this API, I believe adrian got the furtherst in terms of actually implementing stuff, but, theres a lot of headaches to deal with, i.e. this PR generally seems to forget the behavioral differences between modern versions and latest which need to be looked into properly and handled as such

PR #1166, opened 2023-12-25
Per server IP & profile forwarding configuration

### LINKED ISSUE #304: Playerdata forwarding
Add the option for users to specify on a per server basis if they want to use the old bungee cord forwarding method.

Or the new better one.

This would allow hybrid network of more secure newer servers.
And legacy support for older servers on the network.

e.g.
Server1: 
   -'Modern Skyblock 2'
 bungeelegacy: true
Server2:
   -'Vanilla1_15

For security it should either default to modern or autodetect,

paulzzh:
Make it possible to configure forwarding method for each server.
tests:
fabric 1.20.4 + modern : ok
forge 1.12.2 with spongeforge + legacy : ok
Can not test forge 1.13+ because some classes moved and [Ambassador](https://github.com/adde0109/Ambassador) break.

Fix #304
Make #291 mergeable.

config file
<details>
player-info-forwarding-mode = "none" #set default here

[servers]

Modpack.address = "127.0.0.1:26665"
Modpack.forwarding = "legacy" #use bungeecord legacy forwarding

Vanilla.address = "127.0.0.1:26666"
Vanilla.forwarding = "modern" #use velocity modern forwarding

Sky = "127.0.0.1:26667" #default
</details>


denisapain:
pls

paulzzh:
#1357

PR #1172, opened 2023-12-28
Restore nullability for displayName in UpsertPlayerInfo

Timongcraft:
This makes the UpsertPlayerInfo's displayName truly nullable as before the ComponentHolder was introduced.


LucidAPs:
Merge this please!

electronicboy (approved):

PR #1173, opened 2023-12-28
full support for UnixDomainSocket backend

Coloryr:
Support paper backend with UnixDomainSocket
**Is must work with [this](https://github.com/PaperMC/Paper/pull/10088)**
{4E839E59-23CF-4fa3-B0F9-2C75F679304E}
image

refer to https://github.com/PaperMC/Velocity/pull/991

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/backend/VelocityServerConnection.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/backend/VelocityServerConnection.java
@@ -42,6 +42,7 @@ import com.velocitypowered.proxy.protocol.packet.Handshake;
 import com.velocitypowered.proxy.protocol.packet.PluginMessage;
 import com.velocitypowered.proxy.protocol.packet.ServerLogin;
 import com.velocitypowered.proxy.server.VelocityRegisteredServer;
+import io.netty.bootstrap.Bootstrap;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
 import io.netty.channel.ChannelFutureListener;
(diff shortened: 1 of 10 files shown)


Owen1212055:
Seems to already be in 5.0.0 as was done in https://github.com/PaperMC/Velocity/pull/991 ?

Coloryr:
> Seems to already be in 5.0.0 as was done in #991 ?

It's look like change the bind, instead of backend support

Coloryr:
My modification is for the `group server`

Coloryr:
My idea is `bind` not use UDS, and use UDS for `backend server` transport

proxy/src/main/java/com/velocitypowered/proxy/connection/backend/VelocityServerConnection.java
```diff
@@ -140,13 +152,26 @@ String getPlayerRemoteAddressAsString() {
...
+  }
+
   private String createLegacyForwardingAddress(UnaryOperator<List<Property>> propertiesTransform) {
     // BungeeCord IP forwarding is simply a special injection after the "address" in the handshake,
     // separated by \0 (the null byte). In order, you send the original host, the player's IP, their
     // UUID (undashed), and if you are in online-mode, their login properties (from Mojang).
-    StringBuilder data = new StringBuilder().append(proxyPlayer.getVirtualHost().orElseGet(() ->
-                    registeredServer.getServerInfo().getAddress()).getHostString())
-        .append('\0')
+    StringBuilder data = new StringBuilder().append(getRemoteAddress())
```
electronicboy, line 174: this changes the intended behavior here

proxy/src/main/java/com/velocitypowered/proxy/connection/backend/VelocityServerConnection.java
```diff
@@ -174,9 +199,10 @@ private void startHandshake() {
 
     // Initiate the handshake.
     ProtocolVersion protocolVersion = proxyPlayer.getConnection().getProtocolVersion();
-    String playerVhost =
-        proxyPlayer.getVirtualHost().orElseGet(() -> registeredServer.getServerInfo().getAddress())
-            .getHostString();
+//    String playerVhost =
+//        proxyPlayer.getVirtualHost().orElseGet(() -> registeredServer.getServerInfo().getAddress())
+//            .getHostString();
+    String playerVhost = getRemoteAddress();
```
electronicboy, line 206: no

proxy/src/main/java/com/velocitypowered/proxy/network/TransportType.java
```diff
@@ -25,6 +25,8 @@
 import io.netty.channel.epoll.EpollEventLoopGroup;
 import io.netty.channel.epoll.EpollServerSocketChannel;
 import io.netty.channel.epoll.EpollSocketChannel;
+import io.netty.channel.epoll.EpollDomainSocketChannel;
```
electronicboy, line 28: epoll is epoll, we can't just slap it on every single connection type, that is not how that works

proxy/src/main/java/com/velocitypowered/proxy/connection/backend/VelocityServerConnection.java
Coloryr: ok

Coloryr: unix socket hoststring like `/xxx/xxx` is a filename, but inetsocketaddress.getHostString is a ipaddress, because of these two differences, resulting in different behaviors?

Coloryr: oh i missing a `.append('\0')`

proxy/src/main/java/com/velocitypowered/proxy/network/TransportType.java
Coloryr: so i patch the paper

Coloryr: so i patch the paper

Lordfirespeed:
the referenced pull request to paperMC has been closed; what's the situation here?

PR #1174, opened 2023-12-28
Use monotonic time source for comparing timestamps

0xAda:
When calculating ping and startup time, velocity uses `System.currentTimeMillis`, if the system time changes due to NTP, Daylight Savings Time or just manual changes, this will make the calculations incorrect. Using `System.nanoTime` fixes this.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/Velocity.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/Velocity.java
@@ -20,6 +20,7 @@ package com.velocitypowered.proxy;
 import io.netty.util.ResourceLeakDetector;
 import io.netty.util.ResourceLeakDetector.Level;
 import java.text.DecimalFormat;
+import java.util.concurrent.TimeUnit;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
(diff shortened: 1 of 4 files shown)


4drian3d (approved):
Except for the initialization time of Velocity where it is not so necessary, it seems a good change

electronicboy:
nanoTime is slower to call than currentTimeMillis, and none of these really look like a huge deal given that they're mostly used for identity or display, outside of ping?

0xAda:
Which system are you measuring that nanoTime is slower than
currentTimeMillis on?

Writing a quick JMH benchmark, locally I'm getting 20 nanoseconds per call
for both methods (5950x, JDK 17, Kernel 6.1). On any Linux system with vdso
(past 8 years or so iirc) nanoTime should be very fast, based on the date I
believe https://shipilev.net/blog/2014/nanotrusting-nanotime/ is written
before vdso and measures a nanoTime call at around 30 nanoseconds. An old
oracle blog
https://web.archive.org/web/20140528053206/https://blogs.oracle.com/dholmes/entry/inside_the_hotspot_vm_clocks
suggests calls on windows are similar but like any system will take
microseconds in the worst case, I would assume the worst case is much less
common on modern hardware & windows versions as that article is over a
decade old.

I wouldn't call nanoTime in a very hot and tight loop, but in this case
where it's ran every few seconds, I don't think there will be a
measurable difference.

electronicboy:
Well, I guess most platforms of care optimise the clock, ik freebsd doesn't or didn't for a long time, and I know linux has in the past disabled the thing

in reality, we're not calling it enough for it to be a big deal vs say, something like timings on paper

astei (approved):
I am fine with this. The performance concerns are a nothingburger - we already recommend people run Velocity on modern Linux platforms anyway where we have vDSO.

PR #1176, opened 2023-12-29
Add ModernForgeConnectionType to supports Forge-1.20.2+

Mgazul:
~~This is incomplete support. Although it can correctly identify Forge as a Forge client, there was a Packets error that I couldn't solve, which is also the reason why I sent the PR. I hope more people can solve this problem~~
This PR is fully working^_^

Testing environment:
* mohist-1.20.2 Latest Build
* velocity-3.3.0 Latest Build
* Pixelmon-1.20.2-9.2.5-universal

image
From Mohist logs, it can be seen that it has been successfully identified as ConnectionType MODDED and connect to the server

~~The final result is that the server cannot be accessed:  exception encountered in com.velocitypowered.proxy.connection.backend.BackendPlaySessionHandler@ee2fab1
com.velocitypowered.proxy.util.except.QuietRuntimeException: A packet did not decode successfully (invalid data). For more information, launch Velocity with -Dvelocity.packet-decode-logging=true to see more.
https://haste.mohistmc.com/emarutavac.css~~

This PR is different from most fix plugins and mods:
They are using legacy mode
But we forwarded FORGE directly in modern mode

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/backend/VelocityServerConnection.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/backend/VelocityServerConnection.java
@@ -19,6 +19,7 @@ package com.velocitypowered.proxy.connection.backend;
 
 import static com.velocitypowered.proxy.VelocityServer.GENERAL_GSON;
 import static com.velocitypowered.proxy.connection.forge.legacy.LegacyForgeConstants.HANDSHAKE_HOSTNAME_TOKEN;
+import com.velocitypowered.proxy.connection.forge.modern.ModernForgeConstants;
 import static com.velocitypowered.proxy.network.Connections.HANDLER;
 
 import com.google.common.base.Preconditions;
@@ -188,6 +189,8 @@ public class VelocityServerConnection implements MinecraftConnectionAssociation,
       handshake.setServerAddress(createBungeeGuardForwardingAddress(secret));
     } else if (proxyPlayer.getConnection().getType() == ConnectionTypes.LEGACY_FORGE) {
       handshake.setServerAddress(playerVhost + HANDSHAKE_HOSTNAME_TOKEN);
+    } else if (proxyPlayer.getConnection().getType() == ConnectionTypes.MODERN_FORGE) {
+      handshake.setServerAddress(playerVhost + ModernForgeConstants.MARKER);
     } else {
       handshake.setServerAddress(playerVhost);
     }
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/backend/VelocityServerConnection.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/backend/VelocityServerConnection.java
@@ -19,7 +19,6 @@ package com.velocitypowered.proxy.connection.backend;
 
 import static com.velocitypowered.proxy.VelocityServer.GENERAL_GSON;
 import static com.velocitypowered.proxy.connection.forge.legacy.LegacyForgeConstants.HANDSHAKE_HOSTNAME_TOKEN;
-import static com.velocitypowered.proxy.connection.forge.modern.ModernForgeConstants.MODERN_HANDSHAKE_HOSTNAME_TOKEN;
 import static com.velocitypowered.proxy.network.Connections.HANDLER;
 
 import com.google.common.base.Preconditions;
@@ -37,6 +36,7 @@ import com.velocitypowered.proxy.connection.MinecraftConnection;
 import com.velocitypowered.proxy.connection.MinecraftConnectionAssociation;
 import com.velocitypowered.proxy.connection.MinecraftSessionHandler;
 import com.velocitypowered.proxy.connection.client.ConnectedPlayer;
+import com.velocitypowered.proxy.connection.forge.modern.ModernForgeConstants;
 import com.velocitypowered.proxy.connection.util.ConnectionRequestResults.Impl;
 import com.velocitypowered.proxy.protocol.StateRegistry;
 import com.velocitypowered.proxy.protocol.packet.Handshake;
@@ -190,7 +190,7 @@ public class VelocityServerConnection implements MinecraftConnectionAssociation,
     } else if (proxyPlayer.getConnection().getType() == ConnectionTypes.LEGACY_FORGE) {
       handshake.setServerAddress(playerVhost + HANDSHAKE_HOSTNAME_TOKEN);
     } else if (proxyPlayer.getConnection().getType() == ConnectionTypes.MODERN_FORGE) {
-      handshake.setServerAddress(playerVhost + MODERN_HANDSHAKE_HOSTNAME_TOKEN);
+      handshake.setServerAddress(playerVhost + ModernForgeConstants.getModernForgeHostnameToken());
     } else {
       handshake.setServerAddress(playerVhost);
     }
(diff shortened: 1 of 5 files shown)


electronicboy:
The packet error you're seeing is because a mod uses a custom argument type, and velocity needs to be able to decode the commands packet in order to inject its own command defs into there, 

cross ref, https://github.com/VelocityPowered/CrossStitch basically

SimonMage:
> The packet error you're seeing is because a mod uses a custom argument type, and velocity needs to be able to decode the commands packet in order to inject its own command defs into there,
> 
> cross ref, https://github.com/VelocityPowered/CrossStitch basically

This mod is only for fabric, you mean that is necessary to implement similar method into mohist for having modern forge compatibility?

electronicboy:
Yes

SimonMage:
> Yes

Will the pull request be accepted?
I think without it, nothing can work

electronicboy:
pull request to what?
This PR, I have nothing against it, but no commitments to it

SimonMage:
> pull request to what? This PR, I have nothing against it, but no commitments to it

He implemented a check for modern forge support, only with this merge, velocity can check if a client is modded or not in 1.20.2+

Mgazul:
> The packet error you're seeing is because a mod uses a custom argument type, and velocity needs to be able to decode the commands packet in order to inject its own command defs into there,
> 
> cross ref, https://github.com/VelocityPowered/CrossStitch basically

Thanks for your help, I have modified mohist and it is perfectly compatible with the pr

M4RC0Sx:
The work done by @Mgazul  is tremendous and should be included in Velocity by default. The Minecraft server community with mods needs stability and effective server management solutions.

proxy/src/main/java/com/velocitypowered/proxy/connection/forge/modern/ModernForgeConstants.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+package com.velocitypowered.proxy.connection.forge.modern;
+
+/**
+ * Constants for use with Modern Forge systems.
+ */
+public class ModernForgeConstants {
+
+    public static final String MARKER = "\0FORGE";
```
electronicboy: Would prefer this name to be marginally more verbose if it's going to be statically imported

Mgazul: Changes have been made

proxy/src/main/java/com/velocitypowered/proxy/connection/backend/VelocityServerConnection.java
```diff
@@ -188,6 +189,8 @@ private void startHandshake() {
       handshake.setServerAddress(createBungeeGuardForwardingAddress(secret));
     } else if (proxyPlayer.getConnection().getType() == ConnectionTypes.LEGACY_FORGE) {
       handshake.setServerAddress(playerVhost + HANDSHAKE_HOSTNAME_TOKEN);
+    } else if (proxyPlayer.getConnection().getType() == ConnectionTypes.MODERN_FORGE) {
+      handshake.setServerAddress(playerVhost + MODERN_HANDSHAKE_HOSTNAME_TOKEN);
```
IzzelAliz: Per Forge's implementation, it is legal to add net version after the `FORGE` marker, which should also be forwarded.

https://github.com/MinecraftForge/MinecraftForge/blob/f09fd1f05b39df98f6415b44478b13723cee74cd/src/main/java/net/minecraftforge/network/NetworkContext.java#L96-L110

Mgazul: There is currently no need to forward the version. You need to wait for subsequent forge updates before making a decision.

IzzelAliz: It can happen that your implementation forward only `FORGE` when a newer `FORGE2` client connects. Given the facts forge has bump the net version to 3 in prior 1.20.1, this can be considered as bug. You can either reject clients with extra digits after the marker or forward the full content, but not forward part.

IzzelAliz: Given the facts forge has bump the net version to 3 in prior 1.20.1, I prefer implement it properly and save a future pr.

electronicboy: Looks like supporting that is pretty much an essential thing, as otherwise it will fail if they ever bump that

Mgazul: Completed

proxy/src/main/java/com/velocitypowered/proxy/connection/client/HandshakeSessionHandler.java
```diff
@@ -152,6 +153,13 @@ private void handleLogin(Handshake handshake, InitialInboundConnection ic) {
   }
 
   private ConnectionType getHandshakeConnectionType(Handshake handshake) {
+    if (handshake.getServerAddress().contains(ModernForgeConstants.MODERN_FORGE_TOKEN)
+            && handshake.getProtocolVersion().compareTo(ProtocolVersion.MINECRAFT_1_20_2) >= 0) {
+      String host = handshake.getServerAddress();
+      String[] split = host.split("\0", 2);
+      ModernForgeConstants.MODERN_FORGE_HOSTNAME_TOKEN = "\0" + split[1];
```
electronicboy: Now you're leaking connection state globally... should probably also mirror forges logic of reading the number out and storing that, then it can be reconstructed later

Mgazul: Corrected, hoping this is the expected way

IzzelAliz: You are still leaking the state. Just initialize a connection type instance every time. https://github.com/ArclightPowered/lightcity/blob/master/src/main/java/io/izzel/arclight/lightcity/ModernForgeConnectionType.java

Mgazul: forge does not process GameProfile, nor does it forward GameProfile in modern mode.

SimonMage: @IzzelAliz Hey , is it possible approving the PR without this thing or is it necessary for the future?

electronicboy: it's neccesary, otherwise it's just half baked support that will break as soon as forge touches anything with it

SimonMage: > it's neccesary, otherwise it's just half baked support that will break as soon as forge touches anything with it

Oh ok I understood

Mgazul: I've tried my best to fix it

SimonMage (approved):

SimonMage:
@electronicboy now the code is perfect right?

electronicboy:
I'm not 100% fond of the way this logic is handled, but that's moreso something that is seperate to this PR and should probably be cleaned up elsewhere

electronicboy (approved):

PR #1183, opened 2024-01-11
Update Kyori Bom to release version

CoryBorek:
When I was trying to use Kyori Adventure 4.15.0-SNAPSHOT, it couldn't be found by any maven search. Version 4.15 is also stable, as of about 3 weeks ago, so it should work.


4drian3d:
duplicate of https://github.com/PaperMC/Velocity/pull/1163

PR #1187, opened 2024-01-12
Fix player list header/footer not sending translated version

WonderfulPanic:


alexstaeding:
Don't open a second pull request for the same change in the future. You can update an existing pull request with additional commits or even force push to keep it in one commit.

WonderfulPanic:
sorry, didnt know you could update an existing request

PR #1193, opened 2024-01-17
Fully replaced AsyncHttpClient

4drian3d:
The AsyncHttpClient is not used anywhere else, so it is no longer necessary to maintain it


electronicboy:
aaaaaaaaaa

PR #1198, opened 2024-01-18
Update 1.20.5

4drian3d:
Current Supported Version: 1.20.5 Release Candidate 3


AoElite:
Maybe there could be a option to validate their authentication with cookies if they were transferred from another trusted server. Would reduce calls to mojang's API.

4drian3d:
> Maybe there could be a option to validate their authentication with cookies if they were transferred from another trusted server. Would reduce calls to mojang's API.

The cookies feature is only for mods environments, the server does not register any cookie by default and if it receives one, you will be immediately kicked out

proxy/src/main/java/com/velocitypowered/proxy/protocol/StateRegistry.java
```diff
@@ -419,11 +450,13 @@ public enum StateRegistry {
           map(0x3D, MINECRAFT_1_19_3, true),
           map(0x41, MINECRAFT_1_19_4, true),
           map(0x43, MINECRAFT_1_20_2, true),
-          map(0x45, MINECRAFT_1_20_3, true));
+          map(0x45, MINECRAFT_1_20_3, true),
+          map(0x46, MINECRAFT_1_20_5, true));
```
Luccboy: Shouldn't the respawn packet id be 0x47?

Luccboy (commented):
~The SignedChatCommand packet has been split into one that is for commands with signed arguments and one for commands without. This change is currently not implemented, right?~ (already done) Thanks for your great work!

proxy/src/main/java/com/velocitypowered/proxy/protocol/StateRegistry.java
Gerrygames: good catch!

Gerrygames:
The new unsigned command packet should be handled correctly now

proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/ServerLoginSuccessPacket.java
```diff
@@ -118,6 +121,9 @@ public void encode(ByteBuf buf, ProtocolUtils.Direction direction, ProtocolVersi
         ProtocolUtils.writeProperties(buf, properties);
       }
     }
+    if (version.noLessThan(ProtocolVersion.MINECRAFT_1_20_5)) {
+      buf.writeBoolean(false);
```
U5B: Might want to add a note that this probably will get removed in 1.21
> Changed the approach to game client disconnecting on invalid data in packets sent from a server
> To ease the transition period, modded servers can opt out by setting the appropriate field in the handshake/game_profile packet. This option will be removed in the next release

4drian3d: I don't think a comment about it is so necessary, but I added a system variable to disable that check in the client, by default the check will be enabled

proxy/src/main/java/com/velocitypowered/proxy/config/migration/TransferIntegrationMigration.java
```diff
@@ -0,0 +1,40 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package com.velocitypowered.proxy.config.migration;
+
```
LemonCaramel, line 19:
```suggestion

import com.electronwill.nightconfig.core.CommentedConfig;
```

proxy/src/main/java/com/velocitypowered/proxy/config/migration/TransferIntegrationMigration.java
```diff
@@ -0,0 +1,40 @@
...
+public final class TransferIntegrationMigration implements ConfigurationMigration {
+  @Override
+  public boolean shouldMigrate(final CommentedFileConfig config) {
+    return configVersion(config) < 2.7;
+  }
+
+  @Override
+  public void migrate(final CommentedFileConfig config, final Logger logger) {
+    config.set("accepts-transfers", false);
+    config.setComment("accepts-transfers", """
```
LemonCaramel:
```suggestion
    final CommentedConfig advanced = config.get("advanced");
    advanced.set("accept-transfers", false);
    advanced.setComment("accept-transfers", """
```

Wrongly migration

proxy/src/main/java/com/velocitypowered/proxy/config/migration/TransferIntegrationMigration.java
LemonCaramel, line 19:
```suggestion

import com.electronwill.nightconfig.core.CommentedConfig;
```

LemonCaramel (changes requested):

proxy/src/main/java/com/velocitypowered/proxy/config/migration/TransferIntegrationMigration.java
```diff
@@ -0,0 +1,40 @@
...
+public final class TransferIntegrationMigration implements ConfigurationMigration {
+  @Override
+  public boolean shouldMigrate(final CommentedFileConfig config) {
+    return configVersion(config) < 2.7;
+  }
+
+  @Override
+  public void migrate(final CommentedFileConfig config, final Logger logger) {
+    config.set("advanced.accepts-transfers", false);
+    config.setComment("advanced.accepts-transfers", """
```
LemonCaramel, line 35: Wrongly migration (2)
```suggestion
    config.set("advanced.accept-transfers", false);
    config.setComment("advanced.accept-transfers", """
```

kashike (approved):

PR #1203, opened 2024-01-18
Improve protocol version checking

kashike:


api/src/main/java/com/velocitypowered/api/util/Ordered.java
```diff
@@ -0,0 +1,93 @@
...
+ */
+package com.velocitypowered.api.util;
+
+import org.jspecify.annotations.NullMarked;
+
+/**
+ * Something that is ordered.
+ *
+ * @param <T> the type
+ * @since 1.0.0
```
powercasgamer:
```suggestion
 * @since 3.3.0
```
same for the other methods

powercasgamer (changes requested):

PR #1204, opened 2024-01-19
Fixed possible ConcurrentModificationException when iterating across TabList entries

### LINKED ISSUE #1002: VelocityTabList uses a non-thread safe map for storing player entries

4drian3d:
When returning the immutable list of TabList entries, Velocity unnecessarily used a stream to generate the list, which caused a ConcurrentModificationException in some cases where the tablist is modified where multiple streams may be running at the same time

resolves #1002 
closes #1003

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/tablist/VelocityTabList.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/tablist/VelocityTabList.java
@@ -38,7 +38,6 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.UUID;
-import java.util.stream.Collectors;
 import net.kyori.adventure.text.Component;
 import org.apache.logging.log4j.LogManager;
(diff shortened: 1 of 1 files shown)


electronicboy:
this still leaves a hole

electronicboy (approved):

PR #1206, opened 2024-01-20
Implement some missing resource pack API

pop4959:
Just taking a quick jab at this, looking to re-ignite some discussion around filling some gaps in the current API / implementation for resource packs.

~This does not attempt to implement adventure's new resource pack API currently, as nothing else in Velocity currently uses it, and it seemed strange to me to use it only for the new methods added here - should probably be addressed in a separate PR (possibly resulting in a quick deprecation along with a lot of the other resource pack API).~
We're going on an adventure.

The tracking for pending/applied resource packs gets a little weird since older packs don't have an ID, but it at least implements the current stub API method making it usable.

Also for `getAppliedResourcePack` and `getPendingResourcePack` I added a note regarding the return value as this is always going to be incorrect when used on players on 1.20.3+ since they can have several packs. The API will just return the last pack applied or pending. I think it's worth considering strengthening the deprecation to for removal as well.

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/proxy/Player.java
+++ b/api/src/main/java/com/velocitypowered/api/proxy/Player.java
@@ -231,22 +231,6 @@ public interface Player extends
   void sendResourcePackOffer(ResourcePackInfo packInfo);
 
   /**
-   * Requests that the player remove all applied resource packs. To
-   * monitor the status of any removed resource packs, subscribe to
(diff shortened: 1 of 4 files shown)


pop4959:
Adventure-ified with a suppression because one of the overridden methods is "too long" and checkstyle doesn't like it

kashike (approved):

electronicboy (approved):

PR #1211, opened 2024-01-21
Allow to register main class as listener after unregistering it

### LINKED ISSUE #780: EventManager#unregisterListeners unregisters the plugin main instance
```java
@Plugin(
    id = "testplugin"
)
public class TestPlugin {
  @Inject
  private ProxyServer server;

  @Subscribe
  public void onProxyInitialization(ProxyInitializeEvent event) {
    EventManager eventManager = this.server.getEventManager();
    eventManager.unregisterListeners(this);
    System.out.println("test123");
  }

  @Subscribe
  public void onProxyReload(ProxyReloadEvent event) {
    System.out.println("test321");
  }
}
```

Excepted behaviour:
- It writes ``test123`` on startup
- It writes ``test321`` after ``/velocity reload``

Current behaviour:
- It writes ``test123`` on startup
- It doesn't write ``test321`` after ``/velocity reload``

We can't register the main instance again (``eventManager.registerListeners(this, this)``) as the "The plugin main instance is automatically registered." error appears

4drian3d:
resolves #780 

_Needs testing._


electronicboy:
Would it maybe make more sense to just turn this into a "is the current instance already registered" type thing and then get rid of the specific "main class" check?

PR #1215, opened 2024-01-21
Fix "Add -haproxy-protocol startup parameter" 

BingYanchi:
Fixes https://github.com/PaperMC/Velocity/pull/1136

Wrong use of .defaultsTo(null)


powercasgamer (changes requested):
This only adds it to the options, doesn't override or set it in VelocityServer

BingYanchi:
> This only adds it to the options, doesn't override or set it in VelocityServer

I compiled and ran it and it seems to work fine.

PR #1216, opened 2024-01-21
Revert "Revert "Add -haproxy-protocol startup parameter" (#1214)"

powercasgamer:
This reverts commit b23c2c6e1a21815683d4ad2f0a1ed054f1cad890 which reverts commit https://github.com/PaperMC/Velocity/commit/363e3a7f94b1a8693893f4fe0349fb181b127bbb.

Tested (the `true` message was a debug message I added during testing)
WindowsTerminal_td7u0bVe8B

PR #1217, opened 2024-01-21
Fix "Add -haproxy-protocol startup parameter"

BingYanchi:
Fixes https://github.com/PaperMC/Velocity/pull/1136


4drian3d:
duplicate of #1216

PR #1224, opened 2024-01-29
Implement ProtocolState API

4drian3d:


api/src/main/java/com/velocitypowered/api/network/ProtocolState.java
```diff
@@ -0,0 +1,46 @@
...
+   * Authentication state of a connection, at this moment the player is authenticating
+   * with the authentication servers (Mojang).
+   */
+  LOGIN,
+  /**
+   * Configuration State, at this point the player allows the server to send information
+   * such as resource packs and plugin messages, at the same time the player
+   * will send his client brand and the respective plugin messages if it is a modded client.
+   *
+   * @since Minecraft 1.20.5
```
Luccboy: Doesn't the configuration phase exist since 1.20.2?

4drian3d: Yes, at that time I was thinking about the new HandshakeIntent in 1.20.5, thanks

api/src/main/java/com/velocitypowered/api/network/ProtocolState.java
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Representation of the state of the protocol
+ * in which a connection can be present.
+ */
+public enum ProtocolState {
+  /**
+   * Initial connection state.
+   * <p>This status can be caused by a STATUS, LOGIN or TRANSFER intent.</p>
+   * If the intent is LOGIN or TRANSFER, the connection will proceed, otherwise,
```
electronicboy: all connections will proceed, so the working of this makes little sense

api/src/main/java/com/velocitypowered/api/network/ProtocolState.java
```diff
@@ -0,0 +1,46 @@
...
+ * reference the LICENSE file in the api top-level directory.
+ */
+
+package com.velocitypowered.api.network;
+
+/**
+ * Representation of the state of the protocol
+ * in which a connection can be present.
+ */
+public enum ProtocolState {
```
electronicboy, line 16: I'm not exactly fond of the documentation on this class, somebody with better tonguetangualation would probably be able to advise better than I can

electronicboy (approved):
Overall this looks fine, I'm just not 100% sure on the nature of the documentation in the API class. It's worded somewhat awkwardly in some areas and tends to describe aspects which I fee is a mixture of potentially being overdocumented but also being potentially confusing, i.e. PLAY represents the protocol state in which gameplay packets exist, but, being in that packet state does not mean that a player is actually in game yet

proxy/src/main/java/com/velocitypowered/proxy/protocol/StateRegistry.java
```diff
@@ -623,6 +624,21 @@ public StateRegistry.PacketRegistry.ProtocolRegistry getProtocolRegistry(Directi
     return (direction == SERVERBOUND ? serverbound : clientbound).getProtocolRegistry(version);
   }
 
+  /**
+   * Gets the API representation of the ProtocolState.
```
Timongcraft: Isn't it the "representation of the StateRegistry."?

api/src/main/java/com/velocitypowered/api/proxy/InboundConnection.java
```diff
@@ -43,4 +44,11 @@ public interface InboundConnection {
    * @return the protocol version the connection uses
    */
   ProtocolVersion getProtocolVersion();
+
+  /**
+   * Returns the current protocol state of this connection.
+   *
+   * @return the protocol state of the connection
+   */
+  ProtocolState protocolState();
```
Joo200: I don't like to mix record style getters and the other getters in the same class. Either change the style of all getters and deprecate the old ones or change this to getProtocolState();

4drian3d: While I would like to change all the methods to Adventure/records style, I can't do it just before the release, I will leave it as a regular getter, until the next major release

api/src/main/java/com/velocitypowered/api/event/connection/PreLoginEvent.java
```diff
@@ -75,6 +75,7 @@ public String getUsername() {
    * Returns the UUID of the connecting player. This value is {@code null} on 1.19.1 and lower.
    *
    * @return the uuid
+   * @sinceMinecraft 1.19.1
```
Timongcraft: Is this right? Because if it is null in 1.19.1 and below, why was it added in 1.19.1?

4drian3d: fixed

PR #1225, opened 2024-01-31
Refactor ResourcePacks handling

### LINKED ISSUE #1178: New Resource Pack API in 1.20.4
In 1.20.4, clients allow saving multiple server resource bundles, and multiple resource bundles can be applied simultaneously.

But I don't seem to see the relevant JavaDoc. Velocity's original sending resource package api will not clear the original resource package. After obtaining the new resource package, it will coexist with the old resource package.

It may be that Velocity does not support the new resource pack of 1.20.4.

4drian3d:


proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
```diff
@@ -981,6 +971,18 @@ public void spoofChatInput(String input) {
     }
   }
 
+  /**
+   * Get the ResourcePackHandler corresponding to the player's version.
+   *
+   * @return the ResourcePackHandler of this player
+   */
+  public ResourcePackHandler resourcePackHandler() {
```
electronicboy, line 986: What is the risk that this is called across multiple threads? Do we need to guard this?

proxy/src/main/java/com/velocitypowered/proxy/connection/player/resourcepack/ResourcePackHandler.java
```diff
@@ -0,0 +1,158 @@
...
+    if (outstandingResourcePacks.size() == 1) {
+      tickResourcePackQueue();
+    }
+  }
+
+  /**
+   * Queues a resource-request for sending to the player and sends it immediately if the queue is
+   * empty.
+   */
+  public void queueResourcePack(ResourcePackRequest request) {
```
electronicboy: Need to note to remember to handle the reset from ResourcePackRequest

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
4drian3d: I doubt that in this case anything will happen, but of course, I will make the variable volatile.
I did the implementation this way to avoid creating the respective implementation if the server does not use resource packs

electronicboy: volatile won't fix the concern if we do expect it to be called from various threads, which if it was just purely passthrough wouldn't be much of a deal as it would happen purely on the event loop, if a plugin calls it, low risk, but there is a chance we could double init

proxy/src/main/java/com/velocitypowered/proxy/connection/player/resourcepack/ModernResourcePackHandler.java
```diff
@@ -95,6 +97,20 @@ public void queueResourcePack(final @NotNull ResourcePackInfo info) {
     }
   }
 
+  @Override
+  public void queueResourcePack(@NotNull ResourcePackRequest request) {
+    if (request.packs().size() > 1) {
+      player.getConnection().write(new BundleDelimiterPacket());
```
kashike: Have a single `INSTANCE` field in `BundleDelimiterPacket` instead of creating it each time?

proxy/src/main/java/com/velocitypowered/proxy/connection/player/resourcepack/LegacyResourcePackHandler.java
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  @Override
+  public void clearAppliedResourcePacks() {
+    this.appliedResourcePack = null;
+  }
+
+  @Override
+  public boolean remove(final @NotNull UUID id) {
+    if (appliedResourcePack != null && id.equals(appliedResourcePack.getId())) {
```
electronicboy: Legacy doesn't support removing a packet from the client, and so I don't think that this is a good idea.

I think that there was some questions over if Velocity wanted to provide some hosted empty packs for the sake of offering them as an API thing, no idea if there was anything against that in the org, but, maybe worth considering if we did want to try to "support" this

electronicboy:
also TODO: Check if we're handing configuration phase properly, right now we end up writing stuff to the wrong connection on server switches, that's why switching between servers with resource packs seems to break

electronicboy (approved):

kashike (approved):
Approving code only - I have not tested.

Gerrygames:
Works on production network

PR #1232, opened 2024-02-07
Add API for custom chat completions like on Paper

itsTyrion:
I added 3 API methods to the Player class like they are present on Paper. It compiled successfully and appears to work: 
image

The reason for changing the single enumeration is that it's name, `ALTER`, does not seem to match it's function or the name used by Mojang/Yarn/Searge mappings, which call it `SET`.
The change shouldn't be breaking, as the enum class isn't part of the public API and not directly called in the internal API

Please let me know if something is wrong/not to your likings.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
@@ -967,22 +967,27 @@ public class ConnectedPlayer implements MinecraftConnectionAssociation, Player, 
 
   @Override
   public void addCustomChatCompletions(@NotNull Collection<String> completions) {
+    Preconditions.checkNotNull(completions, "completions");
     this.sendCustomChatCompletionPacket(completions, PlayerChatCompletionPacket.Action.ADD);
   }
 
   @Override
   public void removeCustomChatCompletions(@NotNull Collection<String> completions) {
+    Preconditions.checkNotNull(completions, "completions");
     this.sendCustomChatCompletionPacket(completions, PlayerChatCompletionPacket.Action.REMOVE);
   }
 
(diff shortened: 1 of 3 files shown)


proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
```diff
@@ -964,6 +965,26 @@ void setClientBrand(String clientBrand) {
...
+
+  @Override
+  public void setCustomChatCompletions(@NotNull Collection<String> completions) {
+    this.sendCustomChatCompletionPacket(completions, PlayerChatCompletionPacket.Action.SET);
+  }
+
+  private void sendCustomChatCompletionPacket(@NotNull Collection<String> completions,
+                                              PlayerChatCompletionPacket.Action action) {
+    connection.write(new PlayerChatCompletionPacket(completions.toArray(new String[0]), action));
+  }
```
4drian3d: You should check if the player has version 1.19 or higher since in previous versions this feature did not exist

4drian3d: Also, you must add null checks

4drian3d (changes requested):

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
itsTyrion: Oh right, I'm spoiled from writing mostly Kotlin these days :D 

What's the course of action, just drop if the player is below 1.19?

4drian3d (approved):

electronicboy (approved):

PR #1250, opened 2024-02-22
Add javadocs warnings about unsupported operations

### LINKED ISSUE #634: On the tendency to err by calling no-op methods on Audience
It is a very common mistake to accidentally call methods inherited from Audience which are not implemented in any meaningful way on Velocity.

Consider, as evidence, code which appears well and normal to one familiar with Adventure but not closely acquainted with Velocity:
> Space_Interprise:
> player.playSound(Sound.sound(Key.key("minecraft:entity.enderdragon.death"), Sound.Source.MASTER, 1.0f, 1.0f), Sound.Emitter.self());
> Any clue why this doesn't work?

While this is a deliberate design decision on the behalf of Adventure to default to "no-op" implementations, it inevitably creates room for error and surprise. Regardless of the motivation behind this decision, it cannot be reversed without violating contract, which is sacrosanct.

Velocity is in a position to mitigate this issue. By leveraging that such mistakes will presumably occur on types from the Velocity API, methods can be overridden to document that, currently, they have no effect. This being done, API users may see the documentation and then realize the mistake.

The problem will remain to some extent, unavoidably; documentation will prevent only some instances. It might be tempting to deprecate these overriding methods; I do not suggest that and would advocate against it.

### LINKED ISSUE #1077: Document additional whitespace in CommandExecuteEvent
> [3:58 PM]Five: Also here's a fair warning
> [3:58 PM]Five: If you use the commandexecuteevent you may want to strip preceding whitespace
> [3:58 PM]Five: If it's for denying commands

If there's a fair warning about anything, it should be documented in the API.

4drian3d:
resolves #634 
resolves #1077

### DIFF
--- a/api/build.gradle.kts
+++ b/api/build.gradle.kts
@@ -70,6 +70,12 @@ tasks {
             "https://javadoc.io/doc/com.github.ben-manes.caffeine/caffeine"
         )
 
+        o.tags(
+            "apiNote:a:API Note:",
+            "implSpec:a:Implementation Requirements:",
+            "implNote:a:Implementation Note:"
+        )
+
         // Disable the crazy super-strict doclint tool in Java 8
         o.addStringOption("Xdoclint:none", "-quiet")
     }
--- a/api/src/main/java/com/velocitypowered/api/event/command/CommandExecuteEvent.java
+++ b/api/src/main/java/com/velocitypowered/api/event/command/CommandExecuteEvent.java
@@ -39,6 +39,11 @@ public final class CommandExecuteEvent implements ResultedEvent<CommandResult> {
     this.result = CommandResult.allowed();
   }
 
+  /**
+   * Gets the source responsible for the execution of this command.
+   *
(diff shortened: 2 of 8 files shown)


electronicboy:
Warning like this afaik breaks the general contract that adventure imposes, best we could probably do is override these methods and add some jd or annotation or something

4drian3d:
> Warning like this afaik breaks the general contract that adventure imposes, best we could probably do is override these methods and add some jd or annotation or something

As I mentioned in Discord, it would not be throwing an error when detecting that a plugin is executing these methods, but it is showing a message in console, which allows any plugin to execute these methods without problems. But in a while I will add additional documentation

electronicboy:
I said warning, but there is effecitively nothing that distinguishes between the two for end users, adventures general expectation towards its apis and consumers is that such methods are a no-op. As said, best we can probably do is override them with some annotation to provide some form of ide hint or whatever

PR #1256, opened 2024-02-25
Add CommandSource#sendRichMessage and #sendPlainMessage

4drian3d:
As with Paper, this would greatly simplify introducing more developers to the Components system in Minecraft and reduce a lot of boilerplate in operations that constantly send messages created in MiniMessage format to some CommandSource


VinesMC:
XD bro

PR #1265, opened 2024-03-05
Add Virtualhost support for server list pings

scratchyone:
When a player connects to a server through velocity, velocity correctly forwards the virtualhost used to connect. This allows velocity to be run in front of proxy/routing services such as [mc-router](https://github.com/itzg/mc-router), and gives the proxying services running behind velocity the information they need to know which final game-server to forward the request to.

However, velocity doesn't forward that information for server list pings, meaning that even with ping passthrough enabled, any proxy/router running behind velocity won't have the information it needs to be able route pings to the correct game-server based on the virtual host. Instead, velocity would just create the ping packet with the host set to whatever was configured in `velocity.toml`. This PR fixes that issue, and makes it so virtualhosts are now forwarded with server list pings as well.

I'm not super familiar with the velocity codebase as a whole, so if this PR needs any changes to align with the overall goals or structure of the project, I'd be happy to make any requested modifications.


electronicboy (approved):

4drian3d (approved):

scratchyone:
I added an additional commit to add support for this to the public API so plugins can use it.

scratchyone:
Although, the public API change I've proposed might not be the best way to go about things. The expected behavior should pretty much always be to use a virtualhost if the player connected with one, but with this method plugin devs have to go out of their way to support it. Shouldn't the easiest way to do things be the correct way? Maybe `player.getServer()` (and similar functions) should return a new class (maybe `ConnectedServer`?) that implements `RegisteredServer` and stores virtualhost information for pings and connections and stuff. I'm curious to hear people's thoughts on this change?

api/src/main/java/com/velocitypowered/api/proxy/server/RegisteredServer.java
```diff
@@ -50,4 +50,14 @@ public interface RegisteredServer extends ChannelMessageSink, Audience {
...
+
+  /**
+   * Attempts to ping the remote server with the given virtualhost and return the server list ping
+   * result according to the options provided.
+   *
+   * @param pingOptions the options provided for pinging the server
+   * @param virtualHostString the virtual hostname to pass to the server for the ping
+   * @return the server ping result from the server
+   */
+  CompletableFuture<ServerPing> ping(PingOptions pingOptions, String virtualHostString);
```
4drian3d: Instead of adding another ping method, add a virtual host field to PingOptions

scratchyone: Sounds good! Should I make this change for the private internal API as well?

4drian3d (approved):

PR #1271, opened 2024-03-16
Kikatsuru

Kikatsuru:
SORRY my fault


4drian3d:
I know it's a pull request made by mistake, but all this could be simplified in one single line

PR #1277, opened 2024-03-29
Use the default Minecraft port instead of custom one

Phoenix616:
Currently Velocity uses the custom port `25577` in its default config which does not make any sense for 99.9% of deployments and has caused confusion and unnecessary support requests time and time again without any obvious benefit as usually you want the proxy to be reachable instead of your normal servers which wouldn't even be usable anymore as soon as you enable proxy support on the Minecraft server. (I kinda assume this was originally just copied from Bungee's default being 25577)

While a non-standard port can be useful for testing in some situations when already having a Minecraft server running in others it might actually be decremental (e.g. because you need to add another server into your server list instead of just using one localhost with port 25565). For cases where this behaviour is desired a custom port can always be specified in the config anyways.

Existing setups should not get influenced by this change as they already have a config with the value and even in a case where a container image is used which relies on a specific port this change should not cause any issues as the config should either be persisted or be part of the container image if it is setup right. (e.g. like how the [Velocity Pterodactyl egg](https://github.com/parkervcp/eggs/blob/54209dbe130b1eb00d8cf3702e64f2146c138103/game_eggs/minecraft/proxy/java/velocity/velocity.toml#L5) does it)

_And if there was discussion/issues/PRs open for that before then I'm sorry but I couldn't find them..._

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
@@ -59,7 +59,7 @@ public class VelocityConfiguration implements ProxyConfig {
   private static final Logger logger = LogManager.getLogger(VelocityConfiguration.class);
 
   @Expose
-  private String bind = "0.0.0.0:25577";
+  private String bind = "0.0.0.0:25565";
   @Expose
   private String motd = "<aqua>A Velocity Server";
   @Expose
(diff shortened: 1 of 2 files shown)


rafi67000:
bungee does the same so idk they probably wanted to keep it the same as bungee.

electronicboy:
I generally hit my config over to the default port locally, the only reason for being a non-default port that I can see for quick testing, but, you generally need to configure a bunch of other stuff anyways

I'm generally for merging this, but, need to do it in the next version, not here

kashike (approved):

pop4959 (approved):

4drian3d (approved):

NoahvdAa (approved):

RealBauHD (approved):

PR #1302, opened 2024-04-24
Correctly handle "select known packs" packet

Binero:
This is required for connecting to 1.20.5 servers to succeed. Implementation was tested using https://github.com/OKTW-Network/FabricProxy-Lite/pull/84.


kashike:
What exactly is this required for?

Binero:
@kashike The contents of the registry packet depend on whatever the client sends in its `KnownPacks` packet. Velocity right now doesn't forward this packet towards the server. This means the Minecraft server will perpetually wait for the client. This eventually causes a timeout.

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientConfigSessionHandler.java
```diff
@@ -131,6 +132,15 @@ public boolean handle(PingIdentifyPacket packet) {
     return true;
   }
 
+  @Override
+  public boolean handle(KnownPacksPacket packet) {
+    if (player.getConnectionInFlight() != null) {
+      player.getConnectionInFlight().ensureConnected().write(packet);
```
Binero, line 138: This is the important bit of the PR, this ensures that the client's response makes it to the server. (By default Velocity drops unknown packets during the configuration state.)

electronicboy (approved):

PR #1303, opened 2024-04-24
Limit maximum amount of known packs to vanilla defaults

UserNugget:
## Problem
Currently, "attacker" can spam gigabytes of arrays and so easly crash Velocity

## Modification
Limit maximum amount of KnownPacks to vanilla defaults, but with ability to change that limit via system properly.

## Result
Velocity will never fully decode corrupted/malicious KnownPacksPacket


electronicboy (approved):

electronicboy:
Thanks

PR #1309, opened 2024-04-27
Assign ServerConnection to Player before processing post-login packets

wallenjos01:
Currently, when transitioning to a new server, the proxy begins auto-reading (and therefore processing) packets from the pending server connection before assigning that connection to the player via `ConnectedPlayer::setConnectedServer`. This causes a race condition if the backend server sends certain packets (particularly plugin messages) right after sending a `JoinGamePacket`. For example, the following has inconsistent behavior depending on when a plugin message is sent:
```java
@Subscribe
public void onPluginMessage(PluginMessageEvent event) {
    if(event.getSource() instanceof ServerConnection sc) {

        // Will work consistently
        sc.sendPluginMessage(MinecraftChannelIdentifier.create("test", "message"), "Hello, World".getBytes()); 
        
        // Will often fail if the backend server sends a plugin message immediately after join.
        sc.getPlayer()
            .getCurrentServer()
            .orElseThrow()
            .sendPluginMessage(MinecraftChannelIdentifier.create("test", "message"), "Hello, World".getBytes());
    }
}
```
This PR reverses the order of two lines in `TransitionSessionHandler`, making the proxy wait to process packets sent by a backend server until after assigning the player's connection.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/backend/TransitionSessionHandler.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/backend/TransitionSessionHandler.java
@@ -137,7 +137,7 @@ public class TransitionSessionHandler implements MinecraftSessionHandler {
           smc.setActiveSessionHandler(StateRegistry.PLAY,
               new BackendPlaySessionHandler(server, serverConn));
 
-          // Now set the connected server. 
+          // Now set the connected server.
           serverConn.getPlayer().setConnectedServer(serverConn);
 
           // Clean up disabling auto-read while the connected event was being processed.
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/backend/TransitionSessionHandler.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/backend/TransitionSessionHandler.java
@@ -137,10 +137,12 @@ public class TransitionSessionHandler implements MinecraftSessionHandler {
           smc.setActiveSessionHandler(StateRegistry.PLAY,
               new BackendPlaySessionHandler(server, serverConn));
 
-          // Now set the connected server.
+          // Now set the connected server. 
           serverConn.getPlayer().setConnectedServer(serverConn);
 
           // Clean up disabling auto-read while the connected event was being processed.
+          // Do this after setting the connection, so no incoming packets are processed before
+          // the API knows which server the player is connected to.
           smc.setAutoReading(true);
 
           // Send client settings. In 1.20.2+ this is done in the config state.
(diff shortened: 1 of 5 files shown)


proxy/src/main/java/com/velocitypowered/proxy/connection/backend/TransitionSessionHandler.java
```diff
@@ -137,12 +137,12 @@ public boolean handle(JoinGamePacket packet) {
...
               new BackendPlaySessionHandler(server, serverConn));
 
-          // Clean up disabling auto-read while the connected event was being processed.
-          smc.setAutoReading(true);
-
           // Now set the connected server.
           serverConn.getPlayer().setConnectedServer(serverConn);
 
+          // Clean up disabling auto-read while the connected event was being processed.
+          smc.setAutoReading(true);
```
kashike, line 146: Would you mind expanding the comment above this line to include your description (or similar)?

Gerrygames:
Pretty sure that this does not make a difference. The code in question runs on the event loop of the server connection, so no packets from that connection can be handled during its execution.

electronicboy:
Has this been tested? Because I'd imagine that the more bigger issue here is going to be that disabling auto-read doesn't promise to disable reading *everything*, and so there can still be stuff which ends up being processed, at least that is my understanding; The auto-reading re-enable + field update both occur within the same area.

wallenjos01:
> Has this been tested? 

I created this PR because I ran into errors caused by this bug while working on a plugin for my private server, but just to be sure, I spent some time writing a plugin to test it. I tested with both Velocity build 385, downloaded from the website, and my own fork. I tested both Spigot 1.20.5 and Fabric 1.20.5 backends. I got about a 40% error rate when using build 385, and a 0% error rate using my fork. I have attached the relevant logs and code.

[pmtest.zip](https://github.com/PaperMC/Velocity/files/15142692/pmtest.zip)

proxy/src/main/java/com/velocitypowered/proxy/connection/backend/TransitionSessionHandler.java
```diff
@@ -137,12 +137,14 @@ public boolean handle(JoinGamePacket packet) {
           smc.setActiveSessionHandler(StateRegistry.PLAY,
               new BackendPlaySessionHandler(server, serverConn));
 
+          // Now set the connected server.
```
R00tB33rMan: Trim whitespace after the period (line 140)

electronicboy (approved):

kennytv (approved):
From a quick look, this seems to be caused by the manual channel.read call in MinecraftConnection#setAutoReading, so the PR should be good

Issue #1251, opened 2024-02-23
Unexpectedly disconnected from remote server


Teslicek:
Sometimes, when a player has a weird connection, they are kicked to the fallback server without properly joining it, which causes an error on proxy, as well as some random message on the backend server.

Upon asking, I was told that this error is caused by the client being stuck in the configuration phase, because the proxy did not receive any packets from the client.

Happened on Velocity 359 with no resource pack.

Backend message:
```
[...]

PR #1316, opened 2024-05-05
Fix config phase packets getting lost

valaphee:
When the config-phase is initiated externally, packets are getting lost, as there is no `connectionInFlight`.

This is most noticeable with `PluginMessagePacket` as some client-only mods initiate their state in the config phase. The servers request is receipt, but the client has no way to acknowledge the pm.

One common scenario where it would occur is when you chain multiple Velocity servers. But it should be the expected behavior nonetheless.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientConfigSessionHandler.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientConfigSessionHandler.java
@@ -120,6 +120,8 @@ public class ClientConfigSessionHandler implements MinecraftSessionHandler {
       // but at this time the backend server may not be ready
     } else if (serverConn != null) {
       serverConn.ensureConnected().write(packet.retain());
+    } else {
+      player.getConnection().write(packet.retain());
     }
     return true;
   }
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientConfigSessionHandler.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientConfigSessionHandler.java
@@ -128,18 +128,23 @@ public class ClientConfigSessionHandler implements MinecraftSessionHandler {
 
   @Override
   public boolean handle(PingIdentifyPacket packet) {
-    if (player.getConnectionInFlight() != null) {
-      player.getConnectionInFlight().ensureConnected().write(packet);
+    final VelocityServerConnection serverConn = player.getConnectionInFlight();
+    if (serverConn != null) {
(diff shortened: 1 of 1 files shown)


R00tB33rMan:
> When the config-phase is initiated externally, packets are getting lost, as there is no `connectionInFlight`.
> 
> This is most noticeable with `PluginMessagePacket` as some client-only mods initiate their state in the config phase. The servers request is receipt, but the client has no way to acknowledge the pm.
> 
> One common scenario where it would occur is when you chain multiple Velocity servers. But it should be the expected behavior nonetheless.

Does this theoretically resolve #1251?

valaphee:
> > When the config-phase is initiated externally, packets are getting lost, as there is no `connectionInFlight`.
> > This is most noticeable with `PluginMessagePacket` as some client-only mods initiate their state in the config phase. The servers request is receipt, but the client has no way to acknowledge the pm.
> > One common scenario where it would occur is when you chain multiple Velocity servers. But it should be the expected behavior nonetheless.
> 
> Does this theoretically resolve #1251?

Shouldn't impact this issue. At first this PR only impacts client -> server packets, and only normally unrelevant packets. (at least when you don't use modded Minecraft)

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientConfigSessionHandler.java
```diff
@@ -120,24 +120,31 @@ public boolean handle(final PluginMessagePacket packet) {
...
 
   @Override
   public boolean handle(PingIdentifyPacket packet) {
-    if (player.getConnectionInFlight() != null) {
-      player.getConnectionInFlight().ensureConnected().write(packet);
+    final VelocityServerConnection serverConn = player.getConnectionInFlight();
+    if (serverConn != null) {
+      serverConn.ensureConnected().write(packet);
+    } else {
+      player.getConnection().write(packet);
```
electronicboy, line 135: isn't this sending data to the wrong thing? you either want to get the connection in flight, or the connected server?

electronicboy:
Closed in favour of #1326 as this is sending packets in the wrong direction

valaphee:
connection is the clientbound connection, and this is the proxy handler acting as client? But the other PR is probably a better solution, I'll try.

But this definitely works, as we use a client mod and the handshake worked correctly afterwards.

electronicboy:
ClientConfigSessionHandler is receiving packets from the client, hence why it has logic to write to the connection in-flight, so I'm honestly not sure how this would fix anything *scratches head*

PR #1317, opened 2024-05-06
Allow expired or invalid key if key is not forced

VeVeVeVel:
I'm using the `force-key-authentication = false` configuration, but players are sometimes encountering 'invalid_public_key' issues.
If key is expired or invalid, do not kick the player but set the key to null.


electronicboy:
I'm not generally sure that it's in our interest to maintain clients partaking in such broken behaviour, we have no idea of the side effects of this, long or short term, and cannot generally vouch that in the long run, this will not further issues, etc.

PR #1338, opened 2024-05-30
MC update 1.21

Gerrygames:
TODO:
- [ ] API for custom report details (adventure?)
- [ ] API for server links (adventure?)


electronicboy:
We should use the mojang names for packets going forward

kennytv (approved):

PR #1341, opened 2024-06-06
Refactor: Replace enum comparison with .equals method

RisDN:
- Updated the comparison method for PlayerInfoForwarding mode to use .equals instead of '=='.
- Ensured consistent and reliable comparison for PlayerInfoForwarding and PlayerDataForwarding.


RealBauHD:
Since it is a normal enum, I see no reason to make a whole PR for it

4drian3d:
why?

PR #1353, opened 2024-06-13
feat: Add basic API for server links

WiIIiam278:
Adds a (very simple) API for setting a list of server links to display to the client with either built-in or component labels. Strings for URLs as is the case in the resource pack API.

I was thinking about doing something more complex with a builder or something, but I'm not sure it's really needed in this case?

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/proxy/Player.java
+++ b/api/src/main/java/com/velocitypowered/api/proxy/Player.java
@@ -21,6 +21,7 @@ import com.velocitypowered.api.proxy.player.TabList;
 import com.velocitypowered.api.proxy.server.RegisteredServer;
 import com.velocitypowered.api.util.GameProfile;
 import com.velocitypowered.api.util.ModInfo;
+import com.velocitypowered.api.util.ServerLink;
 import java.net.InetSocketAddress;
 import java.util.Collection;
 import java.util.List;
@@ -461,4 +462,14 @@ public interface Player extends
    * @sinceMinecraft 1.20.5
    */
   void requestCookie(Key key);
+
+  /**
+   * Send the player a list of custom links to display in their client's pause menu.
+   *
+   * @param links an ordered list of {@link ServerLink}s to send to the player
+   * @throws IllegalArgumentException if the player is from a version lower than 1.21
+   * @since 3.3.0
+   * @sinceMinecraft 1.21
+   */
+  void setServerLinks(List<ServerLink> links);
 }
--- a/api/src/main/java/com/velocitypowered/api/proxy/Player.java
+++ b/api/src/main/java/com/velocitypowered/api/proxy/Player.java
@@ -465,6 +465,7 @@ public interface Player extends
 
   /**
    * Send the player a list of custom links to display in their client's pause menu.
+   * <p>Note that later packets sent by the backend server may override links sent by the proxy.
    *
    * @param links an ordered list of {@link ServerLink}s to send to the player
    * @throws IllegalArgumentException if the player is from a version lower than 1.21
--- a/api/src/main/java/com/velocitypowered/api/proxy/Player.java
+++ b/api/src/main/java/com/velocitypowered/api/proxy/Player.java
@@ -471,5 +471,5 @@ public interface Player extends
    * @since 3.3.0
    * @sinceMinecraft 1.21
    */
-  void setServerLinks(List<ServerLink> links);
+  void setServerLinks(@NotNull List<ServerLink> links);
 }
--- a/api/src/main/java/com/velocitypowered/api/proxy/Player.java
+++ b/api/src/main/java/com/velocitypowered/api/proxy/Player.java
@@ -465,6 +465,7 @@ public interface Player extends
 
   /**
    * Send the player a list of custom links to display in their client's pause menu.
+   *
    * <p>Note that later packets sent by the backend server may override links sent by the proxy.
    *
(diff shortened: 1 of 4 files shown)


api/src/main/java/com/velocitypowered/api/proxy/Player.java
```diff
@@ -461,4 +462,14 @@ default void openBook(@NotNull Book book) {
    * @sinceMinecraft 1.20.5
    */
   void requestCookie(Key key);
+
+  /**
+   * Send the player a list of custom links to display in their client's pause menu.
```
RealBauHD, line 467: There should be a note that it can conflict with backend servers

api/src/main/java/com/velocitypowered/api/util/ServerLink.java
```diff
@@ -0,0 +1,93 @@
...
+  private @Nullable Component label;
+  private final String link;
+
+  /**
+   * Construct a server link with a custom component label.
+   *
+   * @param label a custom component label to display
+   * @param link  the URL to open when clicked
+   */
+  public ServerLink(Component label, String link) {
```
RealBauHD: A static method instead of the constructor like adventure does e. g. `ServerLink#serverLink`

api/src/main/java/com/velocitypowered/api/util/ServerLink.java
```diff
@@ -0,0 +1,93 @@
...
+    this.link = URI.create(link).toString();
+  }
+
+  /**
+   * Construct a server link with a built-in type.
+   *
+   * @param type the {@link Type built-in type} of link
+   * @param link the URL to open when clicked
+   */
+  public ServerLink(Type type, String link) {
```
RealBauHD: same

api/src/main/java/com/velocitypowered/api/util/ServerLink.java
```diff
@@ -0,0 +1,93 @@
...
+
+  /**
+   * Construct a server link with a custom component label.
+   *
+   * @param label a custom component label to display
+   * @param link  the URL to open when clicked
+   */
+  public ServerLink(Component label, String link) {
+    this.label = Preconditions.checkNotNull(label, "label");
+    this.link = URI.create(link).toString();
```
RealBauHD: Would it be convenient if we kept the URI?

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
```diff
@@ -1057,6 +1060,25 @@ public void requestCookie(final Key key) {
...
+    Preconditions.checkArgument(
+        this.getProtocolVersion().noLessThan(ProtocolVersion.MINECRAFT_1_21),
+        "Player version must be at least 1.21 to be able to set server links");
+
+    if (connection.getState() != StateRegistry.PLAY
+        && connection.getState() != StateRegistry.CONFIG) {
+      throw new IllegalStateException("Can only send server links in CONFIGURATION or PLAY protocol");
+    }
+
+    connection.write(new ClientboundServerLinksPacket(ImmutableList.copyOf(links).stream()
```
RealBauHD: The stream is a bit clunky

RealBauHD (changes requested):

api/src/main/java/com/velocitypowered/api/util/ServerLink.java
WiIIiam278: Sure, can do

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
WiIIiam278: I reduced the clunkiness a bit by making a constructor in the packet ServerLink class. Could make this an array + loop if needed, or something else you had in mind.

api/src/main/java/com/velocitypowered/api/util/ServerLink.java
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  /**
+   * Construct a server link with a custom component label.
+   *
+   * @param label a custom component label to display
+   * @param link  the URL to open when clicked
+   */
+  public static ServerLink serverLink(Component label, String link) {
+    return new ServerLink(label, link);
```
alexdev03, line 43: Add a check to the link

api/src/main/java/com/velocitypowered/api/util/ServerLink.java
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  /**
+   * Construct a server link with a built-in type.
+   *
+   * @param type the {@link Type built-in type} of link
+   * @param link the URL to open when clicked
+   */
+  public static ServerLink serverLink(Type type, String link) {
+    return new ServerLink(type, link);
```
alexdev03, line 53: Also here

WiIIiam278:
@alexdev03 The link is checked in the constructor above the static methods - hence is not needed twice.

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
novitpw: Use List#copyOf

4drian3d (approved):

powercasgamer (approved):

PR #1359, opened 2024-06-17
Resolve Cookie API clientbound error

R00tB33rMan:
Since 07f1f9e, errors such as listed below appeared when utilizing anti-VPN plugins (most notably): https://github.com/v4Guard/connector

This effectively eliminates this issue and ensures that everything through the connection phase is handled correctly.
image

Steps for reproduction:
1. Install latest build of Velocity.
2. Install v4Guard connector.
3. Enable a VPN (Cloudflare WARP works).
4. Join server and see this particular error appear.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/MinecraftConnection.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/MinecraftConnection.java
@@ -33,6 +33,7 @@ import com.velocitypowered.natives.encryption.VelocityCipher;
 import com.velocitypowered.natives.encryption.VelocityCipherFactory;
 import com.velocitypowered.natives.util.Natives;
 import com.velocitypowered.proxy.VelocityServer;
+import com.velocitypowered.proxy.connection.client.ConnectedPlayer;
 import com.velocitypowered.proxy.connection.client.HandshakeSessionHandler;
 import com.velocitypowered.proxy.connection.client.InitialLoginSessionHandler;
 import com.velocitypowered.proxy.connection.client.StatusSessionHandler;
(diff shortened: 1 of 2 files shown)


Luccboy:
If I undestand your changes correctly, you queue player disconnects triggered by Velocity that would happen in login phase and execute them later, when the protocol is switched to play phase? I don't quite see how this solves the issue you described. Also, Velocity just forwards the cookie packets as they are, if no plugin intervenes via the cookie events. The plugin you mentioned doesn't seem to even touch cookies. Correct me if I'm wrong, but I think, some other plugin on the proxy or backend server just sent an invalid cookie request. Could you perhaps provide logs, I think they might help?

R00tB33rMan:
https://mclo.gs/PAeWivU - Here's log A, which contains the issue. This only appears for the client, not for the actual proxy.
https://mclo.gs/bkgmv9e - This is log B, which does not contain the issue and is the "fixed" build.

image [User's client without the fix]

image [User's client with the fix]

I can very much confirm that the fix applied did manage to resolve the issue. The title purely reflects what the error mentions, so I do apologize if that is inaccurately defined. I can confirm that this occurs with only the v4Guard plugin on its own.

Client log: https://mclo.gs/fBLiagN

Luccboy:
This seems to be a weird issue and I don't understand why and how it occures. When I disconnect a player during login phase, everything works fine. I'm not sure if queuing disconnects until play phase is reached, is a proper solution, even though it works in your particular situation.

electronicboy:
I'm out of action due to surgery recovery, but, I'd imagine that this is a protocol state desync, disconnection packet in LOGIN is the same ID as the cookie request packet in CONFIG

Dinos3396:
I am having the same issue with my plugin for velocity, the client no longer recieves the proper disconnect message send by the plugin and instead its getting this cookie request error.

0utplay:
We are facing a similar issue that is caused by this. BungeeCord has a similar approach of queueing packets that are not allowed in the current state of the player https://github.com/SpigotMC/BungeeCord/blob/master/proxy/src/main/java/net/md_5/bungee/UserConnection.java#L187

The alternative would involve having checks for the correct state on each call from the caller site

R00tB33rMan:
I am closing this PR, as it isn't the right approach to resolving this issue, which was fixed by correctly specifying how the phase and event should be handled.

Issue #1127, opened 2023-10-28
Failed to validate message acknowledgements


7wOv6ySCjo:
Hello everybody. I built a minecraft:tellraw command on the page: www.minecraftjson.com

I run this command in the console of my backend server Paper:
`minecraft:tellraw Sistan ["",{"text":"JUEGOS:","bold":true,"color":"#F97603"}," ","Sistan"," acaba de completar el parkour ",{"text":"[Angeles]","color":"dark_purple","clickEvent":{"action":"run_command","value":"/angeles"},"hoverEvent":{"action":"show_text","contents":[{"text":"Hacé click para ir","color":"green"}]}}]`

The tellraw message arrives at the client Minecraft 1.19.4 and I simply click on one of the hoverEvent in the message. Velocity kicks me with: `Failed to validate message acknowledgements from Sistan`

Logs:
Velocity:
```
[04:25:57.642] [Velocity Async Event Executor - #14/INFO] [com.velocitypowered.proxy.protocol.packet.chat.CommandHandler]: [connected player] Sistan (/###.###.##.###:63108) -> executed command /angeles
[04:25:57.668] [Netty epoll Worker #13/INFO] [com.velocitypowered.proxy.connection.client.ConnectedPlayer]: [connected player] Sistan (/###.###.##.###:63108): kicked from server survival_oscuro: 
[04:25:57.668] [Netty epoll Worker #13/INFO] [com.velocitypowered.proxy.connection.MinecraftConnection]: [server connection] Sistan -> survival_oscuro has disconnected
```

Backend:
```
[04:25:57.643] [Netty Epoll Server IO #3/WARN]: Failed to validate message acknowledgements from Sistan
[04:25:57.661] [Server thread/INFO]: Sistan lost connection: Chat message validation failure
```
What solution could I use to prevent this from happening? It is happening massively.

456dev:
does just running the command "/angeles" manually cause the same issue?
does it occur with any other commands in the tellraw?

7wOv6ySCjo:
> does just running the command "/angeles" manually cause the same issue? does it occur with any other commands in the tellraw?

happens with any clickable command in the message.

Jeremylo1:
> does just running the command "/angeles" manually cause the same issue? does it occur with any other commands in the tellraw?

I encounter the same problem! From time to time a same clickable message sent with tellraw command can be executed or disconnect the player with the following message in the console :
[...]

PR #1366, opened 2024-07-03
A couple of fixes for handling of secure chat

Gegy:
This PR addresses a couple of issues in the implementation of secure chat handling in Velocity:
* Any command with a 'last seen' update (every command pre-1.20.5) could not be cancelled by the proxy
    * This PR introduces a system to 'delay' these acknowledgements to ensure that we can always pass a state to the server that is consistent with messages signed by the client
* Modified or spoofed commands would not have a last seen update attached, which would be rejected by pre-1.20.5 servers

Most of the diff is refactoring `ChatQueue` to be able to capture the state that we need - recommend to review commit-by-commit.

This potentially resolves some cases of #1127 - but it's hard to say given possible interaction from other plugins and servers, and would need to get some specific testing on that. However, from personal experience running these patches on a 1.20.4 server over a few months - we haven't been able to reproduce any issues. 🙂 

Thanks!

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/chat/session/SessionCommandHandler.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/chat/session/SessionCommandHandler.java
@@ -20,10 +20,12 @@ package com.velocitypowered.proxy.protocol.packet.chat.session;
 import com.velocitypowered.api.event.command.CommandExecuteEvent;
 import com.velocitypowered.proxy.VelocityServer;
 import com.velocitypowered.proxy.connection.client.ConnectedPlayer;
+import com.velocitypowered.proxy.protocol.MinecraftPacket;
 import com.velocitypowered.proxy.protocol.packet.chat.ChatAcknowledgementPacket;
 import com.velocitypowered.proxy.protocol.packet.chat.CommandHandler;
 import java.util.concurrent.CompletableFuture;
 import net.kyori.adventure.text.Component;
+import org.checkerframework.checker.nullness.qual.Nullable;
 
 public class SessionCommandHandler implements CommandHandler<SessionPlayerCommandPacket> {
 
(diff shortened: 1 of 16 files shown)


proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/chat/session/SessionCommandHandler.java
```diff
@@ -18,11 +18,12 @@
 package com.velocitypowered.proxy.protocol.packet.chat.session;
 
 import com.velocitypowered.api.event.command.CommandExecuteEvent;
-import com.velocitypowered.api.network.ProtocolVersion;
 import com.velocitypowered.proxy.VelocityServer;
 import com.velocitypowered.proxy.connection.client.ConnectedPlayer;
+import com.velocitypowered.proxy.protocol.MinecraftPacket;
 import com.velocitypowered.proxy.protocol.packet.chat.ChatAcknowledgementPacket;
 import com.velocitypowered.proxy.protocol.packet.chat.CommandHandler;
+import javax.annotation.Nullable;
```
powercasgamer:
```suggestion
import org.checkerframework.checker.nullness.qual.Nullable;
```

proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/chat/session/SessionPlayerCommandPacket.java
```diff
@@ -25,6 +25,8 @@
 import com.velocitypowered.proxy.protocol.packet.chat.LastSeenMessages;
 import com.velocitypowered.proxy.util.except.QuietDecoderException;
 import io.netty.buffer.ByteBuf;
+
+import javax.annotation.Nullable;
```
powercasgamer:
```suggestion
import org.checkerframework.checker.nullness.qual.Nullable;
```

proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/chat/session/UnsignedPlayerCommandPacket.java
```diff
@@ -19,8 +19,11 @@
 
 import com.velocitypowered.api.network.ProtocolVersion;
 import com.velocitypowered.proxy.protocol.ProtocolUtils;
+import com.velocitypowered.proxy.protocol.packet.chat.LastSeenMessages;
 import io.netty.buffer.ByteBuf;
 
+import javax.annotation.Nullable;
```
powercasgamer:
```suggestion
import org.checkerframework.checker.nullness.qual.Nullable;
```

powercasgamer (changes requested):
minor import nitpick

kashike (approved):

Issue #1368, opened 2024-07-04
PlayerEnterConfigurationEvent is not called on initial connect


scrayos:
### Expected Behavior

I expect the event to be always called, regardless of whether it's the first server or any other server.

### Actual Behavior

The event is only triggered for server switches, not for the initial connection

### Steps to Reproduce

1. Listen for the event
2. Join the server
3. See that the event was not triggered

### Plugin List

Only the test plugin

### Velocity Version

[23:18:18 INFO]: Velocity 3.3.0-SNAPSHOT (git-aa4e8780-b401)
[23:18:18 INFO]: Copyright 2018-2023 Velocity Contributors. Velocity ist lizenziert unter den Bedingungen der GNU General Public License v3.
[23:18:18 INFO]: velocitypowered.com - GitHub

Issue #1367, opened 2024-07-04
Timeout after 15 seconds in CONFIGURATION state (1.21 backend server)


scrayos:
### Expected Behavior

I expect clients to be able to stay in configuration phase indefinitely. There are [special keep alive packets](https://wiki.vg/Protocol#Clientbound_Keep_Alive_.28configuration.29) that should allow clients to stay connected that way, without any limitation. 

### Actual Behavior

Clients are always disconnected after 15 seconds.

### Steps to Reproduce

1. Add the listener below
2. Connect to velocity
3. Get disconnected after 15 seconds

```java
    @Subscribe
    public EventTask onConfigurationWait(@NotNull final PlayerFinishConfigurationEvent e) {
        return EventTask.resumeWhenComplete(new CompletableFuture<>());
    }
 ```

### Plugin List

Only the test plugin

### Velocity Version

[22:58:43 INFO]: Velocity 3.3.0-SNAPSHOT (git-aa4e8780-b401)
[22:58:43 INFO]: Copyright 2018-2023 Velocity Contributors. Velocity ist lizenziert unter den Bedingungen der GNU General Public License v3.
[22:58:43 INFO]: velocitypowered.com - GitHub

Gerrygames:
I can't reproduce this during the initial configuration. This seems to only affect reconfiguration when switching servers.

scrayos:
> I can't reproduce this during the initial configuration. This seems to only affect reconfiguration when switching servers.

I actually only tested this for the initial connection. I installed this listener and then joined the server. After 15 seconds, I reliably time out. I did not even test reconfiguration on server-switch, so that's weird. And you were able to stay connected this way indefinitely on initial connect?

Did you modify the configuration on either velocity or your backend server? (I kept all defaults)

Gerrygames:
This was an issue when the configuration API was still WIP. Users that took longer than 30 seconds to download our resource pack timed out. I did fix this before the configuration API was merged. I explicitly tested this again by holding the PlayerFinishConfigurationEvent for 30 seconds and could not reproduce this issue. I am running Paper 1.20.4 and tested with 1.20.4 and 1.21 client (with ViaVersion). There aren't any modifications done to the configuration progress. I will test again with Paper 1.21, maybe there is some kind of difference.

Gerrygames:
[...]

PR #1371, opened 2024-07-05
Add PlayerConfigurationEvent and PlayerEnteredConfigurationEvent

Gerrygames:
This addresses issues #1367 and #1368 

- Configuring the player (i.e. sending resource packs) should now be done in the new PlayerConfigurationEvent.
- The new PlayerEnteredConfigurationEvent is called when a player acknowledged the switch to configuration state.
- The PlayerEnterConfigurationEvent is no longer called twice. It is now called when the backed wants to reconfigure the player. Velocity will wait (up to 5 seconds) for this event and only then ask the client to switch to configuration state.
- The PlayerFinishConfigurationEvent should no longer be used to configure the player (i.e. sending resource packs). This is because since 1.20.5 the backend server can't send keep alive packets between switching state anymore and the connection will thus time out. Velocity will only wait 5 seconds for this event.
- Improved java docs. Feedback would be appreciated.

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/event/player/configuration/PlayerConfigurationEvent.java
+++ b/api/src/main/java/com/velocitypowered/api/event/player/configuration/PlayerConfigurationEvent.java
@@ -13,15 +13,14 @@ import com.velocitypowered.api.proxy.ServerConnection;
 import org.jetbrains.annotations.NotNull;
 
 /**
- * This event is executed when a player is about to enter the configuration state.
- * It is <b>not</b> called for the initial configuration of a player after login.
- * <p>Velocity will wait for this event before asking the client to enter configuration state.</p>
+ * This event is executed when a player entered the configuration state and can be configured by Velocity.
+ * <p>Velocity will wait for this event before continuing/ending the configuration state.</p>
  *
- * @param player The player who is about to enter configuration state.
- * @param server The server that wants to reconfigure the player.
+ * @param player The player who can be configured.
+ * @param server The server that is currently configuring the player.
  * @since 3.3.0
  * @sinceMinecraft 1.20.2
  */
 @AwaitingEvent
-public record PlayerEnterConfigurationEvent(@NotNull Player player, ServerConnection server) {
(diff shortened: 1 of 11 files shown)


4drian3d (approved):

Timongcraft:
Is it intentional that the new `PlayerConfigurationEvent` is only called for the initial configuration and not for subsequent ones, unlike the (_old_) `PlayerFinishConfigurationEvent`?

Gerrygames:
No that's a mistake then. I will investigate that later.

Timongcraft:
I just noticed that the Javadoc for the `PlayerEnterConfigurationEvent` specifies that it is called **before** the client enters the configuration state. However, it also states that it only waits 5 seconds because the backend can't keep the connection alive during the configuration state. Since the client is not in the configuration state at that moment, does this mean that the backend will terminate the connection because it requested the change to the configuration state? If so, I think the wording is a bit confusing.

Gerrygames:
The "clientbound" connection between velocity and the backend is already in configuration state but the clientbound connection between velocity and the client is still in play state. The java doc states that the backend can't keep the connection alive during state changes which is correct. If you have any better wording for it, you're welcome to suggest it.

Timongcraft:
How about this?
```html
 * <p>Velocity will wait for this event before asking the client to enter configuration state.
 * <br>However, for the backend server, the player is already in the configuration state. Since the server cannot maintain the connection during state changes,
 * Velocity will only wait for a maximum of 5 seconds.</p>
 ```

> The "clientbound" connection between velocity and the backend is already in configuration state but the clientbound connection between velocity and the client is still in play state. The java doc states that the backend can't keep the connection alive during state changes which is correct. If you have any better wording for it, you're welcome to suggest it.

Timongcraft:
> No that's a mistake then. I will investigate that later.

Have you found anything?

Gerrygames:
> > No that's a mistake then. I will investigate that later.
> 
> 
> 
> Have you found anything?

Fixed by https://github.com/PaperMC/Velocity/commit/d999ee20583445f96934e491da72e043ddf4c5af

PR #1382, opened 2024-07-17
Decompose TabCompleteEvent into PlayerTabCompleteEvent and TabCompletionResponseEvent

Turboop8:
Decompose TabCompleteEvent into PlayerTabCompleteEvent and TabCompletionResponseEvent. The original TabCompleteEvent has been deprecated.

Now we can trigger PlayerTabCompleteEvent when the player issues a Tab complete action and modify the "partialMessage" sent by the player to the remote server. When the remote server returns the content of Tab complete, a TabCompletionResponseEvent is triggered and the contents of the returned "suggestions" list can be modified.

If TabCompleteEvent and TabCompletionResponseEvent are listening at the same time and both the suggestions of them have been modified, the modified suggestions provided by TabCompletionResponseEvent will be used.

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/event/player/PlayerTabCompleteEvent.java
+++ b/api/src/main/java/com/velocitypowered/api/event/player/PlayerTabCompleteEvent.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright (C) 2019-2023 Velocity Contributors
+ *
+ * The Velocity API is licensed under the terms of the MIT License. For more details,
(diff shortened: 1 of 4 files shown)


powercasgamer (changes requested):
The original event should still be called. Breaking change otherwise

Turboop8:
> The original event should still be called. Breaking change otherwise

So can i make TabCompleteEvent lower than the priority of TabCompletionResponseEvent, that is, if TabCompletionResponseEvent is cancelled, TabCompleteEvent will not be called

Issue #1384, opened 2024-07-19
Client disconnects when Velocity does a server switch during bundle session


cijaaimee:
### Expected Behavior

Velocity should either prematurely end the bundle session or wait until it's finished before sending any terminal packets (that is, packets mainly related to the configuration stage or to resource packs).

### Actual Behavior

Velocity does not take into account the bundle session and sends packets the Minecraft client qualifies as "terminal", resulting in players disconnecting with an error message saying "Terminal message received in bundle".

### Steps to Reproduce

1. Connect to a MC server through Velocity
2. Send a bundle delimiter packet from the server
3. Switch to another server **without having sent another bundle delimiter packet**
4. Your client should disconnect with a DecoderException with message "Terminal message received in bundle" (see image below)
![error](https://github.com/user-attachments/assets/be840cc5-001a-45a7-96e8-6aa2869fb4e7)

### Plugin List

ViaVersion 5.0.1.

### Velocity Version

We run a fork of Velocity 3.3.0 based off commit 350ea7f31eb40b2c7b28aed0ba070453f48b6e7e. I've however checked the latest Velocity code and I haven't found any fix for this either.

### Additional Information

Here's the list of terminal packets:
* Serverbound client intention packet
* Clientbound game profile packet
* Clientbound start config packet
* Clientbound finish config packet
* Serverbound resource pack packet (if not accepted or downloaded)
* Serverbound finish config packet
* Serverbound config ack packet
* Serverbound login ack packet

Warriorrrr:
Resolved by https://github.com/PaperMC/Velocity/commit/a4476253abc4f4486385aa5dda4d8e6d77332f13

PR #1385, opened 2024-07-19
End bundle session on server switch

cijaaimee:
This is a fix for #1384. I'm not very familiar with Velocity, this is what I've added to our custom fork. Please let me know if this is unsafe and/or not the right approach.

I've had some trouble testing because the Minecraft implementation for bundles makes it hard to send a bundle delimiter packet manually.


electronicboy (approved):

electronicboy:
The logic here seems to sit in the right place for this, i.e. right after we kill the old server and right before we jump into configuration, can't really fault it with my knowledge

Gerrygames (approved):

PR #1398, opened 2024-07-31
[ci skip] make 'runShadow' workingDir use the 'run' directory

powercasgamer:
good: 
jetbrains_client64_qv4zIOrpX0
bad: 
jetbrains_client64_sBozxa6QNQ


electronicboy:
1) Long fabled todo list but didn't wanna break my env
2) I did wanna look at getting the console working inside of IJ, as it's annoying that it doesn't as the process just seems to cause weird freezing issues as-is when tryna shutdown

electronicboy:
(on that note, the task I always use is literally just called run, never needed run shadow, but, if we hit one, we should hit both

powercasgamer:
> (on that note, the task I always use is literally just called run, never needed run shadow, but, if we hit one, we should hit both

Afaik the `run` task doesn't even work?
WindowsTerminal_9QWDGW2jIg

electronicboy:
it screams but it's running fine

powercasgamer:
> (on that note, the task I always use is literally just called run, never needed run shadow, but, if we hit one, we should hit both

Done

electronicboy (approved):

PR #1399, opened 2024-08-01
replace old velocitypowered.com links

RealBauHD:
This PR replaces the old `velocitypowered.com`, because they are now being forwarded anyway, and so we can avoid the detour.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/command/builtin/VelocityCommand.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/command/builtin/VelocityCommand.java
@@ -174,10 +174,10 @@ public final class VelocityCommand {
       if (version.getName().equals("Velocity")) {
         final TextComponent embellishment = Component.text()
             .append(Component.text()
-                .content("velocitypowered.com")
+                .content("papermc.io/software/velocity")
                 .color(NamedTextColor.GREEN)
                 .clickEvent(
-                    ClickEvent.openUrl("https://velocitypowered.com"))
+                    ClickEvent.openUrl("https://papermc.io/software/velocity"))
                 .build())
             .append(Component.text(" - "))
             .append(Component.text()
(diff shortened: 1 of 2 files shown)


4drian3d (approved):

proxy/src/main/java/com/velocitypowered/proxy/command/builtin/VelocityCommand.java
```diff
@@ -174,10 +174,10 @@ public int run(final CommandContext<CommandSource> context) {
       if (version.getName().equals("Velocity")) {
         final TextComponent embellishment = Component.text()
             .append(Component.text()
-                .content("velocitypowered.com")
+                .content("papermc.io/software/velocity")
```
powercasgamer: This looks really meh

RealBauHD: yeah, any suggestions to make it better?

Something like: _Info_

Leguan16: Website?

Twarug: Homepage?

Leguan16: Yeah idk it doesn't really make sense. Maybe since it's a PaperMC software, it can be replaced with `papermc.io`? Or just leave it really it doesn't hurt anybody.

Twarug: At least old link still works and redirects to desired place

electronicboy: if I were held at gunpoint, I'd probably just maybe name it PaperMC, anything else is just pointlessly verbose, imho

proxy/src/main/java/com/velocitypowered/proxy/command/builtin/VelocityCommand.java
```diff
@@ -174,10 +174,10 @@ public int run(final CommandContext<CommandSource> context) {
       if (version.getName().equals("Velocity")) {
         final TextComponent embellishment = Component.text()
             .append(Component.text()
-                .content("velocitypowered.com")
+                .content("PaperMC")
```
RealBauHD, line 177: Any opinions on also underline this?

PR #1406, opened 2024-08-14
Stabilize and expose suggestions API

AlexProgrammerDE:
This PR stabilizes and exposes the internal suggestions API for third-party developers. My usecase for this API is that I'm building a dashboard for running commands on both bukkit and velocity servers with tab completion. 
Right now there is no API to get any command completions for Velocity commands using the API.
Bukkit already has a public API for tab completions: https://jd.papermc.io/paper/1.21.5/org/bukkit/command/CommandMap.html#tabComplete(org.bukkit.command.CommandSender,java.lang.String)
So I think it would be good if velocity also exposed the internal suggestions API.

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/command/CommandManager.java
+++ b/api/src/main/java/com/velocitypowered/api/command/CommandManager.java
@@ -7,8 +7,10 @@
 
 package com.velocitypowered.api.command;
 
+import com.mojang.brigadier.suggestion.Suggestions;
 import com.velocitypowered.api.event.command.CommandExecuteEvent;
 import java.util.Collection;
+import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.function.Predicate;
 import org.checkerframework.checker.nullness.qual.Nullable;
@@ -111,6 +113,25 @@ public interface CommandManager {
    *         Can be completed exceptionally if an exception is thrown during execution.
    */
   CompletableFuture<Boolean> executeImmediatelyAsync(CommandSource source, String cmdLine);
+
+  /**
+   * Returns suggestions to fill in the given command.
+   *
+   * @param source  the source to execute the command for
+   * @param cmdLine the partially completed command
+   * @return a {@link CompletableFuture} eventually completed with a {@link List}, possibly empty
+   */
+  CompletableFuture<List<String>> offerSuggestions(CommandSource source, String cmdLine);
+
+  /**
+   * Returns suggestions to fill in the given command.
+   *
+   * @param source  the source to execute the command for
+   * @param cmdLine the partially completed command
+   * @return a {@link CompletableFuture} eventually completed with {@link Suggestions}, possibly
+   *         empty
+   */
+  CompletableFuture<Suggestions> offerBrigadierSuggestions(CommandSource source, String cmdLine);
 
   /**
    * Returns an immutable collection of the case-insensitive aliases registered
--- a/api/src/main/java/com/velocitypowered/api/command/CommandManager.java
+++ b/api/src/main/java/com/velocitypowered/api/command/CommandManager.java
@@ -115,7 +115,8 @@ public interface CommandManager {
   CompletableFuture<Boolean> executeImmediatelyAsync(CommandSource source, String cmdLine);
 
   /**
-   * Returns suggestions to fill in the given command.
+   * Asynchronously collects suggestions to fill in the given command {@code cmdLine}.
+   * Returns only the raw completion suggestions without tooltips.
    *
    * @param source  the source to execute the command for
    * @param cmdLine the partially completed command
@@ -124,7 +125,8 @@ public interface CommandManager {
   CompletableFuture<List<String>> offerSuggestions(CommandSource source, String cmdLine);
 
   /**
-   * Returns suggestions to fill in the given command.
+   * Asynchronously collects suggestions to fill in the given command {@code cmdLine}.
+   * Returns the brigadier {@link Suggestions} with tooltips for each result.
    *
    * @param source  the source to execute the command for
    * @param cmdLine the partially completed command
--- a/proxy/src/main/java/com/velocitypowered/proxy/command/VelocityCommandManager.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/command/VelocityCommandManager.java
@@ -279,27 +279,14 @@ public class VelocityCommandManager implements CommandManager {
         () -> executeImmediately0(source, cmdLine), eventManager.getAsyncExecutor());
   }
 
-  /**
-   * Returns suggestions to fill in the given command.
-   *
-   * @param source  the source to execute the command for
-   * @param cmdLine the partially completed command
-   * @return a {@link CompletableFuture} eventually completed with a {@link List}, possibly empty
-   */
+  @Override
   public CompletableFuture<List<String>> offerSuggestions(final CommandSource source,
       final String cmdLine) {
     return offerBrigadierSuggestions(source, cmdLine)
         .thenApply(suggestions -> Lists.transform(suggestions.getList(), Suggestion::getText));
   }
 
-  /**
-   * Returns suggestions to fill in the given command.
-   *
-   * @param source  the source to execute the command for
-   * @param cmdLine the partially completed command
-   * @return a {@link CompletableFuture} eventually completed with {@link Suggestions}, possibly
-   *         empty
-   */
+  @Override
   public CompletableFuture<Suggestions> offerBrigadierSuggestions(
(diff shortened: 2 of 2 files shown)


alexstaeding:
Why not use brigadier? Both paper and velocity now support it.

AlexProgrammerDE:
Could you point to a method where I can get the underlying brigadier dispatcher? This PR is not for writing commands, it's for tab completing any command virtually on the proxy for external use. Bukkit already has an API and Velocity has too, but it's internal. I'm making it public with this PR.

AlexProgrammerDE:
Right now the only way to get command completions for commands to external services like Web dashboards is through reflection. But internally there is already an API. This PR exposes that API publicly.

Timongcraft:
Like you said, it is already usable with reflection. It is internal for a reason and not exposed, as it may change in the future.

AlexProgrammerDE:
Well will it ever become stable or public? I don't understand the timeline on this. I haven't seen any other PR introducing a public suggestions API, so I made this one.

AlexProgrammerDE:
If I were to open a PR where I make custom code for this for a public API, I am 100% sure reviewers would say "we already have this API internally, just make the methods public via the API".

AlexProgrammerDE:
If I may ask, what is the reason it was kept internal and no public API was made? I assumed just noone bothered to stabilize it, so I made this PR.

Timongcraft:
> Well will it ever become stable or public? I don't understand the timeline on this. I haven't seen any other PR introducing a public suggestions API, so I made this one.

I'm saying that it is internal and intentionally not exposed because there is no real reason for it to be in the API. As for third-party services, you can still use reflection to access it if needed. In 'normal' usage, Velocity has no reason to parse commands itself, as that is handled by Brigadier.

AlexProgrammerDE:
I don't think a useful feature that already has methods implemented should be only accessible using reflection...
Reflection may break in future releases.
Velocity has those methods to interact with the brigadier dispatcher internally for suggestions I assume?
I'm just making it public with this PR to allow plugins for usecases like mine to not have to rely on reflection for this.

Timongcraft:
> I don't think a useful feature that already has methods implemented should be only accessible using reflection... Reflection may break in future releases. Velocity has those methods to interact with the brigadier dispatcher internally for suggestions I assume? I'm just making it public with this PR to allow plugins for usecases like mine to not have to rely on reflection for this.

Yes, and I'm saying that it should be kept internal because you normally shouldn't need to use it. I understand that reflection can break, and that's what I'm trying to say. If internal APIs are used in plugins, it should be the responsibility of the plugin developers to update their plugins, rather than placing the burden on the maintainers.

AlexProgrammerDE:
Well what does normally mean? I have a totally normal plugin as anyone else. I already can execute commands using the executeAsync API, I just want to also tab complete commands using offerSuggestions. This isn't some crazy obscure feature. Even Bukkit has it. What's not normal about this?

Timongcraft:
> Well what does normally mean? I have a totally normal plugin as anyone else. I already can execute commands using the executeAsync API, I just want to also tab complete commands using offerSuggestions. This isn't some crazy obscure feature. Even Bukkit has it. What's not normal about this?

It is "obscure" because command handling is completely done internally. Also, Velocity has nothing to do with other software.

AlexProgrammerDE:
Well command handling is also done by plugins that register RawCommands. I don't get why this shouldn't be a feature...
You will also find similar features in both BungeeCord API: https://github.com/SpigotMC/BungeeCord/blob/cd56fb32c207b39b9470d66d7c61f68d9f0c7e78/api/src/main/java/net/md_5/bungee/api/plugin/PluginManager.java#L213
and also in Sponge API: https://jd.spongepowered.org/spongeapi/12.0.0-SNAPSHOT/org/spongepowered/api/command/manager/CommandManager.html#complete(java.lang.String)

There's nothing wrong about looking at other software and seeing that it's totally normal for plugins of their platform to use this API. I don't see why Velocity needs to be an exception to this. Brigadier command registration itself is also fully exposed by Velocity.

Timongcraft:
> Well command handling is also done by plugins that register RawCommands. I don't get why this shouldn't be a feature... You will also find similar features in both BungeeCord API: https://github.com/SpigotMC/BungeeCord/blob/cd56fb32c207b39b9470d66d7c61f68d9f0c7e78/api/src/main/java/net/md_5/bungee/api/plugin/PluginManager.java#L213 and also in Sponge API: https://jd.spongepowered.org/spongeapi/12.0.0-SNAPSHOT/org/spongepowered/api/command/manager/CommandManager.html#complete(java.lang.String)
> 
> There's nothing wrong about looking at other software and seeing that it's totally normal for plugins of their platform to use this API. I don't see why Velocity needs to be an exception to this. Brigadier command registration itself is also fully exposed by Velocity.

These APIs were designed for the pre-1.13 command system, which Velocity was not. It is meant to be a "modern" API, supporting the Brigadier-based command system that handles commands internally. Your use case is not a typical plugin that has any in-game effect, so I don't understand why it should be supported.

AlexProgrammerDE:
That it's pre-1.13 has nothing to do with the API itself. Sponge for example would've removed it as they don't support 1.12.2 or below. You can even see Sponge updated their API to allow brigadier suggestions. This PR also uses post-1.13 tab completions as it directly returns CompletableFuture< Suggestions >. The Velocity API I'm making public with this PR is just a light wrapper on top of the brigadier command dispatcher as Velocity doesn't allow getting the brigadier dispatcher outside of the CommandManager., not even for internal code.
Are you saying you would prefer if velocity instead returned a custom wrapper classes like sponge for command completions rather than brigadier classes directly?

AlexProgrammerDE:
Would you prefer if I added @APIStatus.Internal?

Timongcraft:
No, I'm saying
> Your use case is not a typical plugin that has any in-game effect, so I don't understand why it should be supported.

AlexProgrammerDE:
There are many plugins for velocity that don't have in-game effects. Take bungee-prometheus-exporter as an example: https://github.com/weihao/bungeecord-prometheus-exporter
Or any discord bridge plugin that runs on velocity. 
I know not every plugin needs this API, but no plugin needs to use all powerful APIs Velocity has.

AlexProgrammerDE:
My plugin may not be the "average" plugin, but I still think this is a feature Velocity should offer for usecases like mine. That's why so many other platforms offer this API, to allow usecases like mine.

Timongcraft:
A Discord bridge plugin has an in-game effect, and as I see it, the "bungee-prometheus-exporter" is not a *typical* plugin in that it doesn't directly have an in-game effect.
I don't want to say that every plugin should use the API or anything, but I don't understand why something like this should be exposed, as it may just be another burden when changing internals. I'm not a maintainer; I just mean that it has only a minor use case. However, since it is a very minor addition as well, I may just be exaggerating this. I've just seen some features that were not added to the API because they were unsupported.

AlexProgrammerDE:
Well then the maintainers can decide on this. I am fine if this API won't be "stable". But I'd like to ask for this to be included and annotated as @APIStatus.Internal as a minimum. Then I can actually get compilation errors when velocity changes the method signature.

Timongcraft:
> Well then the maintainers can decide on this. I am fine if this API won't be "stable". But I'd like to ask for this to be included and annotated as @APIStatus.Internal as a minimum. Then I can actually get compilation errors when velocity changes the method signature.

If it isn't stable or new API, the API is typically annotated with `@Beta`, but as far as I know, internal API is never exposed.

AlexProgrammerDE:
Well ig the maintainers can decide...
Maybe this is worth beta. But APIStatus.Internal would be fitting. This is not a very big API to expose.

AlexProgrammerDE:
Here a link to a relevant Discord discussion about this PR: https://discord.com/channels/289587909051416579/908507886420910101/1273240566171435128

AlexProgrammerDE:
Hi, any update on this PR? It's been 3 months.

4drian3d (commented):
The only general use I can see for exposing this internal method would be for junit tests, otherwise, I don't see much need to expose it, if you really need it, there is not too much complication in using reflection
Although I am not totally against this pull request, I would appreciate that you put more detailed documentation in the javadocs to be able to differentiate the 2 methods

AlexProgrammerDE:
@4drian3d sorry for the delay, I've now added the javadocs you requested. I don't want this to be reflection-only API because that can break with new versions, this doesn't.

4drian3d (approved):

AlexProgrammerDE:
@4drian3d Hi, could you look into getting this PR merged? It's rather simple and has been open for roughly a year.

AlexProgrammerDE:
Thanks! <3

PR #1413, opened 2024-08-26
feat: Add extra vHost data in ConnectionHandshakeEvent

koneiii:
I've retrieved a feature from BungeeCord that allows you to retrieve additional data.

> Some clients use this method to send data to the server.

Links (French) :
https://gist.github.com/nathan818fr/93516ab7a4968489938c25b097fd1af1
https://github.com/PactifyLauncherExamples/PactifyPlugin/blob/master/bungee/plugin/src/main/java/nz/pactifylauncher/plugin/bungee/HandshakeFix.java

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/event/connection/ConnectionHandshakeEvent.java
+++ b/api/src/main/java/com/velocitypowered/api/event/connection/ConnectionHandshakeEvent.java
@@ -20,10 +20,19 @@ public final class ConnectionHandshakeEvent {
 
   private final InboundConnection connection;
   private final HandshakeIntent intent;
+  private final String extraDataInHandshake;
 
-  public ConnectionHandshakeEvent(InboundConnection connection, HandshakeIntent intent) {
+  /**
+   * Creates a new event.
+   *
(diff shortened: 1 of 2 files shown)


456dev:
i would instead suggest a `public String getRawVirtualHost()` method that returns the string given by clients as-is, rather than only the data following the null byte (seeing as you would need to parse it yourself anyway, and nothing guarantees the format.)

side note, if this if part of the vhost, i would have it as part of the InboundConnection interface, as you may want to access this data after the connection is successful, like the standard vhost

koneiii:
> i would instead suggest a `public String getRawVirtualHost()` method that returns the string given by clients as-is, rather than only the data following the null byte (seeing as you would need to parse it yourself anyway, and nothing guarantees the format.)
> 
> side note, if this if part of the vhost, i would have it as part of the InboundConnection interface, as you may want to access this data after the connection is successful, like the standard vhost

I only made it as a bungeecord, I don't plan to modify it currently, I added it for my use :)

Warriorrrr:
Superseded by #1423

PR #1418, opened 2024-09-01
Do more speculative VarInt reading optimizations

astei:
Specifically, I tweaked `MinecraftVarintFrameDecoder` a little bit, and added some logic to unroll the first iteration of `ProtocolUtils.readVarInt()` if we can.

I haven't evaluated the performance of these changes.


Cryptite:
I see a few people have pulled this into their velocity forks. Any ideas whether this was a worthwhile PR that should be resurrected?

electronicboy:
This was merged already?

Cryptite:
_ignore me_

PR #1422, opened 2024-09-03
feat: Add primitive support for sound api

Timongcraft:
With this PR, the Adventure-provided sound API is **partially** implemented.

It adds the ability to play sounds for a player (from the player itself or other players on the same server) and to stop sounds,
while making the methods fail silently if the sound can't be played/stopped.

What this PR implements:
- A method to play back (custom) sounds for a specific player (from the player itself or other players) for 1.19.3+
(- This also enables the respective methods on the registered server, by forwarding it to the players)
- A method to stop (custom) sounds for players with version 1.19.3+

What this PR **doesn't** implement:
- A method to play back (custom) sounds for a specific player at a position or another entity for players of all versions
- A method to play back (custom) sounds (globally) at a specific location for players of all versions
- A easily way to access a list of vanilla sounds
*- A method to play back (custom) sounds for a specific player for players with version 1.19.2 and below*
*- A method to stop (custom) sounds for players with version 1.19.2 and below*

Clarification:
This PR only implements a sound API for version 1.19.3+ because only in this version is the server able to play a (vanilla) sound for a player without requiring a version-dependent id for the sound.

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/proxy/Player.java
+++ b/api/src/main/java/com/velocitypowered/api/proxy/Player.java
@@ -383,8 +383,14 @@ public interface Player extends
   /**
    * {@inheritDoc}
    *
-   * <b>This method is not currently implemented in Velocity
-   * and will not perform any actions.</b>
+   * <p>Note: This method is currently only implemented for players from version 1.19.3 and above.
+   * <br>A {@link ServerConnection} is required for this to function, so a {@link #getCurrentServer()}.isPresent() check should be made beforehand.
+   *
+   * @param sound the sound to play
+   * @throws IllegalArgumentException if the player is from a version lower than 1.19.3
+   * @throws IllegalStateException if no server is connected
+   * @since 3.3.0
+   * @sinceMinecraft 1.19.3
    */
   @Override
   default void playSound(@NotNull Sound sound) {
@@ -413,8 +419,12 @@ public interface Player extends
   /**
    * {@inheritDoc}
    *
-   * <b>This method is not currently implemented in Velocity
-   * and will not perform any actions.</b>
+   * <p>Note: This method is currently only implemented for players from version 1.19.3 and above.
+   *
+   * @param stop the sound and/or a sound source, to stop
+   * @throws IllegalArgumentException if the player is from a version lower than 1.19.3
+   * @since 3.3.0
+   * @sinceMinecraft 1.19.3
    */
   @Override
   default void stopSound(@NotNull SoundStop stop) {
--- a/api/src/main/java/com/velocitypowered/api/proxy/Player.java
+++ b/api/src/main/java/com/velocitypowered/api/proxy/Player.java
@@ -48,7 +48,7 @@ public interface Player extends
     /* Fundamental Velocity interfaces */
     CommandSource, InboundConnection, ChannelMessageSource, ChannelMessageSink,
     /* Adventure-specific interfaces */
-    Identified, HoverEventSource<HoverEvent.ShowEntity>, Keyed, KeyIdentifiable {
+    Identified, HoverEventSource<HoverEvent.ShowEntity>, Keyed, KeyIdentifiable, Sound.Emitter {
 
   /**
    * Returns the player's current username.
@@ -403,11 +403,13 @@ public interface Player extends
   /**
    * {@inheritDoc}
    *
+   * <p><b>Note</b>: Due to <a href="https://bugs.mojang.com/browse/MC/issues/MC-146721">MC-146721</a>, stereo sounds are always played globally.
+   *
    * <p>Note: This method is currently only implemented for players from version 1.19.3 and above
-   * and requires a present {@link #getCurrentServer}. Additionally, it only supports {@link Sound.Emitter#self()} for now.
+   * and requires a present {@link #getCurrentServer} for the emitting player as well as this player.
    *
    * @param sound the sound to play
-   * @param emitter the emitter of the sound
+   * @param emitter the emitter of the sound; may be another player of this player's server
    * @since 3.4.0
    * @sinceMinecraft 1.19.3
    */
--- a/api/src/main/java/com/velocitypowered/api/proxy/Player.java
+++ b/api/src/main/java/com/velocitypowered/api/proxy/Player.java
@@ -383,27 +383,25 @@ public interface Player extends
   /**
    * {@inheritDoc}
    *
-   * <p><b>Note</b>: Due to <a href="https://bugs.mojang.com/browse/MC/issues/MC-146721">MC-146721</a>, stereo sounds are always played globally in 1.14+.
    *
-   * <p><b>Note</b>: Due to <a href="https://bugs.mojang.com/browse/MC/issues/MC-138832">MC-138832</a>, the volume and pitch are ignored when using this method in 1.14 to 1.16.5.
-   *
-   * <p>Note: This method is currently only implemented for players on 1.19.3+
-   * and requires a present {@link #getCurrentServer} for the emitting player as well as this player.
-   *
-   * @param sound the sound to play
-   * @since 3.4.0
-   * @sinceMinecraft 1.19.3
+   * @apiNote <b>This method is not currently implemented in Velocity
+   *     and will not perform any actions.</b>
+   * @see #playSound(Sound, Sound.Emitter)
+   * @see <a href="https://docs.papermc.io/velocity/dev/pitfalls/#audience-operations-are-not-fully-supported">
+   *   Unsupported Adventure Operations</a>
    */
   @Override
   default void playSound(@NotNull Sound sound) {
-    this.playSound(sound, Sound.Emitter.self());
   }
 
   /**
    * {@inheritDoc}
    *
-   * <b>This method is not currently implemented in Velocity
-   * and will not perform any actions.</b>
+   * @apiNote <b>This method is not currently implemented in Velocity
+   *     and will not perform any actions.</b>
+   * @see #playSound(Sound, Sound.Emitter)
+   * @see <a href="https://docs.papermc.io/velocity/dev/pitfalls/#audience-operations-are-not-fully-supported">
+   *   Unsupported Adventure Operations</a>
    */
   @Override
   default void playSound(@NotNull Sound sound, double x, double y, double z) {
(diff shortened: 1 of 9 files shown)


Timongcraft:
TODO:
- [x] Resolve merge conflicts

nicolube:
What is the state of this? I would like to have this feature very much.

Timongcraft:
From my end, it is finished, although I would think it makes more sense to make the methods fail silently again.

Timongcraft:
~~We could also look into playing sounds from another player, but if so, I think that should be discussed first.~~

nicolube:
@astei Could this be reviewed pls.

4drian3d (approved):

nicolube:
@4drian3d Since it has been approved, can this also be merged?

TheXplore:
What is the state of this? We are currently discussing implementing this ourselves, but we would way prefer having it implemented in upstream.

Timongcraft:
Apart from maybe adding Javadoc to the RegisteredServer, **this** PR is ready from my end.

TheXplore:
@4drian3d Do you think additional Javadoc entries are required? Is there an estimate when this might get merged?

nicolube:
@electronicboy 
Sry for the ping, but what is blocking this from being merged?

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
```diff
@@ -1026,6 +1030,31 @@ void setClientBrand(final @Nullable String clientBrand) {
     this.clientBrand = clientBrand;
   }
 
+  @Override
+  public void playSound(@NotNull Sound sound, @NotNull Sound.Emitter emitter) {
+    Preconditions.checkNotNull(sound, "sound");
+    Preconditions.checkNotNull(emitter, "emitter");
+    Preconditions.checkArgument(emitter.equals(Sound.Emitter.self()), "non-self emitter not supported");
```
kezz: Better to fail silently, that's more adventure

api/src/main/java/com/velocitypowered/api/proxy/Player.java
```diff
@@ -403,18 +403,26 @@ default void playSound(@NotNull Sound sound, double x, double y, double z) {
   /**
    * {@inheritDoc}
    *
-   * <b>This method is not currently implemented in Velocity
-   * and will not perform any actions.</b>
+   * <p>Note: This method is currently only implemented for players from version 1.19.3 and above
+   * and requires a present {@link #getCurrentServer}. Additionally, it only supports {@link Sound.Emitter#self()} for now.
```
kezz: If we're tracking all the entity IDs now, other emitters could be supported right? Not a requirement for this PR, just a future thought.

kezz (changes requested):

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
kezz: Or possibly just ignore the emitter and play normally

kezz: Also, any reason you only override play sound with an emitter?

api/src/main/java/com/velocitypowered/api/proxy/Player.java
Timongcraft: Yea, like other players but I found out this:
> I did not add support to use other players as emitter, as custom sounds played like that will just appear like if they were coming from the player itself.

And thought it wasn't worth it, right?

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
Timongcraft: > Also, any reason you only override play sound with an emitter?

Yes, because the other one indicates it uses the position of the player, which wouldn't follow it. So I moved the implemented method [here](https://github.com/PaperMC/Velocity/pull/1422/commits/6ec69dc8b3a6223f30fe32e7fe837de940dd1d0e).
Audience javadoc for normal playSound method: https://javadoc.io/doc/net.kyori/adventure-api/latest/net/kyori/adventure/audience/Audience.html#playSound(net.kyori.adventure.sound.Sound)

Timongcraft: But also if we would add more emitters in the future I thought it would make sense to enforce the use of the self emitter, should we really ignore it?

api/src/main/java/com/velocitypowered/api/proxy/Player.java
kezz: That sounds like you were testing with sounds with the wrong number of channels - they definitely should follow the player.

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
kezz: Ah yes, and velocity doesn't track the location so we can't use that packet. 

It's API contract for unsupported adventure methods to not throw exceptions.

api/src/main/java/com/velocitypowered/api/proxy/Player.java
Timongcraft: wdym? I used the player entity sound packet with a entity id of another player, in I think 1.21.4, and it just acted like the sound was coming from the player I sent the packet to. But I could test again.

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
Timongcraft: Alright, then I'll make it silent was well

Timongcraft: ~~Are null checks exempt? Because I see them in some places or should everything fail silently?~~

api/src/main/java/com/velocitypowered/api/proxy/Player.java
Timongcraft: Yea, in my testing (new setup, 1.21.4, with paper and with changed velocity code) when using another player as emitter it only is played from the other player if it is 1. a vanilla sound, and 2. uses the `stream` parameter like `music_disc.cat` but not `minecraft:item.goat_horn.sound.2`.
And thus I did not find it useful and too complicated to understand to include.

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
Timongcraft: @kezz, what do you say?
(also to the other conversation below)

api/src/main/java/com/velocitypowered/api/proxy/Player.java
kezz: If you have a stereo sound it will never play following the entity and instead will play globally. https://bugs.mojang.com/browse/MC-146721

Timongcraft: Ah thanks, didn't know about that.
I'll test and implement it again.
Would it make sense to add that to the adventure javadoc, I'd open a pr if you want?

Timongcraft:
Even though the sound should be ignored for invalid entity ids, I still choose to test for the current server and thus require it on the receiver because afaik they are not that unique and if there's some other entity with that id exists, it could cause unwanted consequences.

kezz (approved):

api/src/main/java/com/velocitypowered/api/proxy/Player.java
kezz: Would be a good addition to the Sound class yes.

Timongcraft: > Would be a good addition to the Sound class yes.

Why the '[Sound](https://github.com/KyoriPowered/adventure/blob/1b8f5e418ebb26b12b6f3755d16bfaa7ada8a548/api/src/main/java/net/kyori/adventure/sound/Sound.java)' class instead of [the specific 'playSound' method in 'Audience'](https://github.com/KyoriPowered/adventure/blob/1b8f5e418ebb26b12b6f3755d16bfaa7ada8a548/api/src/main/java/net/kyori/adventure/audience/Audience.java#L668)?

Krymonota:
Is there any chance that this could be merged soon? I'd love to use it! :)

Timongcraft:
> Is there any chance that this could be merged soon? I'd love to use it! :)

I'd wait for 1.21.6 and see how to handle the implementation of Adventure for [the new "UI" sound category](https://www.minecraft.net/en-us/article/minecraft-1-21-6-pre-release-3).

Edit: Seems like it was just added to the end of the enum, in vanilla and adventure, so it shouldn't be a problem.

TheXplore:
Hi, is there a current estimate when this might get merged? :)

4drian3d (approved):
lgtm
It works quite well

Timongcraft (commented):
I did not implement the Player#playSound(Sound) method as adventure defines it like this:
> Plays a sound at the location of the recipient of the sound.

Since we can't fulfill that, would it be fine to break that contract by replacing the javadoc?

https://github.com/KyoriPowered/adventure/blob/-/api/src/main/java/net/kyori/adventure/audience/Audience.java#L629C6-L631

4drian3d:
> I did not implement the Player#playSound(Sound) method as adventure defines it like this:
> 
> > Plays a sound at the location of the recipient of the sound.
> 
> Since we can't fulfill that, would it be fine to break that contract by replacing the javadoc?
> 
> https://github.com/KyoriPowered/adventure/blob/-/api/src/main/java/net/kyori/adventure/audience/Audience.java#L629C6-L631

_I'll be honest, I tried this pull request for 30 minutes without knowing why the sound wasn't playing, since stopSound was working for me, until I saw that the `#playSound(Sound)` method wasn't implemented_

From a plugin developer's perspective, this method should be implemented, but from a contract compliance perspective, it's okay not to implement it as required by the contract. With this in mind, good documentation at https://docs.papermc.io/velocity/dev/pitfalls/#audience-operations-are-not-fully-supported would suffice, so I am now reverting my change

Timongcraft:
What do you say about stuffing override dialog api methods into this pr as well for documentation purposes?

4drian3d (approved):

RealBauHD (approved):

PR #1423, opened 2024-09-03
Expose raw vhost string

456dev:
Supersedes #1413 in a more general purpose way, including bypassing other hostname cleaning methods, and implementing it on all inbound connections.

this should be as raw as possible as to what the client sends.

I'm not 100% sure if there is a desire to expose this tho

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/proxy/InboundConnection.java
+++ b/api/src/main/java/com/velocitypowered/api/proxy/InboundConnection.java
@@ -26,10 +26,19 @@ public interface InboundConnection {
 
   /**
    * Returns the hostname that the user entered into the client, if applicable.
-   *
+   * <br/>
+   * This is partially processed, including removing a trailing dot, and discarding data after a null byte.
+
    * @return the hostname from the client
    */
   Optional<InetSocketAddress> getVirtualHost();
+
+  /**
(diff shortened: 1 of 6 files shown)


electronicboy (approved):

electronicboy:
I'm inclined to lean towards this, given the whole "putting data inside the vhost" isn't really protocol compliant, etc, and the nul byte is pretty much a weird thing which only exists as an agreed upon thing because everybody copied forge. Who knows what people might wanna do here in the future, not really our role to parse data out of it that we don't need.

Flash303:
> I'm inclined to lean towards this, given the whole "putting data inside the vhost" isn't really protocol compliant, etc, and the nul byte is pretty much a weird thing which only exists as an agreed upon thing because everybody copied forge. Who knows what people might wanna do here in the future, not really our role to parse data out of it that we don't need.

Yes, but we should still allow the option to retrieve the address sent by the client. Some clients use this data to be identified by the server. I don't see any issue with giving developers the ability to access this information.

electronicboy:
Yes, which is why I'm supporting this PR rather than the other PR, which tries to process that data on its own

Flash303:
> Yes, which is why I'm supporting this PR rather than the other PR, which tries to process that data on its own

Ah yes, do you know when this feature will be available in a production release?

PR #1425, opened 2024-09-06
use vanilla limit for KnownPacks

Outfluencer:
vanilla has not limit for it in latest, its limit is the limit of the collection decode so its 65536

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/config/KnownPacksPacket.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/config/KnownPacksPacket.java
@@ -26,7 +26,7 @@ import io.netty.buffer.ByteBuf;
 
 public class KnownPacksPacket implements MinecraftPacket {
 
-    private static final int MAX_LENGTH_PACKS = Integer.getInteger("velocity.max-known-packs", 64);
+    private static final int MAX_LENGTH_PACKS = Integer.getInteger("velocity.max-known-packs", 65536);
     private static final QuietDecoderException TOO_MANY_PACKS =
         new QuietDecoderException("too many known packs");
 


Outfluencer:
Proofs
image

image (1)

image (2)

Timongcraft:
This is also limited as a security measure and not only to match vanilla behaviour. View #1303 for details.

Outfluencer:
> This is also limited as a security measure and not only to match vanilla behaviour. View #1303 for details.

65536 is as secure as 64

You cant make the server go out of memory with an array of the size 65536.

Every sting in velocity can be that size

the limit is only about preventing sending Integer.MAX_VALUE as size, but such small values are secure anyway

Timongcraft:
it is about the packs and not the size of packs.

Outfluencer:
okay what risks do i encounter if the limit is at 65536, compared to 64?

Outfluencer:
as the data is limited to 2 mb or 6 mb with compression anyways there is litrally no security risk with using the vanilla limit here

Outfluencer:
> it is about the packs and not the size of packs.

Also i think you don't understand its about the array allocation and not the KnownPacks amount itself
The problem was that you could allocate the following: new KnownPack[2000000000] in that case the jvm would throw an OutOfMemoryException or crash.

Just so you can understand better 65536 is 30517,57 times smaller, so it cant cause any out of memory errors or performace issues

UserNugget:
> vanilla has not limit for it in latest, its limit is the limit of the collection decode so its 65536

Its still in the latest vanilla version. You are looking at clientbound limit, but there are also a serverbound one.
And it seems that i forgot to add clientbound limit in https://github.com/PaperMC/Velocity/pull/1303 👍

Outfluencer:
> > vanilla has not limit for it in latest, its limit is the limit of the collection decode so its 65536
> 
> Its still in the latest vanilla version. You are looking at clientbound limit, but there are also a serverbound one. And it seems that i forgot to add clientbound limit in #1303 👍

You are right, my friend send me the wrong packet.

        STREAM_CODEC = StreamCodec.composite(KnownPack.STREAM_CODEC.apply(ByteBufCodecs.list(64)), ServerboundSelectKnownPacks::knownPacks, ServerboundSelectKnownPacks::new);

Timongcraft:
At least I understand it correctly now, thanks

UserNugget:
> > > vanilla has not limit for it in latest, its limit is the limit of the collection decode so its 65536
> > 
> > 
> > Its still in the latest vanilla version. You are looking at clientbound limit, but there are also a serverbound one. And it seems that i forgot to add clientbound limit in #1303 👍
> 
> You are right, my friend send me the wrong packet.
> 
> ```
>     STREAM_CODEC = StreamCodec.composite(KnownPack.STREAM_CODEC.apply(ByteBufCodecs.list(64)), ServerboundSelectKnownPacks::knownPacks, ServerboundSelectKnownPacks::new);
> ```

No need to close this PR, KnownPacksPacket is used for both clientbound and serverbound directions, so this change is still useful (but should be changed to affect only clientbound direction).

PR #1427, opened 2024-09-14
Fix ShutdownCommand message styling

456dev:
the docs specify the command as accepting json components, as well as minimessage
in the previous implementation, neither were true, as the minimessage deserialisation after parsing legacy escaped any tags internally.

This PR fixes that by only deserialising the string once before parsing it on.
due to this, legacy "&" support has been removed

to support parsing as both minimessage + json, parsing as json will only kick in if it looks like a valid json text component:
- "plain format, starting with a quote."
- {"text":"Object format, starting with a curly brace"}
- ["Array format", {"text": "starting with a square bracket"}]

Docs PR: PaperMc/docs#493


electronicboy (approved):

proxy/src/main/java/com/velocitypowered/proxy/command/builtin/ShutdownCommand.java
```diff
@@ -52,11 +54,22 @@ public static BrigadierCommand command(final VelocityServer server) {
...
             .executes(context -> {
               String reason = context.getArgument("reason", String.class);
-              server.shutdown(true, MiniMessage.miniMessage().deserialize(
-                  MiniMessage.miniMessage().serialize(
-                      LegacyComponentSerializer.legacy('&').deserialize(reason)
-                  )
-              ));
+              Component reasonComponent = null;
+
+              if (reason.startsWith("{") || reason.startsWith("[") || reason.startsWith("\"")) {
```
novitpw, line 60: maybe we should do it like this?

```java
        char firstChar = reason.charAt(0);
        char lastChar = reason.charAt(reason.length() - 1);

        if ((firstChar == '{' && lastChar == '}')
                || (firstChar == '[' && lastChar == ']')
                || (firstChar == '"' && lastChar == '"')) {
                try {
                    reasonComponent = GsonComponentSerializer.gson()
                            .deserializeOrNull(reason);
                } catch (JsonSyntaxException e) {
                    log.warn("Failed to parse reason as JSON object: {}", reason, e);
                }
        }
```

4drian3d (approved):

proxy/src/main/java/com/velocitypowered/proxy/command/builtin/ShutdownCommand.java
electronicboy: generally not fond of printing a warning for that kinda thing, and generally not sure what exactly this would solve vs the current approach

novitpw: this will allow server administrators to know the exact error when formatting is incorrect

novitpw: > generally not fond of printing a warning for that kinda thing, and generally not sure what exactly this would solve vs the current approach


this will allow server administrators to know the exact error when formatting is incorrect

electronicboy: The command accepts json or MiniMessage, the intent is that it will silently fall through to parsing the alt format rather than blowing up

novitpw: > The command accepts json or MiniMessage, the intent is that it will silently fall through to parsing the alt format rather than blowing up

How then to quickly determine where the error occurred? Maybe in that case it is worth trying log.debug?

astei:
@456dev can you fix the checkstyle issue here?

4drian3d (approved):

PR #1435, opened 2024-09-20
feat: Add event to modify and/or block backend updates to the tablist

### LINKED ISSUE #870: Optimize API/Events for TabList

Timongcraft:
This is a *concept* implementation to allow plugins to modify the tablist that is sent by the backend server.
While it is completly functional, there are aspects I'm not quite fond about, for example that there's a bit of duplicate code now.
Please let me know if you think there's a better design that could be implemented!
Also, if you think the version-specific behavior should *somehow* be handled internally, please let me know.

Should fix https://github.com/PaperMC/Velocity/issues/870
Inspired by https://github.com/PaperMC/Velocity/pull/874

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/event/player/ServerUpdateTabListEvent.java
+++ b/api/src/main/java/com/velocitypowered/api/event/player/ServerUpdateTabListEvent.java
@@ -18,7 +18,7 @@ import java.util.Set;
 import java.util.UUID;
 
 /**
- * This event is fired, when a {@link com.velocitypowered.api.proxy.player.TabList Tablist} is updated.
+ * This event is fired, when a {@link com.velocitypowered.api.proxy.player.TabList Tablist} is updated by a {@link com.velocitypowered.api.proxy.ServerConnection server}.
  * It can be used to override or cancel updates for {@link TabListEntry}s.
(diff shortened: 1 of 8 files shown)


Timongcraft:
Since no suggestions/change requests were made on Discord or here, I'm gonna mark this as ready

Timongcraft:
The current implementation is flawed and needs to be re-thought.
(Blocking the event for too long will make nearby players invisible)

Timongcraft:
Closing this, as a implementation would need to hack unlisted info packets or pause multiple connections, and that just doesn't reflect the goal of "designed with performance and stability in mind" IMO.

PR #1436, opened 2024-09-20
Adjust HAProxy's existance to log when the proxy protocol is enabled …

Xeravax:
This is a VERY minor change, so minimal but VERY useful for system administrators. (I am one, and this was so confusing).
Other proxy softwares log when they enable proxy protocol, but velocity doesn't. I probably had some mandella effect, but it seemed like Velocity never used to do this. 

But as a system administrator, it would be nice to check the console when the server boots to check if HAProxy is enabled. Hence this change.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/network/ConnectionManager.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/network/ConnectionManager.java
@@ -109,7 +109,8 @@ public final class ConnectionManager {
           final Channel channel = future.channel();
           if (future.isSuccess()) {
             this.endpoints.put(address, new Endpoint(channel, ListenerType.MINECRAFT));
-            LOGGER.info("Listening on {}", channel.localAddress());
+            
+            LOGGER.info((this.server.getConfiguration().isProxyProtocol() ? "HAProxy listening on " : "Listening on ") + "{}", channel.localAddress());
 
             // Fire the proxy bound event after the socket is bound
             server.getEventManager().fireAndForget(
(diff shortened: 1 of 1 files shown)


clrxbl:
At some point I remember complaining about this on the Discord aswell, but I think it'd be nicer to mimic what BungeeCord does and log it as a warning on startup, letting the user know proxy protocol is enabled and that it can be a security risk, see https://github.com/SpigotMC/BungeeCord/blob/acb85e30faf3aa10d21ab51efa8fe478f13b8361/proxy/src/main/java/net/md_5/bungee/BungeeCord.java#L335

Warning log would also have the advantage that it would be more visible

Xeravax:
I didn't want to be to overbearing. This is a minimal change, but I can adjust it to be a WARNING if needs be

JasmeowTheCat:
Fully approve this honestly, super helpful for me and my team ha. I swear Velocity used to have this but nope, other proxy software integrations do, with the following message for example:

00:04:18 [WARNING] Using PROXY protocol for listener /0.0.0.0:25565, please ensure this listener is adequately firewalled.
00:04:18 [WARNING] Since PROXY protocol is in use, internal connection throttle has been disabled.

I would recommend changing it to WARN.

Xeravax:
@clrxbl @JasmeowTheCat 
How's this? Better?

astei (approved):

electronicboy (approved):

Xeravax:
Wohoo!
image

(You cannot imagine the pain I went through with this fata morgana of a feature)
I had vivid memories of a warning message in velocity. But seems like I was wrong.

PR #1445, opened 2024-10-20
Ability to specify server mapping from the command line

theminecoder:
Also adds the ability to skip registering servers from the config on startup, removing a small step for configuring setups for larger networks.

Open to editing the argument names, they have come from late night thinking

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/ProxyOptions.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/ProxyOptions.java
@@ -17,11 +17,18 @@
 
 package com.velocitypowered.proxy;
 
+import com.velocitypowered.api.proxy.server.ServerInfo;
 import java.io.IOException;
+import java.net.InetSocketAddress;
 import java.util.Arrays;
+import java.util.List;
+
+import com.velocitypowered.proxy.util.AddressUtil;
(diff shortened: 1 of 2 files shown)


proxy/src/main/java/com/velocitypowered/proxy/ProxyOptions.java
```diff
@@ -75,4 +93,40 @@ boolean isHelp() {
...
+      String[] split = s.split(":", 3);
+      if (split.length < 2 || split.length > 4) {
+        throw new ValueConversionException("Invalid server format. Use <name>:<address>");
+      }
+      InetAddress address;
+      try {
+        address = InetAddress.getByName(split[1]);
+      } catch (UnknownHostException e) {
+        throw new ValueConversionException("Invalid hostname for server flag with name: " + split[1]);
+      }
```
astei: Use `AddressUtil.parseAddress()` for consistency with the `velocity.toml` reader. (Long term, I'd love to make Velocity use a YAML file and allow at least the proxy itself to be a 12-factor app, but I lack the time.)

theminecoder: Have swapped over :)

astei (approved):

PR #1453, opened 2024-10-28
Allow responding to cookie requests

masmc05:
Currently, if a server is awaiting the cookie response, it is impossible to block the request without forcing the server to wait indefinably, as <https://jd.papermc.io/velocity/3.4.0/com/velocitypowered/api/event/player/CookieRequestEvent.ForwardResult.html#handled()> just discards any requests without sending any responses and there is no api to manually send a response.

The current approach is similar to <https://jd.papermc.io/velocity/3.4.0/com/velocitypowered/api/event/player/ServerLoginPluginMessageEvent.html>, but if you consider that instead of this, it would be better to add a method in <https://jd.papermc.io/velocity/3.4.0/com/velocitypowered/api/proxy/ServerConnection.html> to send a response then I would change it

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/event/player/CookieRequestEvent.java
+++ b/api/src/main/java/com/velocitypowered/api/event/player/CookieRequestEvent.java
@@ -68,15 +68,19 @@ public final class CookieRequestEvent implements ResultedEvent<CookieRequestEven
    */
   public static final class ForwardResult implements Result {
 
-    private static final ForwardResult ALLOWED = new ForwardResult(true, null);
-    private static final ForwardResult DENIED = new ForwardResult(false, null);
+    private static final ForwardResult ALLOWED = new ForwardResult(true, null, false, null);
+    private static final ForwardResult DENIED = new ForwardResult(false, null, false, null);
 
     private final boolean status;
     private final Key key;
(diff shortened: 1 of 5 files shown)


Luccboy (approved):

electronicboy:
Need to update the toString inside of the event

masmc05:
Applied the requested changes

api/src/main/java/com/velocitypowered/api/event/player/CookieRequestEvent.java
```diff
@@ -68,15 +69,19 @@ public String toString() {
    */
   public static final class ForwardResult implements Result {
```
astei, line 70: I would much prefer a separate `RespondResult` here.

astei (changes requested):

api/src/main/java/com/velocitypowered/api/event/player/CookieRequestEvent.java
masmc05: Then it wouldn't make sense with getResult(). Also, since we're responding on behalf of the player, we also answer on how this should be forwarded, as we can't allow 2 responses (and the client always responds)

PR #1457, opened 2024-11-05
Minecraft 1.21.4

RealBauHD:


kennytv (approved):
Should be fine:tm:

kennytv:
Actually the getter is kind of ugly

proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/UpsertPlayerInfoPacket.java
```diff
@@ -256,6 +262,10 @@ public ComponentHolder getDisplayName() {
       return displayName;
     }
 
+    public boolean getShowHat() {
```
kennytv:
```suggestion
    public boolean isShowHat() {
```
Someone can emergecy change this later if needs be

RealBauHD: Oh yes IntelliJ auto-generated getter moment, thanks Kenny

PR #1459, opened 2024-11-14
Pass through packets in config phase

valaphee:
Re-adds https://github.com/PaperMC/Velocity/pull/1326 which got reverted by https://github.com/PaperMC/Velocity/pull/1371

also adds pass through for the PluginMessagePacket

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientConfigSessionHandler.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientConfigSessionHandler.java
@@ -125,8 +125,10 @@ public class ClientConfigSessionHandler implements MinecraftSessionHandler {
       // but at this time the backend server may not be ready
     } else if (serverConn != null) {
       serverConn.ensureConnected().write(packet.retain());
+      return true;
     }
-    return true;
+
+    return false;
   }
 
   @Override
@@ -142,7 +144,11 @@ public class ClientConfigSessionHandler implements MinecraftSessionHandler {
   @Override
   public boolean handle(KnownPacksPacket packet) {
     callConfigurationEvent().thenRun(() -> {
-      player.getConnectionInFlightOrConnectedServer().ensureConnected().write(packet);
+      if (player.getConnectionInFlight() != null) {
+        player.getConnectionInFlight().ensureConnected().write(packet);
+      } else {
+        handleGeneric(packet);
+      }
     }).exceptionally(ex -> {
       logger.error("Error forwarding known packs response to backend:", ex);
       return null;


proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientConfigSessionHandler.java
```diff
@@ -141,14 +143,17 @@ public boolean handle(PingIdentifyPacket packet) {
 
   @Override
   public boolean handle(KnownPacksPacket packet) {
-    callConfigurationEvent().thenRun(() -> {
-      player.getConnectionInFlightOrConnectedServer().ensureConnected().write(packet);
-    }).exceptionally(ex -> {
-      logger.error("Error forwarding known packs response to backend:", ex);
-      return null;
-    });
+    if (player.getConnectionInFlight() != null) {
```
electronicboy: I think some aspects of this need reverting, the connection isn't always going to be inflight when we jump through configuration, but, events should still be fired, etc

electronicboy (changes requested):

Stis780:
This is very good, thank you!

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientConfigSessionHandler.java
Stis780: So that IF could just be moved in front of 

>        player.getConnectionInFlight().ensureConnected().write(packet);

tho... it might affect plugins listening to config event?

electronicboy: that would prevent plugins from being able to interact with reconfiguration while on a server, this block basically needs reverting

valaphee: you are right theoretically it shouldn't matter if I not handle the packet (handle it via handleGeneric), or handle the packet by default and send the KnownPacksPacket using connectedServer

But for some reason this doesn't work; The only difference is that handleGeneric also checks serverConnection.getPhase().consideredComplete() before sending

electronicboy: Well, does the latest commit work or not, because, that looks pretty on par with what I'd expect from this?

electronicboy: I wouldn't be calling handleGeneric, however, we should ideally be handling all packets inside of this phase, as that we, any packet we don't know about is going to be a modded packet, which is where the special handling gets into play here

valaphee:
Closing this, as this is not a proper fix.

PR #1461, opened 2024-11-18
Fix SystemChatPacket: read boolean instead of varint

shoker537:
Since Velocity does not use SystemChatPacket's decode, it's missing proper attention and has a mistake in decoding.
But plugins managing protocol changes on proxy rely on it, so it's better to keep it working (working code is better than non-working code, right?).

The packet provides a boolean which determines if it is an actionbar message or a chat message, but for some reason SystemChatPacket reads varint which breaks ChatType determination during decode.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/chat/SystemChatPacket.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/chat/SystemChatPacket.java
@@ -47,8 +47,7 @@ public class SystemChatPacket implements MinecraftPacket {
   @Override
   public void decode(ByteBuf buf, ProtocolUtils.Direction direction, ProtocolVersion version) {
     component = ComponentHolder.read(buf, version);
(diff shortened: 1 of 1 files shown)


electronicboy:
it's read as a varint because that's what it used to be, this needs to be properly updated to handle the two different versions of this packet

shoker537:
> it's read as a varint because that's what it used to be, this needs to be properly updated to handle the two different versions of this packet

That's true, I was looking for SystemChatPacket only, but now I see there was ChatMessage packet before 1.19 that actually had 3 states which varint read.

So, it appears the code should look like this if I'm not missing something else:

```java
    if (version.noLessThan(ProtocolVersion.MINECRAFT_1_19_1)){
      type = buf.readBoolean() ? ChatType.GAME_INFO : ChatType.SYSTEM;
    } else {
      type = ChatType.values()[ProtocolUtils.readVarInt(buf)];
    }
```

electronicboy:
Yea, that would look about right to me

electronicboy (approved):

PR #1479, opened 2024-12-19
Fix typo in TabListEntry latency docs

NonSwag:
just fixes a little typo


Timongcraft:
I'd personally rename it to 'greater' as that is often used in code and partly because it is used in the [Ordered](https://github.com/PaperMC/Velocity/blob/dev/3.0.0/api/src/main/java/com/velocitypowered/api/util/Ordered.java) class.

NonSwag:
Yeah, you are right, that does fit better
I'll change it

Timongcraft (approved):

electronicboy (approved):

PR #1481, opened 2024-12-22
Bump adventure to 4.18.0

RealBauHD:


electronicboy:
This requires more than just a simple version bump

RealBauHD:
I went through the changes from adventure and didn't notice anything that needed attention.
Or have I overlooked something in the VirtualComponents?
4.16.0 and 4.17.0 were also just simple bumps.

electronicboy:
The shadow component field needs integrating into our component nbt
serialiser

RealBauHD:
Oh you're right, I look into that

RealBauHD:
It already works
because the serializer translates the gson types and not the components

electronicboy (approved):
I'd forgotten that we'd gone that route in the interim

PR #1485, opened 2025-01-02
Expose shutdownInProgress to the API.

kyngs:
This is a simple PR to expose whether the proxy is currently shutting down or not.
This can be useful for example in the listener for a DisconnectEvent, which is called for every player on proxy shutdown.

If this is intentional and you have no intent of changing this, close this PR.

*The ProxyShutdownEvent is not a substitute for this, as it gets called after all players have been kicked when the proxy shuts down.*


4drian3d (approved):

kyngs:
Any hope of getting this merged?

PR #1495, opened 2025-01-17
Transfer applied resource packs

Gerrygames:


Owen1212055:
Is it worth making this a generic record storage in case additional fields could be transferred in the future? 
For now though it may just be something like:
```java
record TransferSession(Collection<ResourcePackInfo> appliedPacks) {}
```

May just make it easier to add onto in the future.

electronicboy:
The size of a cookie is fairly limited, if we want to store additional info it would make more sense to send additional cookies, IMHO; I guess you could make it more abstracted out, but, we'd still be sending separate cookies out, etc; so, I'm not 100% sure entirely what we'd get trying to combine stuff too much (other than a bunch more code to maintain)

PR #1496, opened 2025-01-19
Add --add-plugin argument

sidey383:
Add --add-plugin for specify path to extra plugins jars to be loaded is additional to those in the plugins folder.
Similar to how it works in paper.


Timongcraft:
Wouldn't it be nicer if it is a system property?

sidey383:
> Wouldn't it be nicer if it is a system property?

There will be problems with separating paths to plugins because system property cannot contain a list of values.
Plus, it will be obvious due to the fact that it works exactly the same as in paper.
https://github.com/PaperMC/Paper/blob/73f7736eb93a64cc0270929cb01992404696d66d/paper-server/src/main/java/org/bukkit/craftbukkit/Main.java#L172

PR #1498, opened 2025-01-20
Some updates

0xSoul24:


4drian3d:
Before updating a dependency, you have to analyze if there are any changes that may affect compatibility in the ecosystem, such as [guice](https://github.com/PaperMC/Velocity/pull/1016) and [snakeyaml](https://github.com/PaperMC/Velocity/pull/1047) updates

PR #1501, opened 2025-01-27
Performance fix/string concatenation

Alvari-Kupari:
Refactored the InformationUtils class to use a stringbuilder when concatenating a string. This boosts performance as concatenating a string in a loop is a costly operation.


astei:
The improvement you made is in a non-performance critical part of the code base. In addition, [JEP 280](https://openjdk.org/jeps/280) permits the JVM to optimize string concatenation, and in fact, the JVM internally rewrites these as `StringBuilder` invocations if that would make sense.

Closing as a non-substantive improvement.

PR #1502, opened 2025-01-31
Use real vanilla limits for LegacyChat

Outfluencer:
Wouldn't https://github.com/PaperMC/Velocity/commit/371e68607695b48117e1cd57d19c67e40384c33a break the chat with this limit??

If the proxy reads a message from the backend that is longer than 256 it would disconnect.

Client -> server
< 1.11 has 100
| >= 1.11 has 256

Server -> client has always 262144


Outfluencer:
@electronicboy right now i think all players will be kicked if the backend sends a message longer than 256 chars

electronicboy (approved):

Outfluencer:
pls unban me on paper dc i fixed some code 👉👈

mbax:
The method for appealing bans is documented in our [community guidelines](https://papermc.io/community/guidelines).

PR #1509, opened 2025-02-09
Adding TabCompleteRequestEvent to be triggered when the client sends a tab to the server

Hexsook:
Added TabCompleteRequestEvent to be triggered when the client sends a tab to the remote server and to get or modify the content sent.


proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientPlaySessionHandler.java
```diff
@@ -783,6 +790,20 @@ private void finishRegularTabComplete(TabCompleteRequestPacket request,
               request, response, ex);
           return null;
         });
+    server.getEventManager().fire(new TabCompleteResultEvent(player, request.getCommand(), offers))
```
novitpw: What is this event for?

novitpw (approved):

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientPlaySessionHandler.java
Hexsook: Used to distinguish from the previous TabCompleteEvent and improve the recognition of the function (that is, triggered when the server produces a result in response to a tab request), although the logic is almost the same as TabCompleteEvent

novitpw: At least I don't see any use for this event, since TabCompleteEvent allows you to do everything you need. Maybe you just didn't delete the code above? If so, I advise you to reconsider the event names.

Hexsook: You do have a point. So I've restored TabCompleteEvent and removed TabCompleteResultEvent in the latest commit, since the two classes do essentially the same thing.

novitpw (approved):

Hengsterx (approved):

PR #1510, opened 2025-02-09
Fix duplicate checks in VarInt writing

jonesdevelopment:
I've noticed that a certain portion of the code in `writeVarInt` performs exactly the same checks as the ones performed in `writeVarIntFull` which is called by `writeVarInt` under certain circumstances.

image

I have merged these methods, which should slightly improve performance and code readability.

The functionality is exactly as before, just without the unnecessary checks in between.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/protocol/ProtocolUtils.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/protocol/ProtocolUtils.java
@@ -17,9 +17,6 @@
 
 package com.velocitypowered.proxy.protocol;
 
-import static com.google.common.base.Preconditions.checkArgument;
-import static com.velocitypowered.proxy.protocol.util.NettyPreconditions.checkFrame;
-
 import com.velocitypowered.api.network.ProtocolVersion;
 import com.velocitypowered.api.proxy.crypto.IdentifiedKey;
 import com.velocitypowered.api.util.GameProfile;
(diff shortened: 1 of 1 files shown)


electronicboy:
I think that the idea here was that by having the shorter method it would make it easier to be compiled and even maybe inlined

jonesdevelopment:
> I think that the idea here was that by having the shorter method it would make it easier to be compiled and even maybe inlined

`writeVarInt` does the same as `writeVarIntFull`, just with extra steps. I don't think inlining would make make much of a difference here.

astei:
As the original author of the code, the function being split like that was completely intentional. The purpose is that writeVarInt would be more likely to be inlined by the JIT and to make the function as small as possible, thus it only contains two cases (which are also the most common ones).

Closing this. I can add a comment to make it clear why it’s like this.

jonesdevelopment:
> As the original author of the code, the function being split like that was completely intentional. The purpose is that writeVarInt would be more likely to be inlined by the JIT and to make the function as small as possible, thus it only contains two cases (which are also the most common ones).
> 
> Closing this. I can add a comment to make it clear why it’s like this.

Alright, thanks a lot for clarifying.

Outfluencer:
> As the original author of the code, the function being split like that was completely intentional. The purpose is that writeVarInt would be more likely to be inlined by the JIT and to make the function as small as possible, thus it only contains two cases (which are also the most common ones).
> 
> Closing this. I can add a comment to make it clear why it’s like this.

Isn't it very likely to be jit compiled anyway? 

Do you have some statistic that proof your point?

i am just curios

Also i read that final methods are even more likely to be inlined

astei:
> Isn't it very likely to be jit compiled anyway?

Well, that's my point: the methods that call into `writeVarInt()` will eventually get JIT compiled, and it would help to have the function inlined if possible.

> Do you have some statistic that proof your point?

The last time I did any work in this area was 2020, and I don't really actively work on Velocity that much any more.

> Also i read that final methods are even more likely to be inlined

This is a `static` method. By definition it _can't_ be overridden, the JVM can precisely resolve the method to use. Also, the JVM is also perfectly capable of doing devirtualization, even when a method isn't marked as `final`: https://blog.tsunanet.net/2012/02/devirtualizing-method-calls-in-java.html

PR #1514, opened 2025-02-18
Add find command

xXkguyXx:
Adds the /find command

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
@@ -40,6 +40,7 @@ import com.velocitypowered.api.util.GameProfile;
 import com.velocitypowered.api.util.ProxyVersion;
 import com.velocitypowered.proxy.command.VelocityCommandManager;
 import com.velocitypowered.proxy.command.builtin.CallbackCommand;
+import com.velocitypowered.proxy.command.builtin.FindCommand;
 import com.velocitypowered.proxy.command.builtin.GlistCommand;
 import com.velocitypowered.proxy.command.builtin.SendCommand;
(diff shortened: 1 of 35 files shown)


xXkguyXx:
@4drian3d Whats the reason for the thumbs down?

electronicboy:
Not really fond of adding more commands to the proxy which are generally better served by plugins which can do this much better thematically, etc. Not sure of the implications of messing with the locale files on the git repo, afaik that mostly should be handled by crowdin from the base english local; and, generally, not sure if I this is because I don't have my glasses but, the code generally looks hard to read

xXkguyXx:
> Not really fond of adding more commands to the proxy which are generally better served by plugins which can do this much better thematically, etc. Not sure of the implications of messing with the locale files on the git repo, afaik that mostly should be handled by crowdin from the base english local; and, generally, not sure if I this is because I don't have my glasses but, the code generally looks hard to read

I can understand that. Just thought it would be a nice QoL thing (along with maybe adding /alert, baffles me why velocity just doesn't have these two basic commands, just thought it would be nice as I've worked on multiple networks and just a convenience thing). 

With that said, I followed the same format as the other builtin commands (essentially just copy pasted from the other command and kept the format / style the same😆). If you think that's a case you should take a look at cleaning those up.

electronicboy:
> baffles me why velocity just doesn't have these two basic commands

Because they're fluff commands used by few which requires a bunch of deliberation around how they're serviced, do we require that you use json for /alert? Because, that would of been the only option for us on the street when making velocity; We generally figured that plugins can do a better job of dealing with the UX aspects and people would standardise on a good solution there which delivers the features they need, especially in a multi-proxy environment.

Things like hardcoding strings are also an obvious no go

PR #1519, opened 2025-02-26
Adding more options for configuring ping passthrough.

TheMiningTeamYT:
When using Velocity, I have been frustrated by the lack of fine control over ping passthrough, specifically the lack of support for passing through favicons without passing through the whole ping request. I've also seen people on line asking how to pass through favicons, so I decided to add the ability to control which aspects of server list ping requests are passed through from the back end server, and which aspects aren't, through 5 new boolean configuration options in the config file.
Please don't judge my code too harshly, this is my first time writing Java, but it does compile and Checkstyle is okay with it.
\(If you're wondering, yes this is a duplicate pull request -- sorry about that, [I was told to make a new pull request from a different branch](https://github.com/PaperMC/Velocity/pull/1290#issuecomment-2670874059). However, the code has been updated since I originally opened the old pull request. Not since I closed the previous pull request though.\)


TheMiningTeamYT:
I cleaned up my commits (specifically I removed the times I merged from upstream from the ping-passthrough-dev branch), so there is actually a reason for this new pull request now. I hope you will reevaluate my pull request in light of this.

TheMiningTeamYT:
I've now gone through the trouble of resolving the merge conflict (so I can have an up-to-date version of Velocity build from my fork), merged ping-passthrough-dev, and tested the result. It builds, and, although I haven't thoroughly tested it, it does work.
I believe this is now in a merge-able state. It functions, maintains compatibility with existing configurations, and brings a useful new feature to Velocity: more granular control of ping passthrough. (I made this fork out of necessity in the first place, and I'm sure others will find this feature useful too). Plus, I've already gone through the trouble of resolving the merge conflicts, and running the build tests (when I merged the branch on my fork). All you have to do is click the button.

kennytv (changes requested):
This should be its own section rather than having the same prefix in the root section

TheMiningTeamYT:
I realized, after looking at the new code for the packet-limiter section, that moving options into a different section is absurdly easy. (So easy that y'all could've suggested it as a change or at least told me "hey add a period after the prefix to make it a section"). But, this pull request has been open so long that the branch it's pulling into is no longer current. I'm going to close this pull request and open a new one against dev/4.0.0.

electronicboy:
Nobody really touches the configuration system outside of a need, and we generally want to migrate away from that system because of it's many flaws including its inability to actually populate new config values without an explict migration being added, etc; Nobody really wants to suggest touching the config system because we want to replace it

PR #1491, opened 2025-01-11
Fix nonsensical deprecation for specifying listener priority

Vankka:
Currently specifying a priority for a listener without getting a deprecation is impossible, and suppressing deprecations causes it's own issues

Screenshot from 2025-01-11 15-05-10

PR #1522, opened 2025-03-01
Improving the API of event priorities

novitpw:
In this request, I propose to improve the event priority API. This will allow most users not to get confused in the API changes that was introduced in 2024 and after the changes that were merged: #1491 

In addition to the current changes, I have two more options for fixing the situation with the event order API:

> Create an `EventPriority` class similar to my current `PostOrder` version and return `PostOrder` to the enum type.

**OR**

> Deprecate and then remove the `Subscribe#priority` method and reuse the `Subscribe#order` method.

Perhaps I am not entirely clear why the `Subscribe#order` method was left unchanged and received the deprecated status.

Looking forward to your feedback!

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/event/EventManager.java
+++ b/api/src/main/java/com/velocitypowered/api/event/EventManager.java
@@ -40,21 +40,6 @@ public interface EventManager {
    * Requests that the specified {@code handler} listen for events and associate it with the {@code
    * plugin}.
    *
-   * @param plugin the plugin to associate with the handler
-   * @param eventClass the class for the event handler to register
-   * @param postOrder the order in which events should be posted to the handler
-   * @param handler the handler to register
-   * @param <E> the event type to handle
-   * @deprecated use {@link #register(Object, Class, short, EventHandler)} instead
-   */
-  @Deprecated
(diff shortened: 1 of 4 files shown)


electronicboy:
We cannot break ABI mid version, especially not in core API

novitpw:
Maybe the next variant will be better?
```java
package com.velocitypowered.api.event;

public interface EventPriority {
    short FIRST = Short.MAX_VALUE - 1;
    short EARLY = Short.MAX_VALUE / 2;
    short NORMAL = 0;
    short LATE = Short.MIN_VALUE / 2;
    short LAST = Short.MIN_VALUE + 1;
}
```

```java
@Deprecated(forRemoval = true)
public enum PostOrder {
    // No has changes 
}
```

Example usage:
```java
@Subscribe(priority = EventPriority.EARLY)
public void onProxyInitialize(LoginEvent event) {
    ...
}
```

electronicboy:
having a class with the conversions would probably be an idea just for the sake of being able to provide a more direct 1:1 mapping for plugin devs; Long term, an ABI break will likely be in order when we get around to doing a proper version bump; sadly, was constrained due to health, and now time constrained because new job

PR #1524, opened 2025-03-07
feat: implement command rate limiter

bridgelol:
Velocity proxies can still be crashed with any other commands besides `/velocity info`. A proper fix for this is to rate limit the frequency of which a client can send commands to the server. Currently there is none.

### DIFF
--- a/api/build.gradle.kts
+++ b/api/build.gradle.kts
@@ -61,6 +61,7 @@ tasks {
         o.encoding = "UTF-8"
         o.source = "17"
 
+        o.use()
         o.links(
             "https://www.slf4j.org/apidocs/",
             "https://guava.dev/releases/${libs.guava.get().version}/api/docs/",
--- a/api/src/main/java/com/velocitypowered/api/proxy/config/ProxyConfig.java
+++ b/api/src/main/java/com/velocitypowered/api/proxy/config/ProxyConfig.java
@@ -164,7 +164,7 @@ public interface ProxyConfig {
   boolean isCancelCommandsIfRateLimited();
 
   /**
-   * Get the rate limit for commands that are rate limited.
+   * Get the kick limit for commands that are rate limited.
    * If this limit is 0 or less, the player will be not be kicked.
    *
    * @return the rate limited command rate limit
@@ -178,5 +178,29 @@ public interface ProxyConfig {
    */
   default boolean isKickOnCommandRateLimit() {
     return getKickAfterRateLimitedCommands() > 0;
+  }
+
+  /**
+   * Get the rate limit for how fast a player can tab complete.
+   *
+   * @return the tab complete rate limit (in milliseconds)
+   */
+  int getTabCompleteRatelimit();
+
+  /**
+   * Get the kick limit for tab completes that are rate limited.
+   * If this limit is 0 or less, the player will be not be kicked.
+   *
+   * @return the rate limited command rate limit
+   */
+  int getKickAfterRateLimitedTabCompletes();
+
+  /**
+   * Get whether the proxy should kick players who are tab complete rate limited.
+   *
+   * @return whether to kick players who are rate limited
+   */
+  default boolean isKickOnTabCompleteRateLimit() {
+    return getKickAfterRateLimitedTabCompletes() > 0;
   }
 }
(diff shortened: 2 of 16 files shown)


electronicboy:
We don't have a means of updating config files properly, and the nature of this will induce weird protocol state if somebody does trigger this; I would generally mostly prefer that we address the bottleneck rather than introducing mechanisms like this which, with the current state of the protocol, creates a whole bunch of other issues if people start configuring this for anything but kicking bad clients

bridgelol:
> We don't have a means of updating config files properly, and the nature of this will induce weird protocol state if somebody does trigger this; I would generally mostly prefer that we address the bottleneck rather than introducing mechanisms like this which, with the current state of the protocol, creates a whole bunch of other issues if people start configuring this for anything but kicking bad clients

You will need some sort of rate limiting mechanism to prevent this, can send thousands of packets a second on any velocity command

bridgelol:
The defaults should also be fine to the extent where most shouldn't need to configure it

bridgelol:
Could also add a hard minimum which isn't configurable so users can't break it

Timongcraft:
Is there a reason why literal kick messages are used instead of translatables?

bridgelol:
Could add, figured it wouldn't be needed as literal messages are used pretty often for kicks (i.e VelocityServer L521)

GrowlyX:
You should probably use translatables and change the English message to "You are tab completing too fast." rather than the message you have right now

Timongcraft:
> Could add, figured it wouldn't be needed as literal messages are used pretty often for kicks (i.e VelocityServer L521)

I don't know why that is, but you should be abel to just add a key to the .properties file(s) and use that in the translatable, we can translate at this point because the player was already connected.
Also what happens if a plugin registered a command with a signed argument, wouldn't this break the chat chain?

electronicboy (approved):
The biggest issue with this is that we have no means to update peoples config files properly, will probably have to figure out some quick mitigation for that in this case, I know that we've been able to handle this somewhat manually but it's far from ideal; I don't really wanna add a bunch of new config options, especially which have side effects, which don't show up in peoples existing config files

Timongcraft:
This pr currently has an issue with signed commands e.g. the built-in `/msg` command.
By setting the `command-rate-limit` to `2000` for demonstration purposes, you can see the command being canceled and any other message with a signed context now disconnecting the client as chat chain was broken when canceling the command.

Quick demo:
https://github.com/user-attachments/assets/84a3334c-a268-41fa-acbe-51257223564b

bridgelol:
> This pr currently has an issue with signed commands e.g. the built-in `/msg` command. By setting the `command-rate-limit` to `2000` for demonstration purposes, you can see the command being canceled and any other message with a signed context now disconnecting the client as chat chain was broken when canceling the command.
> 
> Quick demo: https://github.com/user-attachments/assets/84a3334c-a268-41fa-acbe-51257223564b

My latest commit should resolve this

bridgelol:
> The biggest issue with this is that we have no means to update peoples config files properly, will probably have to figure out some quick mitigation for that in this case, I know that we've been able to handle this somewhat manually but it's far from ideal; I don't really wanna add a bunch of new config options, especially which have side effects, which don't show up in peoples existing config files

Could make the defaults extra tame

bridgelol:
Additionally, now that it forwards the packet to the server upon hitting rate limit, I could disable kicks by default

Timongcraft:
Then would you rename `cancel-commands-if-rate-limited` to smh. like `forward-commands-if-rate-limited`, as cancel doesn't seem appropriate anymore? Or what is the plan for the config?

CullanP:
Is it intentional that tab completes don't reset to 0 when not failing, if a player fails once it never resets so over time it will eventually kick them, the limits currently set are good, but without it resetting to 0 you will get players that are kicked if not adjusted

bridgelol:
> Is it intentional that tab completes don't reset to 0 when not failing, if a player fails once it never resets so over time it will eventually kick them, the limits currently set are good, but without it resetting to 0 you will get players that are kicked if not adjusted

Wasn't intentional no, fixed :D

Skullians:
Any update on this?

electronicboy:
Generally needs testing/additional feedback, been too preoccupied to deal with this myself

Skullians:
> Generally needs testing/additional feedback, been too preoccupied to deal with this myself

Okay, thanks for the clarification.

bridgelol:
Our server and many others medium / large servers have been running this in production for weeks now, without any issues

electronicboy:
I would generally want 3rd party confirmation of that, but I'll look at merging this tomorrow morning unless somebody else gets to it

PR #1542, opened 2025-04-03
fix: typo

SpigotRCE:


Warriorrrr:
This is more than just fixing a typo, why change the default config? The actual default is also not affected by this

Timongcraft:
The default values were made extra *tame* because they will be automatically applied when updating Velocity also this is simply to avoid crashing Velocity rather than setting a limit so low that it might affect certain play styles (specifically, mods).

references:
https://github.com/PaperMC/Velocity/pull/1524#pullrequestreview-2705799351
https://github.com/PaperMC/Velocity/pull/1524#issuecomment-2743575944

SpigotRCE:
ok i fixed it

bridgelol:
Ideally you change the comment to say 25ms (10 commands per second) :) 25 is better to prevent false flags

SpigotRCE:
well if you want then

SpigotRCE:
u mean 100 right? or my math is not mathing

PR #1543, opened 2025-04-06
wip: config support for how many max channels a player can register

SpigotRCE:


Warriorrrr (commented):
This seems like something better suited as system property, all the different small changes you made also make it harder to find what's actually changed.

It's also normal "etiquette" when making a change to include *some* sort of message in the pr description, right now it seems you arbitrarily lowered the limit to 128 which seems more likely to interfere with mods.

SpigotRCE:
Alright, I'll do it by tomorrow cause I was on vacation.

PR #1552, opened 2025-04-14
Fix MinecraftChannelIdentifier parsing to align with vanilla

booky10:
As of 163a85a46834dbe5075c1e77fa838c71b57a59e2, Velocity will attempt to parse all custom payload channels registered/unregistered by clients and kick the client if it appears to be "invalid".
Because of this change, some minor differences between vanilla's `ResourceLocation` and Velocity's `MinecraftChannelIdentifier` caused players using certain mods to be kicked on login.

This PR fixes the parsing of `MinecraftChannelIdentifier` to match vanilla.

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/proxy/messages/MinecraftChannelIdentifier.java
+++ b/api/src/main/java/com/velocitypowered/api/proxy/messages/MinecraftChannelIdentifier.java
@@ -11,7 +11,6 @@ import static com.google.common.base.Preconditions.checkArgument;
 
 import com.google.common.base.Strings;
 import java.util.Objects;
-import java.util.regex.Pattern;
 import net.kyori.adventure.key.Key;
 import org.checkerframework.checker.nullness.qual.Nullable;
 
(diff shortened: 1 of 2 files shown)


electronicboy:
Thanks, but, I'm generally somewhat iffy on the nature of allowing purely namespace keys, though, I did forget to potentially add skipping of empty strings to mitigate some forge oddities on server switching

booky10:
> Thanks, but, I'm generally somewhat iffy on the nature of allowing purely namespace keys, though, I did forget to potentially add skipping of empty strings to mitigate some forge oddities on server switching

I think the issue you are referring to is [forge always appending an extra null byte to their channel-register packet](https://github.com/MinecraftForge/MinecraftForge/blob/207c1997c95787c7d43a1e5a4675cad410c72369/src/main/java/net/minecraftforge/network/ChannelListManager.java#L124)

Ignoring empty channel names in plugin message packets sounds good, but I would still argue that Velocity should try to keep the same logic as vanilla and allow parsing empty strings as `MinecraftChannelIdentifier` for as long as vanilla allows this too

astei (approved):

PR #1576, opened 2025-05-20
Update some dependencies

Toffikk:
Updated ASM to 9.8 to support java 25, terminal dependencies and netty to 4.2.1 for bugfixes.
Held back on reverting to the default adaptive alloc since the upcoming netty 4.2.2 contains important fixes for it; this pr shouldn't break any API

edit: also updated slf4j

if there are any suggestions as to other deps that could be updated feel free to comment

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -2,8 +2,8 @@
 configurate3 = "3.7.3"
 configurate4 = "4.1.2"
 flare = "2.0.1"
-log4j = "2.24.1"
-netty = "4.2.0.Final"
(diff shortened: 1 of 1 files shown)


electronicboy:
We generally want l4j to stay inline with paper

Toffikk:
@electronicboy so would it be better to pr to paper to update l4j too or just revert that change?

Toffikk:
Opened a pr on the paper repo to bump log4j and other terminal deps

electronicboy (approved):

PR #1578, opened 2025-05-25
Fix HoverEvent.showEntity() in protocol versions prior to 1.21.5

### LINKED ISSUE #1577: Client unexpectedly disconnect due to Component with HoverEvent.showEntity
### Expected Behavior

1. Create a Component with a `HoverEvent.showEntity()`.
2. send it to a player connected to the proxy by `player.sendMessage()`.
3. The client of the player show the component in the chat correctly with the hover event.

### Actual Behavior

Immediately after the message is sent. The client unexpectedly disconnected with a reason of internal exception occurred: `JsonSyntaxException: Missing type, expected to find a string`

### Steps to Reproduce

```
Optional<Player> optional = proxyServer.getPlayer(...);
Component message = Component.text(...).hoverEvent(HoverEvent.showEntity(...));
optional.ifPresent(player -> player.sendMessage(message));
```

### Plugin List

Only my plugin on the velocity and no plugins on the backend server.

### Velocity Version

```
> velocity info
[22:50:17 INFO]: Velocity 3.4.0-SNAPSHOT (git-5eb83760-b506)
[22:50:17 INFO]: Copyright 2018-2023 Velocity Contributors。Velocity 以 GNU 通用公共许可证第三版授权。
[22:50:17 INFO]: PaperMC - GitHub
```

### Additional Information

The **client** minecraft version is `1.20.1`. Possible reason of this bug may be the PR #1564 that introduces support for new hover event schema in 1.21.5 controlled by `JSONOptions.EMIT_HOVER_SHOW_ENTITY_KEY_AS_TYPE_AND_UUID_AS_ID`. This option is **FALSE** by default which means it has to be explicitly set to true in the `OptionSchema`. However in the `com.velocitypowered.proxy.protocol.ProtocolUtils` only `MODERN_SERIALIZER` _set that option to false explicitly_ , leaving all other serializers use the default value false, which may cause all versions prior to 1.21.5 to encounter same problem because the `type` key is missing.

Argon4W:
This PR fixes #1577

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/protocol/ProtocolUtils.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/protocol/ProtocolUtils.java
@@ -81,6 +81,7 @@ public enum ProtocolUtils {
               .value(JSONOptions.EMIT_RGB, Boolean.TRUE)
               .value(JSONOptions.EMIT_HOVER_EVENT_TYPE, JSONOptions.HoverEventValueMode.CAMEL_CASE)
               .value(JSONOptions.EMIT_CLICK_EVENT_TYPE, JSONOptions.ClickEventValueMode.CAMEL_CASE)
+              .value(JSONOptions.EMIT_HOVER_SHOW_ENTITY_KEY_AS_TYPE_AND_UUID_AS_ID, true)
               // before 1.20.3
               .value(JSONOptions.EMIT_COMPACT_TEXT_COMPONENT, Boolean.FALSE)
               .value(JSONOptions.EMIT_HOVER_SHOW_ENTITY_ID_AS_INT_ARRAY, Boolean.FALSE)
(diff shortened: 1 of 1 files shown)


proxy/src/main/java/com/velocitypowered/proxy/protocol/ProtocolUtils.java
```diff
@@ -97,6 +98,7 @@ public enum ProtocolUtils {
               .value(JSONOptions.EMIT_RGB, Boolean.TRUE)
               .value(JSONOptions.EMIT_HOVER_EVENT_TYPE, JSONOptions.HoverEventValueMode.CAMEL_CASE)
               .value(JSONOptions.EMIT_CLICK_EVENT_TYPE, JSONOptions.ClickEventValueMode.CAMEL_CASE)
+              .value(JSONOptions.EMIT_HOVER_SHOW_ENTITY_KEY_AS_TYPE_AND_UUID_AS_ID, true)
```
electronicboy, line 101: I believe that both of these should belong in the "after 1.20.3" section for documentation reasons, at least, that's where it was in the past

electronicboy (changes requested):

proxy/src/main/java/com/velocitypowered/proxy/protocol/ProtocolUtils.java
Argon4W: What do you mean by "both", the JSONOptions.EMIT_HOVER_SHOW_ENTITY_KEY_AS_TYPE_AND_UUID_AS_ID in the PRE_1_20_3_SERIALIZER and PRE_1_21_5_SERIALIZER should belong to "after 1.20.3"? But PRE_1_20_3 means prior to 1.20.3, I think. And "show_entity" indeed uses "type" and "id" after 1.16 instead of after 1.20.3

electronicboy (approved):

PR #1580, opened 2025-05-29
1.21.6

Gerrygames:


5jd:
Hi, i was wondering where are you finding the new mappings and methods?
I manually checked mappings earlier for the 1.21.6 snapshot and compared them with 1.21.5 but there was only one new mapping iirc, so it's definitely wrong.
I would like to have a little guidance here because i don't really know where to look since i couldn't find any documentation regarding it.
Thanks!

electronicboy:
mappings? you mean the protocol IDs? They added a packet early into the registry, that offsets the IDs of every packet after that

5jd:
> mappings? you mean the protocol IDs? They added a packet early into the registry, that offsets the IDs of every packet after that

I've been basically looking through mappings (comparing 1.21.5 to 25w15a) to look for protocol ids changes and add missing ids or changes in offsets into StateRegistry and other classes that mention offsets, although i overlooked the fact that an early new packet could influence all offsets coming after it.

proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/AvailableCommandsPacket.java
```diff
@@ -65,6 +65,7 @@ public class AvailableCommandsPacket implements MinecraftPacket {
   private static final byte FLAG_EXECUTABLE = 0x04;
   private static final byte FLAG_IS_REDIRECT = 0x08;
   private static final byte FLAG_HAS_SUGGESTIONS = 0x10;
+  private static final byte FLAG_IS_RESTRICTED = 0x20;
```
Owen1212055, line 70: This may be out of scope for the PR, but it would be really nice if this was exposed in some kind of way. For example, an api type that wraps around a predicate that indicates the command should be restricted.

electronicboy (approved):

electronicboy:
Tested against a modified rc1 client against a modified rc1 server, client was tested against VV 1.21.6, so, this seems to be good thus far, can probably merge this soon so people can get it out before release

Gerrygames:
Only thing missing should be support for the new text component features. But we can update adventure later when its out.

electronicboy:
Yea, that generally wouldn't block us here, more a "be nice to be done by the end of the week" sorta deal

PR #1582, opened 2025-05-30
Update VelocityServerConnection.java

RivenBytes:
Optimize handshake logic & micro-perf in VelocityServerConnection

- Use switch expression for forwarding mode selection in handshake instead of else-if ladder
- Cache protocolVersion variable to avoid redundant calls
- Use lambda for ChannelFutureListener instead of anonymous class in connect()
- Finalize local variables where possible

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/backend/VelocityServerConnection.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/backend/VelocityServerConnection.java
@@ -96,33 +96,23 @@ public class VelocityServerConnection implements MinecraftConnectionAssociation,
    */
   public CompletableFuture<Impl> connect() {
     CompletableFuture<Impl> result = new CompletableFuture<>();
-    // Note: we use the event loop for the connection the player is on. This reduces context
-    // switches.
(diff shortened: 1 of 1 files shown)


astei:
Yeah… no. This doesn’t look like a particularly useful PR (for states, the “micro-optimization” you note is a Jarvis that is likely to be inlined by the JIT and ultimately turned into a normal field access).

It’s not even a valuable “drive-by” contribution as you’ve removed several comments for no apparent reason. The only real value is using dine slightly more modern language features and while dated, the syntax used here is valid, just older.

RivenBytes:
> Yeah… no. This doesn’t look like a particularly useful PR (for states, the “micro-optimization” you note is a Jarvis that is likely to be inlined by the JIT and ultimately turned into a normal field access).
> 
> It’s not even a valuable “drive-by” contribution as you’ve removed several comments for no apparent reason. The only real value is using dine slightly more modern language features and while dated, the syntax used here is valid, just older.

Thank you for your feedback! I appreciate the time you took to review my PR.

My intention was to modernize the codebase slightly and improve readability by using newer language features. I understand that the performance gain might be negligible and that the current syntax is still valid.

PR #1583, opened 2025-05-30
Add custom disconnect reason to PlayerChooseInitialServerEvent

radek203:
Adds the `setReason(Component reason)` method to `PlayerChooseInitialServerEvent`, allowing plugins to set a custom disconnect message.

If not used, the default reason will be shown, ensuring full backward compatibility.

This is useful for plugin developers who want to display custom messages when no available servers are found for the player.


pingguomc (approved):

api/src/main/java/com/velocitypowered/api/event/player/PlayerChooseInitialServerEvent.java
```diff
@@ -54,6 +57,20 @@ public void setInitialServer(@Nullable RegisteredServer server) {
...
+  public Optional<Component> getReason() {
+    return Optional.ofNullable(reason);
+  }
+
+  /**
+   * Sets a custom disconnect reason for the player.
+   * Passing {@code null} will show the default reason.
+   *
+   * @param reason the disconnect reason to show to the player
+   */
```
4drian3d, line 69: It would be good to detail the functionality of the method, something like: “if a valid server is not provided as the initial server, this message will be used to disconnect the player; if it is not present or null is used, the default message will be used”

Emilxyz: Instead of allowing null here, maybe we just fire the event with the default message as initial reason?

electronicboy:
https://github.com/PaperMC/Velocity/pull/1787 - saw somebody reference this, is maybe more inlined with a better pattern for this sorta thing

PR #1586, opened 2025-06-05
Refactor some code

ghost:


proxy/src/main/java/com/velocitypowered/proxy/scheduler/VelocityScheduler.java
```diff
@@ -56,273 +56,325 @@
...
-    synchronized (tasksByPlugin) {
-      return Set.copyOf(tasks);
+    private final PluginManager pluginManager;
+    private final ScheduledExecutorService timerExecutionService;
+    
+    /**
+     * Tracks tasks by their plugin instance using concurrent collections.
+     * Uses synchronized IdentityHashSet per plugin for efficient task management.
+     */
+    private final ConcurrentHashMap<Object, Set<ScheduledTask>> tasksByPlugin = new ConcurrentHashMap<>();
```
astei, line 72: Just as an example of the slop here: this is not only the wrong code style, this is also going to silently do the wrong thing! `ConcurrentHashMap` does not use the identity hash code of the object.

NEZNAMY: This PR was AI generated. See their other PR's. Just block and ignore this user.

astei: I'm perfectly aware, that's why I called it slop.

PR #1589, opened 2025-06-11
Add "/velocity:help" command.

Mickey42302:
I would like to suggest the addition of a "/velocity:help" command. This can be used to list the syntax for commands that are built into Velocity, such as "/server" and "/velocity".


electronicboy:
namespaced commands are more of a means for server admins to be able to run/deal with overlapping commands, not something we should be considering a 1st class citizen when it comes to user facing commands; The built in commands are also generally simple enough and fairly already well served by the command system built into minecraft, I'm not really sure what this is solving, if we were to add a help system, it would need to be more dynamic than a bunch of hardcoded commands, but, I just don't see the desire for that

PR #1590, opened 2025-06-13
feat: flexible authentication

tommyskeff:
This pull request adds additional flexibility to the login and authentication sequence.

There are two main changes in this PR:
- adds support for offline-mode encryption in 1.20.5+
- decouples the Mojang authentication logic from the login session handler using an event

I have added a flag offlineEncryption to the PreLoginEvent, which defines whether the proxy will attempt to encrypt the connection (if in offline mode) once the event has finished dispatching. However, it is clearly documented that setting this flag will not always force encryption to be enabled. In 1.20.5, a flag was added to the encryption start packet for whether the client should authenticate with Mojang, which gives us the ability to encrypt a connection without forcing the client to authenticate with the session server. In versions below 1.20.5, we must skip encryption even if this flag is set. The default value for this flag is the value of the offline-encryption proxy configuration value (default false).

I have added the AuthAttemptEvent, which is fired just after the PreLoginEvent, regardless of whether the connection is online or offline mode. This event allows plugins to override the authentication mechanism used by the proxy, which is particularly useful in server networks where authentication is handled on a single microservice, or if a custom session server were to be used. There is a default authentication mechanism which listens for this event with a low priority and skips official Mojang auth if the connection is offline-mode, or another plugin has already set the result.

In order to make these changes, I have had to refactor some of the login sequence code. Due to the sensitive nature of this secure process, it is vital that this PR receives careful review before approval. I have thought carefully about the changes, and no login features have been removed, but I would appreciate some care when verifying that the process remains robust.

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/event/connection/AuthAttemptEvent.java
+++ b/api/src/main/java/com/velocitypowered/api/event/connection/AuthAttemptEvent.java
@@ -1,0 +1,127 @@
+package com.velocitypowered.api.event.connection;
+
+import com.velocitypowered.api.event.ResultedEvent;
+import com.velocitypowered.api.event.player.GameProfileRequestEvent;
+import com.velocitypowered.api.proxy.InboundConnection;
+import com.velocitypowered.api.util.GameProfile;
+import net.kyori.adventure.text.Component;
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+/**
+ * <p>This event is fired when a player attempts to authenticate with the proxy, regardless of whether
+ * the connection is online-mode or offline-mode. The proxy implementation will provide a default
+ * authentication mechanism (using the Mojang session server), but plugins can subscribe to this
+ * event to implement custom authentication logic.</p>
+ *
+ * <p>If the connection is {@linkplain AuthAttemptEvent#isOnlineMode() online-mode}, the
+ * {@linkplain AuthAttemptEvent#getSharedSecret() shared secret} will always be present, as encryption
+ * is always enabled in online-mode. However, in offline-mode, encryption may or may not be enabled
+ * (configurable in proxy config file), and thus the shared secret may be {@code null}.</p>
+ *
+ * <p>If the result of this event is {@code null}, online-mode connections will assume failure, but
+ * offline-mode connections will assume success. The default authentication mechanism will not
+ * provide a result for offline-mode connections.</p>
+ *
+ * <p>Note that a successful result from this event does not necessarily determine the player
+ * profile, as plugins may override this using {@link GameProfileRequestEvent}.</p>
+ */
+public final class AuthAttemptEvent implements ResultedEvent<AuthAttemptEvent.AuthResult> {
+
+  private final @NonNull InboundConnection connection;
+  private final boolean onlineMode;
+  private final byte @Nullable [] sharedSecret;
+  private final @NonNull String attemptedUsername;
+
+  private @Nullable AuthResult result;
+
+  /**
+   * Constructs a new authentication attempt event.
+   *
+   * @param connection the inbound connection attempting to authenticate
+   * @param onlineMode whether the connection is in online-mode
+   * @param sharedSecret the shared secret for online-mode connections, or {@code null} for offline-mode
+   * @param attemptedUsername the username that the player is attempting to authenticate with
+   */
+  public AuthAttemptEvent(@NonNull InboundConnection connection, boolean onlineMode,
+                          byte @Nullable [] sharedSecret, @NonNull String attemptedUsername) {
+    this.connection = connection;
+    this.onlineMode = onlineMode;
+    this.sharedSecret = sharedSecret;
+    this.attemptedUsername = attemptedUsername;
+  }
+
+  /**
+   * Gets the inbound connection that is attempting to authenticate.
+   *
+   * @return the inbound connection
+   */
+  public @NonNull InboundConnection getConnection() {
+    return connection;
+  }
+
+  /**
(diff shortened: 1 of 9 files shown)


api/src/main/java/com/velocitypowered/api/event/connection/AuthAttemptEvent.java
```diff
@@ -0,0 +1,143 @@
...
+   */
+  @Override
+  public void setResult(@Nullable AuthResult result) {
+    this.result = result;
+  }
+
+  /**
+   * Interface representing the result of an authentication attempt.
+   */
+  public sealed interface AuthResult extends Result permits SuccessResult, FailureResult {
```
Timongcraft, line 120: Why not use the [GenericResult](https://github.com/PaperMC/Velocity/blob/669fda298c670c55686f34d868383052b192518d/api/src/main/java/com/velocitypowered/api/event/ResultedEvent.java#L52)?

electronicboy:
Alt authentication flows are not something we desire to put maintenance into

electronicboy:
Somebody has pointed out that there might be some desire for this in terms of dealing with things like authentication caching; I'm not 100% happy with this sorta thing, just primarily due to the past uses of such mechanisms always being a headache in terms of the support burden with the things that use them. 

Default implementations of something so foundational being an event just sounds pointlessly fragile and would more suit a registered interface in which if something did register its own handler, it would override the default one (not that it couldn't  delegate to it for the means of providing fallbacks)

The nature of the support burden that such an API introduces would also require it to be pretty clear in the logs when somebody registers such a mechanism, really conflicted because while I understand the advantages of being able to manage this stuff, many usecases can already be solved by just overriding the servers that velocity uses for this stuff through the flags, as is already done by services like minehut, etc. Really not enthusiastic about this one.

tommyskeff:
I am disappointed, yet understand your concerns. 

 Whilst I understand what you say that you can replace the session server, placing that as the only mechanism to modify the authentication flow is needlessly limiting, especially when there are other legitimate fields which a user may want to include in their authentication request.

I understand what you say about potential fragility when using the event manager instead of a direct integration. It would potentially make sense to be more concrete. A clear log message upon custom auth registration would be acceptable.

This feature is non-negotiable for our server, and I imagine others will be in a similar position. I do not understand why fear of a support burden is a limiting factor in the flexibility of Velocity, because I was under the impression that a core design principle was the "unparalleled server support, scalability, and flexibility". This is particularly disappointing to hear.

tommyskeff:
Also, another reason why we need this feature which I omitted before. Mojang session server rate limit is 600 reqs per 10 mins, which averages out to ~1 req/s. Larger servers regularly exceed this rate, and the resolutions are pretty much limited to:
- spin up more velocity proxies with unique public ips
- use outbound http proxies to evade the rate limit

Since a core aim of Velocity is performance (thousands of players on one proxy), it seems unnecessary to allocate resources to more Velocity proxies than you need, just to get around the auth rate limits. A server with 1000 concurrent players should be fine on one Velocity instance, yet could easily exceed the rate limit.

Using outbound proxies is therefore the goto option, and I am aware that Hypixel, for example, modifies the BungeeCord login sequence to make this work. It would be nice to not have to get hacky with Velocity if a server wants to achieve the same thing. Speaking of Hypixel, I know they have a custom mechanism for authenticating players when Mojang auth servers are offline (some sort of client fingerprinting I believe), and this is also something I think other servers should be able to achieve.

These use-cases may seem niche, and they are, but it seems overally restricting to disallow this behaviour in favour of discouraging piracy (which is already possible through every proxy via offline mode).

electronicboy:
The entire premise of a PR is handing over code for somebody else to support and maintain, most servers already doing this just maintain their own forks of the software.

It's not just the nature of piracy, but more so the nature of allowing plugins to silently interfere with a core mechanism of the proxy, especially given how much of a pain it is to provide support as-is, given how little is logged from Mojang's endpoints. In an ideal world, the nature of "how much of a headache is this going to cause in our support channels" would not be a factor in terms of maintaining software, but, sadly, as humans with limited spare time, it is.

tommyskeff:
I understand that you and other maintainers are the ones who ultimately have to be responsible for supporting and maintaining the software. I was hoping that the benefits of this flexibility would be recognised, but as someone with infinitely more knowledge of the framework, I respect your opinion on that. Regardless, I appreciate your efforts to understand and process my request.

You are of course correct that we could just maintain a fork of Velocity for our use case. 

Also, did you have thoughts on offline-mode encryption? This is something Mojang intended to support when they made the change in 1.20.5.

electronicboy:
not sure what the question on offline-mode encryption is, it's not something that I'd ever bother to invest in on myself, but, if it comes as a free side-effect, whatever

tommyskeff:
image

They added the `shouldAuthenticate` flag to extend the "benefits" of encryption to offline-mode connections too (I am aware that Mojang's idea of encryption isn't exactly the gold standard).

tommyskeff:
I'm not going to pursue this PR any further, but if offline-mode encryption is something you think worth having on Velocity, the changes are there.

masmc05:
I doubt calling it "offline-mode encryption" is really correct, as the client will start considering the connection as being made with an online mode server, even if it skips the authentication. You can even see in the screenshot bellow that ignoring the skipped authentication, it'll go trough the same path as an online mode client would do
image
It was added together with transfer packets and cookies most probably with the scope of servers to detect successful past authentication with encrypted/signed cookies (so if someone goes from Germany proxy to an US proxy the server could skip the authentication).
And my personal opinion is that this shouldn't be exposed in config like in this pr and only to the api.

tommyskeff:
@masmc05 You raise a good point. More research would need to be done into other side effects of encrypting the connection.

PR #1598, opened 2025-06-26
1.21.7 Support

jonesdevelopment:
https://minecraft.wiki/w/Java_Edition_1.21.7

There don't seem to be any changes in the protocol except for the protocol version ID.

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/network/ProtocolVersion.java
+++ b/api/src/main/java/com/velocitypowered/api/network/ProtocolVersion.java
@@ -92,7 +92,7 @@ public enum ProtocolVersion implements Ordered<ProtocolVersion> {
   MINECRAFT_1_21_4(769, "1.21.4"),
   MINECRAFT_1_21_5(770, "1.21.5"),
   MINECRAFT_1_21_6(771, "1.21.6"),
-  MINECRAFT_1_21_7(772, 258, "1.21.7");
+  MINECRAFT_1_21_7(772, "1.21.7");
 
   private static final int SNAPSHOT_BIT = 30;
 
--- a/api/src/main/java/com/velocitypowered/api/network/ProtocolVersion.java
+++ b/api/src/main/java/com/velocitypowered/api/network/ProtocolVersion.java
@@ -91,7 +91,8 @@ public enum ProtocolVersion implements Ordered<ProtocolVersion> {
   MINECRAFT_1_21_2(768, "1.21.2", "1.21.3"),
   MINECRAFT_1_21_4(769, "1.21.4"),
   MINECRAFT_1_21_5(770, "1.21.5"),
-  MINECRAFT_1_21_6(771, "1.21.6");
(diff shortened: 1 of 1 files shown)


api/src/main/java/com/velocitypowered/api/network/ProtocolVersion.java
```diff
@@ -91,7 +91,8 @@ public boolean isSupported() {
   MINECRAFT_1_21_2(768, "1.21.2", "1.21.3"),
   MINECRAFT_1_21_4(769, "1.21.4"),
   MINECRAFT_1_21_5(770, "1.21.5"),
-  MINECRAFT_1_21_6(771, "1.21.6");
+  MINECRAFT_1_21_6(771, "1.21.6"),
+  MINECRAFT_1_21_7(-1, 258, "1.21.7");
```
kennytv: You should already set the release protocol version, 772

jonesdevelopment: > You should already set the release protocol version, 772

I've tried doing that, but it wouldn't let me connect to the server. I've seen `-1` being used in previous PR's. I can change it, sure, but I don't think it'll let me connect, then.

I believe the issue is caused by `getProtocol()` always returning the release protocol instead of the snapshot protocol if a release protocol is provided. Is this intended?

https://github.com/PaperMC/Velocity/blob/dev/3.0.0/api/src/main/java/com/velocitypowered/api/network/ProtocolVersion.java#L176

electronicboy: I would probably flip that check so that it checks for the presense of a snapshot version instead of relying on the release version being unset; but, I would probably just add the release protocol and then the snapshot one after that for the sake of continuity, people can then throw these builds out ready for toying with and, if mojang makes no last minute changes, they already have support (ofc, this gets funky if they do sneak something in, but, I generally don't expect them at this phase)

jonesdevelopment: I'm going to set the release version to 772, then. If someone wants to use Velocity for 1.21.7-rc1/rc2, they can download previous builds through GitHub Actions.

api/src/main/java/com/velocitypowered/api/network/ProtocolVersion.java
```diff
@@ -91,7 +91,8 @@ public boolean isSupported() {
   MINECRAFT_1_21_2(768, "1.21.2", "1.21.3"),
   MINECRAFT_1_21_4(769, "1.21.4"),
   MINECRAFT_1_21_5(770, "1.21.5"),
-  MINECRAFT_1_21_6(771, "1.21.6");
+  MINECRAFT_1_21_6(771, "1.21.6"),
+  MINECRAFT_1_21_7(772, 258, "1.21.7");
```
kennytv:
```suggestion
  MINECRAFT_1_21_7(772, "1.21.7");
```

kennytv (approved):

PR #1608, opened 2025-07-07
Login Dialog and Custom Click Action support

Ocelot5836:
Adds support for sending dialogs to clients in the configuration phase.


electronicboy:
We can't really just implement this as a compound tag when we hope to actually expose the ability to expose this stuff in the API as soon as is viable, we need to support reading/writing this back from an actual data structure that can be created from within the API; and there is the behavioral difference between configuration and play which also needs handling within this packet

Issue #1591, opened 2025-06-15
Unable to join Proxy since Dependency Update


CronixZero:
### Expected Behavior

One should be able to join the server

### Actual Behavior

Everything works fine on Build 510 but once I update the proxy the Minecraft screen gets stuck at "Encrypting..." and the connections closes (Disconnected). The only thing the console then says is:

>....
[10:15:32 INFO]: [connected player] CronixZero (/<redacted>:63778) has connected
>....
[10:16:01 ERROR]: [connected player] CronixZero (/<redacted>:63778): read timed out
>....
[10:16:01 INFO]: [connected player] CronixZero (/<redacted>:63778) has disconnected

### Steps to Reproduce

1. Use itzg proxy docker image
2. Use the latest build
3. Join the proxy
4. Connection should get stuck at "Encrypting..."

### Plugin List

/velocity plugins
[10:18:26 INFO]: Plugins: velocity

/plugins
[10:18:53 INFO]: ℹ Server Plugins (0):

### Velocity Version

```
[10:19:24 INFO]: Velocity 3.4.0-SNAPSHOT (git-549b8d62-b511)
[10:19:24 INFO]: Copyright 2018-2025 Velocity Contributors. Velocity is licensed under the terms of the GNU General Public License v3.
[10:19:24 INFO]: PaperMC - GitHub
```

^ Issue persists on 512 but I downgraded the proxy to check when it started

derklaro:
This happens due to netty constructing ByteBufs backed by MemorySegments since 4.2.2. The compression code then passes a ByteBuffer (also backed by the MemorySegment) to `Inflater.inflate(ByteBuffer)` (when using java compressors, same thing applies for Deflaters) which fails due to a bug in the jdk: https://bugs.openjdk.org/browse/JDK-8357145. The netty change was reverted recently (https://github.com/netty/netty/commit/a919dd3a53a50da080dc925a6fb085c81cab6294) to only use MemorySegments on java 25+. This will be part of netty 4.2.3. For now the best bet would be to downgrade to netty 4.2.1 again (or, as a quick fix [which might not be possible for everyone], to not use java 24).

PR #1615, opened 2025-07-18
update to netty 4.2.3.Final

derklaro:
Netty 4.2.3 fixes the issues introduced in 4.2.2 (#1591) and also includes a huge improvement to the adaptive allocator. Due to this I think it should be given another chance instead of using the pooled allocator by default.

The netty changelog can be found here: https://netty.io/news/2025/07/15/4-2-3.html

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -3,7 +3,7 @@ configurate3 = "3.7.3"
 configurate4 = "4.1.2"
 flare = "2.0.1"
 log4j = "2.24.3"
-netty = "4.2.1.Final"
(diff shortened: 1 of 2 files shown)


derklaro:
There netty team did some allocation pattern simulations that show the improvements (note: the blue line (pooled 128) is hidden behind the yellow line (pooled 512)):
4.2.2:
image

4.2.3:
image

electronicboy:
> Due to this I think it should be given another chance instead of using the pooled allocator by default.

I'm currently going through a weird phase and don't have the capacity to deal with the support burden if it is still broken

PR #1621, opened 2025-08-02
feat: map show_dialog & clear_dialog in CONFIG state

### LINKED ISSUE #1618: Dialogs sent during configuration phase do not show on client when Velocity is between client and backend server
### Expected Behavior

A Paper server behind Velocity, should be able to show a dialog to a player when their connection is in the configuration phase.

Attached is a video of Paper `1.21.8-17-main@dbc367b (2025-07-26T21:53:33Z)`

https://github.com/user-attachments/assets/63e2f485-f8fe-4e9d-aa6a-f7baa14119f9

### Actual Behavior

Instead the client stays on the "Joining world..." screen until the configuration phase ends.

Attached is video of this with Velocity infront of Paper.

https://github.com/user-attachments/assets/3029d533-73a2-42f2-afcd-0fa4ad18ec9b

### Steps to Reproduce

Use a plugin with a `AsyncPlayerConnectionConfigureEvent` to send a dialog during connection, and pause the event to simulate loading / waiting on player acceptance.

An example plugin is provided below that generates 5 countdown dialogs at 1 second intervals before allowing the configuration phase to end.

```java
package com.tehbeard.velocitydialog;

import io.papermc.paper.connection.PlayerConfigurationConnection;
import io.papermc.paper.dialog.Dialog;
import io.papermc.paper.event.connection.configuration.AsyncPlayerConnectionConfigureEvent;
import io.papermc.paper.registry.data.dialog.DialogBase;
import io.papermc.paper.registry.data.dialog.body.DialogBody;
import io.papermc.paper.registry.data.dialog.type.DialogType;
import org.bukkit.Bukkit;
import org.bukkit.event.EventHandler;
import org.bukkit.event.Listener;
import org.bukkit.plugin.java.JavaPlugin;

import java.util.List;

public final class VelocityDialogTest extends JavaPlugin implements Listener {


    @Override
    public void onEnable() {
        Bukkit.getPluginManager().registerEvents(this, this);
    }

    @Override
    public void onDisable() {
        // Plugin shutdown logic
    }

    private Dialog countdown(int count) {
        return Dialog.create(builder -> {
            builder.empty().type(DialogType.notice()).base(DialogBase.create(
                Component.text("Loading"),
                Component.text("Loading"),
                false,
                false,
                DialogBase.DialogAfterAction.WAIT_FOR_RESPONSE,
                List.of(
                    DialogBody.plainMessage(Component.text("Loading... ").append(Component.text(count)))
                ),
                List.of()));
        });
    }

   @EventHandler
    public void onConnect(AsyncPlayerConnectionConfigureEvent event) {
        getLogger().info("connValidate: " + event.getConnection().getClass().getName());
        var conn = event.getConnection();

        for (var i = 5; i > 0; i--) {
            conn.getAudience().showDialog(countdown(i));
            getLogger().info("Countdown " + i);
            try {
                Thread.sleep(1000);
            } catch (InterruptedException ex) {

            }
        }
    }
}

```

### Plugin List

**Backend plugins**
```
> plugins
[20:51:26 INFO]: ℹ Server Plugins (1):
[20:51:26 INFO]: Paper Plugins:
[20:51:26 INFO]:  - shard-remote // Test plugin used, stripped down to just the listener given in the example above.
```

**Velocity plugins**
```
> velocity plugins
[20:52:11 INFO]: Plugins: velocity
```

### Velocity Version

Velocity 3.4.0-SNAPSHOT (git-e9940713-b522)

### Additional Information

Paper version used: `1.21.8-17-main@dbc367b (2025-07-26T21:53:33Z)`

Emilxyz:
Maps `show_dialog` & `clear_dialog` during CONFIG state, so that they don't get caught in PlayPacketQueueInboundHandler.

I also included en- & decoding of the PLAY version of `show_dialog`, but decided against mapping it in StateRegistry since it is not needed right now. Maybe in the future when a dialog api is added it will be though, so then it's there (it will probably be changed then anyways, but whatever).

Fixes #1618 

Using the code provided in the linked issue (with the change of sending a clear if i == 3) now results in the expected behavior:

https://github.com/user-attachments/assets/9471bd85-f2ac-4894-b478-69587a3c217f

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/MinecraftSessionHandler.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/MinecraftSessionHandler.java
@@ -24,6 +24,8 @@ import com.velocitypowered.proxy.protocol.packet.BundleDelimiterPacket;
 import com.velocitypowered.proxy.protocol.packet.ClientSettingsPacket;
 import com.velocitypowered.proxy.protocol.packet.ClientboundCookieRequestPacket;
 import com.velocitypowered.proxy.protocol.packet.ClientboundStoreCookiePacket;
+import com.velocitypowered.proxy.protocol.packet.DialogClearPacket;
+import com.velocitypowered.proxy.protocol.packet.DialogShowPacket;
 import com.velocitypowered.proxy.protocol.packet.DisconnectPacket;
 import com.velocitypowered.proxy.protocol.packet.EncryptionRequestPacket;
 import com.velocitypowered.proxy.protocol.packet.EncryptionResponsePacket;
@@ -362,6 +364,14 @@ public interface MinecraftSessionHandler {
   }
 
   default boolean handle(ClientboundServerLinksPacket packet) {
+    return false;
+  }
+
+  default boolean handle(DialogClearPacket packet) {
+    return false;
+  }
+
+  default boolean handle(DialogShowPacket packet) {
     return false;
   }
 }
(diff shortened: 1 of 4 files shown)


electronicboy (approved):
Not really fond of the fact that none of this is being parsed, but, I guess that will have to wait until we get to the point we can actually start working on API for this

Emilxyz:
Yeah, I agree. This is supposed to be the very minimum to fix the issue and I expect this to be replaced once there is an api for dialogs. 

It kind of sucks that we have to decode and encode packets that we don't want to do anything with, just to allow them through the PlayPacketQueueInboundHandler. Maybe there is a better fix for this, but I couldn't really think of anything nice.

PR #1626, opened 2025-08-09
feat: add ProxyPreShutdownEvent before players are disconnected

DartCZ:
This change modifies the proxy shutdown sequence so that all connected players are not immediately disconnected.
Instead, the shutdown process now:

1. Stops accepting new connections (cm.shutdown()).
2. Fires ProxyPreShutdownEvent and waits (up to 10 seconds) for all event handlers to complete.

- This allows plugins to transfer players to another proxy or perform other shutdown logic before disconnection.

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/event/proxy/ProxyPreShutdownEvent.java
+++ b/api/src/main/java/com/velocitypowered/api/event/proxy/ProxyPreShutdownEvent.java
@@ -16,9 +16,11 @@ import com.velocitypowered.api.event.annotation.AwaitingEvent;
  * This is the last point at which you can interact with currently connected players,
  * for example to transfer them to another proxy or perform other cleanup tasks.
  *
- * <p><b>Note:</b> Velocity will wait for all event listeners to complete before disconnecting players,
- * but note that the event will time out after 10 seconds
- * to prevent shutdown from hanging indefinitely.</p>
+ * @implNote Velocity will wait for all event listeners to complete before disconnecting players,
+ *     but note that the event will time out after the configured value of the
+ *     <code>velocity.pre-shutdown-timeout</code> system property, default 10 seconds,
+ *     in seconds to prevent shutdown from hanging indefinitely
+ * @since 3.4.0
  */
 @Beta
 @AwaitingEvent
--- a/api/src/main/java/com/velocitypowered/api/event/proxy/ProxyPreShutdownEvent.java
+++ b/api/src/main/java/com/velocitypowered/api/event/proxy/ProxyPreShutdownEvent.java
@@ -1,8 +1,18 @@
 /*
- * Copyright (C) 2018-2021 Velocity Contributors
+ * Copyright (C) 2018-2025 Velocity Contributors
  *
- * The Velocity API is licensed under the terms of the MIT License. For more details,
- * reference the LICENSE file in the api top-level directory.
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
 
 package com.velocitypowered.api.event.proxy;
@@ -12,9 +22,13 @@ import com.velocitypowered.api.event.annotation.AwaitingEvent;
 /**
  * This event is fired by the proxy after it has stopped accepting new connections,
  * but before players are disconnected.
+ * <p>
  * This is the last point at which you can interact with currently connected players,
  * for example to transfer them to another proxy or perform other cleanup tasks.
- * Velocity will wait for all event listeners to complete before disconnecting players.
+ * <p>
+ * Velocity will wait for all event listeners to complete before disconnecting players,
+ * but note that the event will time out after a certain period (currently 10 seconds)
+ * to prevent shutdown from hanging indefinitely.
  */
 @AwaitingEvent
 public final class ProxyPreShutdownEvent {
--- a/api/src/main/java/com/velocitypowered/api/event/proxy/ProxyPreShutdownEvent.java
+++ b/api/src/main/java/com/velocitypowered/api/event/proxy/ProxyPreShutdownEvent.java
@@ -1,35 +1,26 @@
 /*
  * Copyright (C) 2018-2025 Velocity Contributors
  *
- * This program is free software: you can redistribute it and/or modify
- * it under the terms of the GNU General Public License as published by
- * the Free Software Foundation, either version 3 of the License, or
- * (at your option) any later version.
- *
- * This program is distributed in the hope that it will be useful,
- * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
- * GNU General Public License for more details.
- *
- * You should have received a copy of the GNU General Public License
- * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ * The Velocity API is licensed under the terms of the MIT License. For more details,
+ * reference the LICENSE file in the api top-level directory.
  */
 
 package com.velocitypowered.api.event.proxy;
 
+import com.google.common.annotations.Beta;
 import com.velocitypowered.api.event.annotation.AwaitingEvent;
 
 /**
  * This event is fired by the proxy after it has stopped accepting new connections,
  * but before players are disconnected.
- * <p>
  * This is the last point at which you can interact with currently connected players,
  * for example to transfer them to another proxy or perform other cleanup tasks.
- * <p>
- * Velocity will wait for all event listeners to complete before disconnecting players,
+ *
+ * <p><b>Note:</b> Velocity will wait for all event listeners to complete before disconnecting players,
  * but note that the event will time out after a certain period (currently 10 seconds)
- * to prevent shutdown from hanging indefinitely.
+ * to prevent shutdown from hanging indefinitely.</p>
  */
+@Beta
 @AwaitingEvent
 public final class ProxyPreShutdownEvent {
 
--- a/api/src/main/java/com/velocitypowered/api/event/proxy/ProxyPreShutdownEvent.java
+++ b/api/src/main/java/com/velocitypowered/api/event/proxy/ProxyPreShutdownEvent.java
@@ -1,0 +1,26 @@
+/*
+ * Copyright (C) 2018-2021 Velocity Contributors
+ *
+ * The Velocity API is licensed under the terms of the MIT License. For more details,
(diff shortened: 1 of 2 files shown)


Timongcraft:
I'd rather propose a ProxyPreShutdownEvent rather than changing existing behaviour, which can't really be done without a version bump.
Also AFAIK the ProxyPreShutdownEvent timeouting is also breaking.

Speedy11CZ:
I agree that it would be better to have a new ProxyPreShutdownEvent than to break the current ProxyShutdownEvent. Otherwise, I consider this functionality useful, the ability to transfer during restart, for example, to a temporary Limbo server or to another Velocity instance for high availability, and I would personally use it.

electronicboy:
Having players hanging around while stuff is expecting to destroy itself is just bad practice, we already have enough issues due to this in Bukkit land, I'd rather not blindly export that issue over here; The notion of an earlier event would be the better solution as it would allow plugins to do stuff to players before everything starts going godzilla

Timongcraft:
You should note in the javadoc that the event will time out.

Some minor nitpicks:
IMO, you are logging a bit too much. Nothing else has such specific info logs, for example, and I don't think they are needed.
You could also bump the year in the copyright notice.

DartCZ:
Hello @Timongcraft, thanks for the feedback! All requested changes have been implemented, please take a look when you have a moment 🤟

Timongcraft:
Alright, you did however use the wrong license, as the api module is licensed under MIT:
https://github.com/PaperMC/Velocity/blob/dev/3.0.0/api/LICENSE
Also, try building the PR, as this should fail the checkstyle because the paragraph tag is supposed to be directly in front of the block. (And I personally would put "Note:" in front)

Timongcraft:
Ah yea and forgot new api is typically annotated with Google's Beta annotation

proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
```diff
@@ -578,6 +579,22 @@ public void shutdown(boolean explicitExit, Component reason) {
...
+      try {
+        eventManager.fire(new ProxyPreShutdownEvent())
+                .toCompletableFuture()
+                .get(shutdownTimeout, TimeUnit.SECONDS);
+      } catch (TimeoutException te) {
+        logger.warn("ProxyPreShutdownEvent timed out after {}s; continuing shutdown.",
+                shutdownTimeout);
+      } catch (ExecutionException ee) {
+        logger.error("Exception in ProxyPreShutdownEvent handler; continuing shutdown.", ee);
+      } catch (InterruptedException ie) {
```
Timongcraft:
```suggestion
      } catch (InterruptedException ignored) {
```

DartCZ:
Thanks! The wrong license was copied from existing non-API code in Velocity by mistake. I’ve now updated it to MIT for the API module, fixed the Javadoc `<p>` placement, added the bold “Note:” prefix, and annotated the event with Google's Beta annotation.

proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
```diff
@@ -578,6 +579,22 @@ public void shutdown(boolean explicitExit, Component reason) {
       // done first to refuse new connections
       cm.shutdown();
 
+      int shutdownTimeout = 10;
+
+      try {
+        eventManager.fire(new ProxyPreShutdownEvent())
+                .toCompletableFuture()
+                .get(shutdownTimeout, TimeUnit.SECONDS);
+      } catch (TimeoutException te) {
```
Timongcraft:
```suggestion
      } catch (TimeoutException ignored) {
```

Timongcraft:
Maybe also change the timeout message to match the formatting of the disconnect event timeout message:
`Your plugins took over 10 seconds to shut down.` and change it to include that it was while disconnecting the players.

Timongcraft:
No, sorry that it was confusing I meant the change the other message, from the players. This message is the pre shutdown event and not the disconnect events

DartCZ:
Gotcha I misunderstood that part. If you don’t mind, could you drop a suggestion with the exact wording you’d like for that other message? That way I can make sure it’s exactly how you intended.

proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
```diff
@@ -578,6 +579,22 @@ public void shutdown(boolean explicitExit, Component reason) {
...
       cm.shutdown();
 
+      int shutdownTimeout = 10;
+
+      try {
+        eventManager.fire(new ProxyPreShutdownEvent())
+                .toCompletableFuture()
+                .get(shutdownTimeout, TimeUnit.SECONDS);
+      } catch (TimeoutException ignored) {
+        logger.warn("Your plugins took over {} seconds to shut down while disconnecting players.",
```
Timongcraft: Ig something like this?
```suggestion
        logger.warn("Your plugins took over {} seconds during pre shut down.",
```

Timongcraft:
You could also use the opportunity to document this timeouting behaviour in the DisconnectEvent

Timongcraft (approved):

proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
```diff
@@ -578,6 +579,22 @@ public void shutdown(boolean explicitExit, Component reason) {
...
       cm.shutdown();
 
+      int shutdownTimeout = 10;
+
+      try {
+        eventManager.fire(new ProxyPreShutdownEvent())
+                .toCompletableFuture()
+                .get(shutdownTimeout, TimeUnit.SECONDS);
+      } catch (TimeoutException ignored) {
+        logger.warn("Your plugins took over {} seconds during pre shut down.",
```
powercasgamer:
```suggestion
        logger.warn("Your plugins took over {} seconds during pre shutdown.",
```

powercasgamer (changes requested):

api/src/main/java/com/velocitypowered/api/event/proxy/ProxyPreShutdownEvent.java
```diff
@@ -0,0 +1,31 @@
...
+import com.velocitypowered.api.event.annotation.AwaitingEvent;
+
+/**
+ * This event is fired by the proxy after it has stopped accepting new connections,
+ * but before players are disconnected.
+ * This is the last point at which you can interact with currently connected players,
+ * for example to transfer them to another proxy or perform other cleanup tasks.
+ *
+ * <p><b>Note:</b> Velocity will wait for all event listeners to complete before disconnecting players,
+ * but note that the event will time out after a certain period (currently 10 seconds)
```
powercasgamer:
```suggestion
 * but note that the event will time out after 10 seconds
```
This should be rephrased to something like this imo

powercasgamer (dismissed):

FrederikHeinrich:
Thanks for this change! The current 10s timeout feels too short — in my use case I need to wait until the player is fully ready before disconnecting. Could you make the timeout configurable (e.g. in velocity.toml) and also allow -1 for no timeout at all? A slightly higher default (20–30s) would also help.

Speedy11CZ:
> Thanks for this change! The current 10s timeout feels too short — in my use case I need to wait until the player is fully ready before disconnecting. Could you make the timeout configurable (e.g. in velocity.toml) and also allow -1 for no timeout at all? A slightly higher default (20–30s) would also help.

I personaly don't think this is a approach. 10 seconds should be sufficient for all plugins; a longer time may indicate a deadlock. I consider using properties to be a better option than config files, as it reduces the risk of the user breaking Velocity by not knowing what it does in the background; it is therefore more of a specific fine-tuning.

FrederikHeinrich:
> > Thanks for this change! The current 10s timeout feels too short — in my use case I need to wait until the player is fully ready before disconnecting. Could you make the timeout configurable (e.g. in velocity.toml) and also allow -1 for no timeout at all? A slightly higher default (20–30s) would also help.
> 
> I personaly don't think this is a approach. 10 seconds should be sufficient for all plugins; a longer time may indicate a deadlock. I consider using properties to be a better option than config files, as it reduces the risk of the user breaking Velocity by not knowing what it does in the background; it is therefore more of a specific fine-tuning.

When restarting the proxy, I would simply tell the game server to reconnect the user as soon as possible, and if they are currently playing a mini-game, for example, they should be allowed to finish it.

4drian3d:
> > Thanks for this change! The current 10s timeout feels too short — in my use case I need to wait until the player is fully ready before disconnecting. Could you make the timeout configurable (e.g. in velocity.toml) and also allow -1 for no timeout at all? A slightly higher default (20–30s) would also help.
> 
> I personaly don't think this is a approach. 10 seconds should be sufficient for all plugins; a longer time may indicate a deadlock. I consider using properties to be a better option than config files, as it reduces the risk of the user breaking Velocity by not knowing what it does in the background; it is therefore more of a specific fine-tuning.

I think that a way to configure the timeout should be implemented. Some people will want it to be 5 seconds, others 15 seconds, but it **must** be implemented as a system property

DartCZ:
> > > Thanks for this change! The current 10s timeout feels too short — in my use case I need to wait until the player is fully ready before disconnecting. Could you make the timeout configurable (e.g. in velocity.toml) and also allow -1 for no timeout at all? A slightly higher default (20–30s) would also help.
> > 
> > 
> > I personaly don't think this is a approach. 10 seconds should be sufficient for all plugins; a longer time may indicate a deadlock. I consider using properties to be a better option than config files, as it reduces the risk of the user breaking Velocity by not knowing what it does in the background; it is therefore more of a specific fine-tuning.
> 
> When restarting the proxy, I would simply tell the game server to reconnect the user as soon as possible, and if they are currently playing a mini-game, for example, they should be allowed to finish it.

I see where you’re coming from. In practice though, scenarios like letting players finish their game before a restart are usually better handled at the plugin or orchestration level. That way you can simply hold off on triggering the Velocity shutdown until everything on the game side is ready.

The pre-shutdown timeout in Velocity is intended as a short grace period for plugins to run quick cleanup before players are disconnected, rather than coordinating longer gameplay sessions. Once your plugin signals that the server is clear to restart, Velocity shutdown flow will take over reliably.

DartCZ:
The timeout can be now overridden using a system property:
`-Dvelocity.pre-shutdown-timeout=<seconds> `

4drian3d (approved):

DartCZ:
Hey, just checking in 🙂
Is there anything I could do to help move this PR forward? We’ve tested it on our side and it works as expected. I also know a few servers (including ours) that would really benefit from having this available upstream.
Thanks a lot for your time and work!

Timongcraft:
Currently, it's not clear to anyone using the event that the timeout is configurable.
You should note that in the javadoc.
It would also be nice if the value were accessible, as there is currently no way to read the value
(since the prop itself could be modified after the constant is initialized but before a plugin can read it).
And maybe just move it into a public constant in the event class, so it can easily be read and is also only initialized when the class is loaded and not at boot.

electronicboy:
if we wanted to document that we'd want it on the actual docs and nod to it from the logs with a small amount of info to devs/server owners

electronicboy:
Will look into merging this this weekend, right now it looks fine to me

Timongcraft:
> if we wanted to document that we'd want it on the actual docs and nod to it from the logs with a small amount of info to devs/server owners

Currently the javadoc talks about a fixed 10 seconds and I want that to point out that it is not fixed.

4drian3d (approved):

electronicboy (approved):

electronicboy:
Oh, look, the weekend

PR #1631, opened 2025-08-18
feat: Add more options for ping passthrough

ButterDebugger:
Heyo,

This is an updated version of [this pull request](https://github.com/PaperMC/Velocity/pull/1519) with added config migrations and removal of dead code. The original author of these changes doesn't appear to be responsive as of late, so I've made a new pull request with my additional changes. Please let me know if there is anything you would like me to modify or add, I'd be happy to make any code changes to hopefully see this get merged.

Thank you :D


TheMiningTeamYT:
I'll be honest, after several months of nothing happening I gave up on my pull request ever getting merged, so I stopped paying attention XD.
Thanks for your contributions to my code. I fully support this new pull request and hope it gets merged.

proxy/src/main/java/com/velocitypowered/proxy/connection/util/ServerListPingHandler.java
```diff
@@ -99,58 +100,64 @@ private CompletableFuture<ServerPing> attemptPingPassthrough(VelocityInboundConn
...
+        if (response == fallback) {
+          continue;
+        }
+
+        ServerPing.Version version;
+        if (mode.version) {
+          version = response.getVersion();
+        } else {
+          version = fallback.getVersion();
+        }
```
kennytv, line 121: Ternaries for each of those would make them a bit more compact. And not sure if this means much, but this looks functionally different to the description and mods search from before where it would continue iterating responses if they're null in the current one

proxy/src/main/resources/default-velocity.toml
```diff
@@ -51,20 +51,20 @@ announce-forge = false
...
-#                  used. If no backend servers can be contacted, Velocity won't display any
-#                  mod information.
-# - "description": Uses the description and mod list from the backend server. The first
-#                  server in the try (or forced host) list that responds is used for the
-#                  description and mod list.
-# - "all":         Uses the backend server's response as the proxy response. The Velocity
-#                  configuration is used if no servers could be contacted.
-ping-passthrough = "DISABLED"
+# Should Velocity pass the version number from the backend server when responding to server list ping requests?
+ping-passthrough-version = false
```
kennytv, line 55: This should be its own section rather than having the same prefix in the root section

kennytv (changes requested):

proxy/src/main/java/com/velocitypowered/proxy/connection/util/ServerListPingHandler.java
TheMiningTeamYT: On the ternaries, personal preference but that's an easy fix to make down the road. I don't think that should keep you from merging the pull request.
As for the difference in handling mod/description search: If you were to implement that in this code, you'd run the risk of, say, returning the player count from one server and the mod list from another, or returning the player count & mod list from a different server from the one the player is about to join, so unless someone jumps in here and says "hey I need that," I think it's fine the way it is.

proxy/src/main/resources/default-velocity.toml
TheMiningTeamYT: I agree. When I wrote this code, I did not know that was a thing you could do. Had I known, I would've done so.

ButterDebugger:
Closing this PR in favor of this newer one https://github.com/PaperMC/Velocity/pull/1870

PR #1646, opened 2025-09-07
Feat: HaProxyAddress

PinoOG:
- Exposes for the InboundConnection the SocketAddress related to the HaProxy Address which transports the player tcp connection. Mostly for debugging / diagnostic reasons.
- This can be useful to diagnose if players with high latency are sharing the same HaProxy Address.
- Given that the majority of networks run velocity behind services like TCPShield or similar, exposing the HaProxyAddress can give them some benefit for debugging.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/MinecraftConnection.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/MinecraftConnection.java
@@ -87,6 +87,7 @@ public class MinecraftConnection extends ChannelInboundHandlerAdapter {
   private final Channel channel;
   public boolean pendingConfigurationSwitch = false;
   private SocketAddress remoteAddress;
+  private @Nullable SocketAddress haProxyAddress;
   private StateRegistry state;
   private Map<StateRegistry, MinecraftSessionHandler> sessionHandlers;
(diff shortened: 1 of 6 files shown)


proxy/src/main/java/com/velocitypowered/proxy/connection/MinecraftConnection.java
```diff
@@ -158,6 +159,10 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
       } else if (msg instanceof HAProxyMessage proxyMessage) {
         this.remoteAddress = new InetSocketAddress(proxyMessage.sourceAddress(),
             proxyMessage.sourcePort());
+
+        final var proxyTransportAddress = (InetSocketAddress) ctx.channel().remoteAddress();
+        this.haProxyAddress = new InetSocketAddress(proxyTransportAddress.getAddress().getHostAddress(),
+                proxyTransportAddress.getPort());
```
Emilxyz, line 165:
```suggestion
        this.haProxyAddress = ctx.channel().remoteAddress();
```
wouldn't this work?

PinoOG: Mostly to avoid keeping a direct reference I created a new object.

PR #1655, opened 2025-09-22
Feature: per-server info forwarding mode

### LINKED ISSUE #1533: Per-server forwarding mode
### Requested Feature

Allow overriding forwarding mode for specific servers

### Why is this needed?

With bungeecord getting more and more outdated its support will inevitable start to fall off. For example, Fabric does not have a notable maintained mod for bungeecord forwarding. This means if a server network wanted to host a <=1.12.2 server and fabric 1.21.4 server it would be impossible without disabling forwarding entirely.

### Alternative Solutions

A: Completely disable forwarding: Loses all benefits of forwarding and with offline UUIDs would be unviable for existing setups
B: Middleman proxy: Technically possible with viaproxy (it strips the forwarding info so offline servers work), however there is no decent solution for this and this is very hacky

### Additional Information

Duplicate of #1145 - Making new issue due to former being closed unplanned and as to use new format
Technically duplicate of #566 - This issue is more direct and former has been inactive for 3.5 years

The main issue pointed out in #1145 is incompatibility with existing configuration. This is not necessarily the case. If instead of requiring each server have its mode set manually instead an optional overrides table is provided then this could be supported by checking the overrides for a forwarding mode and if one isn't specified use the default (existing) setting

MarcoLvr:
**What's this?**
With this edit every server can use a different forwarding mode to obtain and forward player info. For instance, if you are running a 1.12 (or lower version) server on a velocity proxy with MODERN player info forwarding the server doesn't support MODERN forwarding. So you can set LEGACY forwarding mode for that server and velocity will use ONLY FOR THAT SERVER the legacy forwarding mode. The default mode is "INHERIT" which uses the player info forwarding mode set in the config.

*Note: the INHERIT mode doesn't physically exists. Just set the server forwarding mode to NULL or avoid valuing it in the config*

The main configuration now supports two syntaxes for describing a server:

classic syntax (INHERIT mode): `server = "address:port"`
new syntax: `server = {address = "address:port", forwarding-mode = "MODE"}`

all servers created without defining a forwarding mode will use the INHERIT mode.

**Example of use**
Imagine a simple network with a lobby, a 1.8 game mode (like bedwars) and maybe a Fabric 1.21 server.
Fabric needs MODERN forwarding but the 1.8 server cannot use it. So you have two choices: 
1. Set the main forwarding mode to legacy (allow <1.13 players to join) and set MODERN mode only for the fabric server.
```
[...]
player-info-forwarding-mode = "legacy"
[...]
[servers]
try = ["lobby"]
lobby = "localhost:25566" //we leave this in INHERIT mode because in this case the lobby could be a 1.8 server too
bedwars = "localhost:25567" //we leave this in INHERIT mode because the main mode is legacy
modded =  { address = "localhost:25568", forwarding-mode = "MODERN" } //fabric needs MODERN forwarding
[...]
```
2. Set the main forwarding mode to MODERN and set LEGACY mode only for the 1.8 server.
```
[...]
player-info-forwarding-mode = "modern"
[...]
[servers]
try = ["lobby"]
lobby = "localhost:25566" //we leave this in INHERIT mode because in this case the lobby could be a >1.13 server
bedwars = { address = "localhost:25567", forwarding-mode = "LEGACY" }  //bedwars (1.8 server) needs legacy mode
modded =  "localhost:25568" //we leave this in INHERIT mode because the main mode is modern
[...]
```

This system could be very useful for these types of situations!


api/src/main/java/com/velocitypowered/api/proxy/config/ProxyConfig.java
```diff
@@ -84,7 +84,7 @@ public interface ProxyConfig {
    *
    * @return registered servers map
    */
-  Map<String, String> getServers();
+  Map<String, BackendServerConfig> getServers();
```
4drian3d: This change breaks compatibility with current versions. It is better to create another method, for example `#getBackendServers()`, to obtain the configuration of each server and deprecate this method

api/src/main/java/com/velocitypowered/api/proxy/server/ServerInfoForwardingMode.java
```diff
@@ -0,0 +1,19 @@
...
+ * reference the LICENSE file in the api top-level directory.
+ */
+
+package com.velocitypowered.api.proxy.server;
+
+/**
+ * Supported server info forwarding methods.
+ */
+public enum ServerInfoForwardingMode {
+    FOLLOWUP,
```
4drian3d: Add documentation on how this type works

proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
```diff
@@ -623,14 +624,31 @@ private Servers() {
...
+          if (entry.getValue() instanceof com.electronwill.nightconfig.core.CommentedConfig c) {
+            String address = null;
+            ServerInfoForwardingMode forwardingMode = null;
+            for (UnmodifiableConfig.Entry entry2 : c.entrySet()) {
+              if (entry2.getKey().equalsIgnoreCase("address")) {
+                address = entry2.getValue();
+              }
+              if (entry2.getKey().equalsIgnoreCase("forwarding-mode")) {
+                forwardingMode = ServerInfoForwardingMode.valueOf(ServerInfoForwardingMode.class, entry2.getValue());
+              }
```
4drian3d, line 659: I think it would be better to leave the forwarding value as optional, to give server administrators the freedom to use the syntax `someserver = {address = ”127.0.0.1:30068“}` so that they can easily configure the forwarding mode at any time

api/src/main/java/com/velocitypowered/api/proxy/server/ServerInfoForwardingMode.java
```diff
@@ -0,0 +1,19 @@
...
+ * The Velocity API is licensed under the terms of the MIT License. For more details,
+ * reference the LICENSE file in the api top-level directory.
+ */
+
+package com.velocitypowered.api.proxy.server;
+
+/**
+ * Supported server info forwarding methods.
+ */
+public enum ServerInfoForwardingMode {
```
4drian3d, line 15: An option representing NONE forwarding needs to be added

proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
```diff
@@ -560,9 +563,7 @@ public static VelocityConfiguration read(Path path) throws IOException {
 
       // Throw an exception if the forwarding-secret file is empty and the proxy is using a
       // forwarding mode that requires it.
-      if (forwardingSecret.length == 0
-              && (forwardingMode == PlayerInfoForwarding.MODERN
-              || forwardingMode == PlayerInfoForwarding.BUNGEEGUARD)) {
+      if (forwardingSecret.length == 0) {
         throw new RuntimeException("The forwarding-secret file must not be empty.");
       }
```
4drian3d, line 589: It would be good if this check were maintained and, in addition, if each server configured with MODERN or BUNGEEGUARD forwarding mode also had a similar check in the `#validate()` method.
I imagine that someone will set up a default forwarding mode of “NONE” and then set the forwarding mode to MODERN on some server, and that the file where they stored their secret has been deleted for some reason.

proxy/src/main/java/com/velocitypowered/proxy/server/VelocityRegisteredServer.java
```diff
@@ -73,9 +75,30 @@ public class VelocityRegisteredServer implements RegisteredServer, ForwardingAud
...
+    ServerInfoForwardingMode mode = serverInfo.getServerInfoForwardingMode();
+    if (mode != ServerInfoForwardingMode.MODERN
+            && mode != ServerInfoForwardingMode.BUNGEEGUARD) {
+      return;
+    }
+    VelocityConfiguration configuration = server.getConfiguration();
+    if (configuration.getForwardingSecret() == null || configuration.getForwardingSecret().length == 0) {
+      throw new IllegalStateException("You don't have a forwarding secret set. This is required if "
+              + "you are using MODERN or BUNGEEGUARD forwarding modes.");
+    }
```
4drian3d: I meant that this method should be transferred to the [`VelocityConfiguration#validate()`](https://github.com/PaperMC/Velocity/blob/dev/3.0.0/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java#L138) method, so that no value is applied if the configuration is reloaded and validated as invalid.
Putting it here will cause any data configured before verifying that the entire server configuration section is valid to be applied

4drian3d (changes requested):

proxy/src/main/java/com/velocitypowered/proxy/config/server/BackendServerConfigImpl.java
```diff
@@ -0,0 +1,90 @@
...
+    return forwardingMode;
+  }
+
+  public void setAddress(String address) {
+    this.address = address;
+  }
+
+  public  void setForwardingMode(ServerInfoForwardingMode forwardingMode) {
+    this.forwardingMode = forwardingMode;
+  }
```
4drian3d: Validations (`requireNonNull`) need to be added so that a null address or forwarding mode cannot be entered

_The empty constructor should also be removed_

4drian3d (changes requested):

api/src/main/java/com/velocitypowered/api/proxy/config/BackendServerConfig.java
```diff
@@ -0,0 +1,49 @@
...
+
+  /**
+   * Set the address of the backend server.
+   */
+  void setAddress(String address);
+
+  /**
+   * Set the forwarding mode of the backend server.
+   */
+  void setForwardingMode(ServerInfoForwardingMode forwardingMode);
```
Emilxyz: What's the point of these setters? They can probably be removed since they effectively do nothing and we don't offer changing any other config values via API.

proxy/src/main/java/com/velocitypowered/proxy/ProxyOptions.java
```diff
@@ -105,17 +106,23 @@ private static class ServerInfoConverter implements ValueConverter<ServerInfo> {
 
     @Override
     public ServerInfo convert(String s) {
-      String[] split = s.split(":", 2);
-      if (split.length < 2) {
-        throw new ValueConversionException("Invalid server format. Use <name>:<address>");
+      String[] split = s.split(":", 3);
+      if (split.length < 3) {
+        throw new ValueConversionException("Invalid server format. Use <name>:<address>:<forwardingmode>");
```
Emilxyz: This change removes the ability to pass an address string with a port as address. E.g., `--add-server=coolserver:127.0.0.1:30066` would now attempt to parse `30066` as forwarding mode. This will also break existing setups by making forwarding mode an required argument; it should probably be an additional optional, so like `<name>:<host>:[port]:[forwardingmode]`.

You should also update the string in `#valuePattern()` to match the new pattern

api/src/main/java/com/velocitypowered/api/proxy/config/ProxyConfig.java
```diff
@@ -86,6 +86,16 @@ public interface ProxyConfig {
    */
   Map<String, String> getServers();
```
Emilxyz, line 89: this should probably be deprecated in favor of the new `getBackendServers()`

proxy/src/main/java/com/velocitypowered/proxy/util/ServerForwardingModeUtil.java
```diff
@@ -0,0 +1,47 @@
...
+
+  /**
+   * Converts server info forward mode to Player info forwarding.
+   *
+   * @param configuration velocity config.
+   * @param mode server info forwarding mode.
+   *
+   * @return player info forwarding
+   */
+  public static PlayerInfoForwarding toPlayerInfoForwarding(VelocityConfiguration configuration, ServerInfoForwardingMode mode) {
```
Emilxyz: I think this method would nicely fit into `VelocityRegisteredServer` as something like `#getEffectivePlayerInfoForwarding()` (not entirely sure on the name here)

Emilxyz (changes requested):

MarcoLvr:
did everything! Let me know if there's any other change to do

api/src/main/java/com/velocitypowered/api/proxy/config/ProxyConfig.java
```diff
@@ -80,10 +80,12 @@ public interface ProxyConfig {
   /**
    * Get a Map of all servers registered in <code>velocity.toml</code>. This method does
    * <strong>not</strong> return all the servers currently in memory, although in most cases it
-   * does. For a view of all registered servers, see {@link ProxyServer#getAllServers()}.
+   * does. For a view of all registered servers, see {@link ProxyServer#getAllServers()}.<br><br>
+   * <b>DEPRECATED:</b> use getBackendServers() instead.
```
4drian3d: image

Instead of a comment, use the deprecated javadocs tag

proxy/src/main/java/com/velocitypowered/proxy/config/server/BackendServerConfigImpl.java
```diff
@@ -0,0 +1,99 @@
...
+import static java.util.Objects.requireNonNull;
+
+import com.velocitypowered.api.proxy.config.BackendServerConfig;
+import com.velocitypowered.api.proxy.server.ServerInfoForwardingMode;
+import java.util.Objects;
+
+/**
+ * The implementation of BackedServerConfig interface.
+ */
+public class BackendServerConfigImpl implements BackendServerConfig {
```
4drian3d: Now that all values are final, this class can easily be converted into a record

4drian3d (approved):
https://streamable.com/axixhc

lgtm

MarcoLvr:
@Emilxyz we are waiting for you :)

api/src/main/java/com/velocitypowered/api/proxy/config/BackendServerConfig.java
```diff
@@ -0,0 +1,46 @@
+/*
+ * Copyright (C) 2018-2021 Velocity Contributors
```
Emilxyz:
```suggestion
 * Copyright (C) 2018-2025 Velocity Contributors
```

api/src/main/java/com/velocitypowered/api/proxy/server/ServerInfoForwardingMode.java
```diff
@@ -0,0 +1,25 @@
+/*
+ * Copyright (C) 2018-2021 Velocity Contributors
```
Emilxyz:
```suggestion
 * Copyright (C) 2018-2025 Velocity Contributors
```

proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
```diff
@@ -639,16 +673,22 @@ private Servers(CommentedConfig config) {
...
     }
 
-    private Servers(Map<String, String> servers, List<String> attemptConnectionOrder) {
+    private Servers(Map<String, BackendServerConfig> servers, List<String> attemptConnectionOrder) {
       this.servers = servers;
       this.attemptConnectionOrder = attemptConnectionOrder;
     }
 
     private Map<String, String> getServers() {
```
Emilxyz: All remaining internal calls to this should be changed to the new `#getBackendServers()`. You can then remove this method and move its content into `VelocityConfiguration#getServers`, or maybe as a default in `ProxyConfig#getServers` itself even

proxy/src/main/java/com/velocitypowered/proxy/ProxyOptions.java
```diff
@@ -105,17 +106,29 @@ private static class ServerInfoConverter implements ValueConverter<ServerInfo> {
...
+          address = AddressUtil.parseAddress(split[1]);
+        }
       } catch (IllegalStateException e) {
         throw new ValueConversionException("Invalid hostname for server flag with name: " + split[0]);
       }
-      return new ServerInfo(split[0], address);
+      if (split.length == 4) {
+        try {
+          mode = ServerInfoForwardingMode.valueOf(split[3].toUpperCase());
+        } catch (IllegalStateException e) {
```
Emilxyz: This would throw a `IllegalArgumentException`, not illegal state

Emilxyz (commented):
> we are waiting for you

I'm not a team member, just sharing my thoughts, so my approval is not actually required. Also a friendly reminder that everyone is doing this on their free time, so it's not really needed to bump a review request after only a day. 

I gave this another look and it basically lgtm other than these minor things in review comments.

Another thought I had (but I would wait with doing any changes) is that maybe we can just use `ServerInfo` for this instead of adding a new type, they basically share the same fields and probably will continue to do so in the future as well.

Also there is the issue of the planned config api switch which is holding up other things already that would make changes to the config, so be prepared for this pr to also be affected by this

api/src/main/java/com/velocitypowered/api/proxy/server/ServerInfoForwardingMode.java
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * Supported per-server player info forwarding methods.
+ *
+ * @since 3.4.0
+ */
+public enum ServerInfoForwardingMode {
+    /**
+     * This type will follow the value of the player-info-forwarding-mode in the velocity configuration.
+     */
+    FOLLOWUP,
```
liorsl: IMHO this shouldn't exist, I think a better handling would be if a forwarding mode was not specified (e.g null), use the global one from config

MarcoLvr: yeah, but when you instantiate a ServerInfo object what server forwarding mode do you'll pass there? null? I prefer keeping the FOLLOWUP mode and also the implicit mode (not writing explicitly in the config the forwarding mode of one or more servers) which already takes the default forwarding mode.

liorsl: @MarcoLvr if a specific mode is not required make it null, then you will have something like:

```
// to determine app forwarding
ServerInfoForwardingMode mode = config.getServerInfoForwardingMode();
if (server.getServerInfoForwardingMode() != null)
  mode = server.getServerInfoForwardingMode()
```

api/src/main/java/com/velocitypowered/api/proxy/config/BackendServerConfig.java
```diff
@@ -0,0 +1,46 @@
...
+ * Exposes server configuration information that plugins may use.<br>
+ *
+ * <b>What's the forwarding mode?</b><br>
+ * The server can use a different mode to obtain and forward player info.<br>
+ * For instance, if you are running a 1.12 (or lower version) server on a velocity proxy with MODERN player info forwarding
+ * the server doesn't support MODERN forwarding. So you need to set LEGACY forwarding mode for that server
+ * and velocity will use ONLY FOR THAT SERVER the legacy forwarding mode.<br><br>
+ *
+ * @param address The address of the backend server.
+ * @param forwardingMode The forwarding mode of the backend server.
```
Emilxyz, line 27: I think this should mention something about what it means if it's `null`

api/src/main/java/com/velocitypowered/api/proxy/server/ServerInfo.java
```diff
@@ -20,6 +20,21 @@ public final class ServerInfo implements Comparable<ServerInfo> {
 
   private final String name;
   private final InetSocketAddress address;
+  private ServerInfoForwardingMode forwardingMode;
```
Emilxyz: This (as well as the getter/constructor) should be marked as `@Nullable`
The field can also be `final` if you assign it to `null` in the other constructor

api/src/main/java/com/velocitypowered/api/proxy/server/ServerInfo.java
```diff
@@ -20,6 +20,21 @@ public final class ServerInfo implements Comparable<ServerInfo> {
...
+
+  /**
+   * Creates a new ServerInfo object.
+   *
+   * @param name the name for the server
+   * @param address the address of the server to connect to
+   * @param forwardingMode the server info forwarding mode
```
Emilxyz: Should also mention here what passing `null` means, same as in BackendServerConfig like `[...], or {@code null} if the mode from the config should be used` (or something like that)

api/src/main/java/com/velocitypowered/api/proxy/server/ServerInfo.java
```diff
@@ -40,11 +55,21 @@ public final InetSocketAddress getAddress() {
     return address;
   }
 
+  /**
+   * Get what mode will the backend server use to communicate with velocity.
+   *
+   * @return FOLLOWUP mode if the server uses the same mode as set in the main config else one of the available modes
```
Emilxyz: This jd also needs updating

proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
```diff
@@ -627,14 +646,31 @@ private Servers() {
 
     private Servers(CommentedConfig config) {
       if (config != null) {
-        Map<String, String> servers = new HashMap<>();
+        Map<String, BackendServerConfig> servers = new HashMap<>();
         for (UnmodifiableConfig.Entry entry : config.entrySet()) {
-          if (entry.getValue() instanceof String) {
-            servers.put(cleanServerName(entry.getKey()), entry.getValue());
+          if (entry.getValue() instanceof com.electronwill.nightconfig.core.CommentedConfig c) {
```
Emilxyz: Why not import this?

MarcoLvr: IntelliJ autocompletition is stupid sometimes 😅

kennytv (approved):
Giving general approval, but notes for whoever wants to merge or consider this:
- There is also https://github.com/PaperMC/Velocity/pull/1357
- There is a PR to expose the forwarding mode that wasn't merged since it's deemed more of an internal property

JyJyJcr:
(I'm not a team member too.) I was also experimenting with implementing a per-server forwarding configuration and came across this PR.

I noticed that there are still 3 calls to `getPlayerInfoForwardingMode()` which assume the forwarding mode is a global setting:

* https://github.com/MarcoLvr/Velocity/blob/dfddfaad0e188c7f621f56964031ebd6182c4ed9/proxy/src/main/java/com/velocitypowered/proxy/connection/client/AuthSessionHandler.java#L88
* https://github.com/MarcoLvr/Velocity/blob/dfddfaad0e188c7f621f56964031ebd6182c4ed9/proxy/src/main/java/com/velocitypowered/proxy/connection/client/AuthSessionHandler.java#L146
* https://github.com/MarcoLvr/Velocity/blob/dfddfaad0e188c7f621f56964031ebd6182c4ed9/proxy/src/main/java/com/velocitypowered/proxy/connection/client/HandshakeSessionHandler.java#L150

These calls are no longer valid after introducing per-server configuration and should instead use `getConfiguredPlayerInfoForwarding()`. (It might also require changes to `MinecraftConnection`, since the callers do not have direct access to `VelocityRegisteredServer`.)

To prevent similar invalid calls in the future, it would be helpful to rename the method to `getDefaultPlayerInfoForwardingMode()`.

WouterGritter:
[VelocityRegisteredServer line 103-109](https://github.com/PaperMC/Velocity/pull/1655/changes#diff-04fcc8438caad1f709b8deb9eee72f7152111c21be224fc82356a18dff4a8d82R103) seems very fishy. It's converting one enum class to another, both with exactly the same value set.

I would consider moving the `PlayerInfoForwarding` enum to the `api` module instead and use that everywhere instead.

> There is a PR to expose the forwarding mode that wasn't merged since it's deemed more of an internal property

Though when making the decision that exposing this through the `api` module is undesired, I think it would be better to make `ServerInfo` an interface and implement it in the proxy module with the additional PlayerInfoForwarding field. I'm pretty sure that would break binary compatibility with plugins though... If that's something that we should be concerned about, extending this class as `class VelocityServerInfo extends ServerInfo` with the additional property would probably be best here. I just really dislike having two identical enums, one in the implementation module and one in the api module, and converting between the two.

PR #1656, opened 2025-10-01
Fix players disconnecting when updating boss bars

### LINKED ISSUE #1529: "Network Protocol Error" disconnect when using BossBar API
### Expected Behavior

Since 1.20.2's configuration phase, server switching clears boss bars. This change should be reflected in the bossbar tracker in some way (ideally resending the bossbars on server switch).

### Actual Behavior

Bossbar tracker does not reflect on these changes, thinking the client still can see bossbar(s) while it can not. As a result, using any kind of update action results in a null pointer in the client, disconnecting with Network Protocol Error since 1.20.5.

### Steps to Reproduce

Run the following code and switch servers using `/server` command.

```
    private BossBar bossBar;

    @Subscribe
    public void a(ServerPostConnectEvent e) {
        if (e.getPreviousServer() == null) {
            bossBar = BossBar.bossBar(
                    Component.text("Server: " + e.getPlayer().getCurrentServer().get().getServer().getServerInfo().getName()),
                    1,
                    BossBar.Color.BLUE,
                    BossBar.Overlay.PROGRESS
            );
            bossBar.addViewer(e.getPlayer());
        } else {
            bossBar.name(Component.text("Server: " + e.getPlayer().getCurrentServer().get().getServer().getServerInfo().getName()));
        }
    }
```

### Plugin List

Custom plugin for reproducing this issue.

### Velocity Version

```
> velocity info
[08:32:57 INFO]: Velocity 3.4.0-SNAPSHOT (git-d9f1016b-b481)
```

### Additional Information

There are basically 2 ways to solve this:
* Resend bossbars on server switch
* Remove player from tracked players 

I would prefer the first solution. It will also be consistent with <1.20.2. If deciding for this solution, make sure it's resent at the right time, so there isn't a window during which a packet could slip through before bossbar is resent (just a tip).

okx-code:
On 1.20.2, the Minecraft client started clearing boss bars after the login packet, which meant that the ProxyServer#showBossbar API would result in the player getting kicked if the boss bar they were previously shown was updated after switching servers.

Therefore, I have added BossBarManager which drops boss bar packets once the client enters the configure phase to ensure that they do not disconnect (if the client receives boss bar packets during the configure phase it seems to queue them up and then immediately disconnect at the end of the configure phase, leaving no opportunity to re-send the boss bar creation packet), and then re-adds the boss bar once the client logs in.

This ensures that clients do not receive boss bar updates for boss bars that they don't exist causing them to disconnect. I have also taken care to ensure that this logic only applies on 1.20.2 and up, as it is not necessary for older clients.

In addition, I removed boss bar tracking for backend servers for 1.20.2 and up clients - as it is no longer necessary for them to be manually removed due to the aforementioned change. 

Fixes #1529

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/adventure/VelocityBossBarImplementation.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/adventure/VelocityBossBarImplementation.java
@@ -53,15 +53,23 @@ public final class VelocityBossBarImplementation implements BossBar.Listener,
           viewer.getProtocolVersion(),
           viewer.translateMessage(this.bar.name())
       );
-      viewer.getConnection().write(BossBarPacket.createAddPacket(this.id, this.bar, name));
+      viewer.getBossBarManager().writeUpdate(this, BossBarPacket.createAddPacket(this.id, this.bar, name));
       return true;
     }
     return false;
   }
 
+  public void createDirect(final ConnectedPlayer viewer) {
+    final ComponentHolder name = new ComponentHolder(
+        viewer.getProtocolVersion(),
+        viewer.translateMessage(this.bar.name())
+    );
+    viewer.getConnection().write(BossBarPacket.createAddPacket(this.id, this.bar, name));
+  }
+
   public boolean viewerRemove(final ConnectedPlayer viewer) {
     if (this.viewers.remove(viewer)) {
-      viewer.getConnection().write(BossBarPacket.createRemovePacket(this.id, this.bar));
+      viewer.getBossBarManager().remove(this, BossBarPacket.createRemovePacket(this.id, this.bar));
       return true;
     }
     return false;
@@ -84,7 +92,7 @@ public final class VelocityBossBarImplementation implements BossBar.Listener,
           this.bar,
           new ComponentHolder(viewer.getProtocolVersion(), translated)
       );
-      viewer.getConnection().write(packet);
+      viewer.getBossBarManager().writeUpdate(this, packet);
     }
   }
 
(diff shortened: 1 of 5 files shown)


R00tB33rMan (approved):

R00tB33rMan (approved):
Not sure if you intended to have me re-review but I gave this a whirl on a test and production instance without issues on versions even below 1.20.2/3. Should be all set as far as I’m concerned!

4drian3d (approved):

NEZNAMY:
This is still not working properly. When sending update action during server switch caused by backend server shutdown, player is disconnected with Network Protocol Error. Works fine when switching server using `/server`.

Partially deobfuscated stack trace:
```
Description: Packet handling error

java.lang.NullPointerException: Cannot invoke "net.minecraft.client.gui.components.LerpingBossEvent.a(net.minecraft.world.BossEvent$BossBarColor)" because "$$3" is null
	at net.minecraft.client.gui.components.BossHealthOverlay$1.updateStyle(SourceFile:136)
	at net.minecraft.network.protocol.game.ClientboundBossEventPacket$UpdateStyleOperation.dispatch(SourceFile:267)
	at net.minecraft.network.protocol.game.ClientboundBossEventPacket.dispatch(SourceFile:91)
	at net.minecraft.client.gui.components.BossHealthOverlay.update(SourceFile:112)
	at net.minecraft.client.multiplayer.ClientPacketListener.handleBossUpdate(SourceFile:2067)
	at net.minecraft.network.protocol.game.ClientboundBossEventPacket.handle(SourceFile:87)
	at net.minecraft.network.protocol.game.ClientboundBossEventPacket.handle(SourceFile:14)
	at xl$a.a(SourceFile:55)
	at xl.b(SourceFile:38)
	at gfj.d(SourceFile:1337)
	at gfj.j(SourceFile:966)
	at net.minecraft.client.main.Main.main(SourceFile:250)
	at org.prismlauncher.launcher.impl.StandardLauncher.launch(StandardLauncher.java:105)
	at org.prismlauncher.EntryPoint.listen(EntryPoint.java:129)
	at org.prismlauncher.EntryPoint.main(EntryPoint.java:70)
```

WhiteProject1:
UP

PR #1659, opened 2025-10-07
feat: allow for custom ping default

stuffyerface:
When Velocity has not yet determined the player's ping, and the Velocity API is called, `-1` is returned. This change allows that value to be set to any `long` value via `velocity.toml`.


RealBauHD:
What do you need that for?

stuffyerface:
If you want to use -1 for other things; or if you want to use a positive value. The client sees a negative ping as unable to connect

4drian3d:
I see no need to replace the default ping value

> If you want to use -1 for other things

If you are making a plugin that obtains the player's ping, it should be in your plugin where you have the logic for displaying the ping

_The only reason I can see for actually changing this would be to modify the ping value that other plugins can obtain, such as placeholder or statistics plugins..._ Why would you want to do that?

stuffyerface:
fair enough

PR #1672, opened 2025-10-19
Refactor Velocity networking to be closer to vanilla

astei:
This is a really large PR, so review carefully.

This PR refactors Velocity's internal networking to use immutable(-ish) packets and splits encoding/decoding into separate codecs. This refactor isn't fully complete but there's enough to start doing more work on this.


proxy/src/main/java/com/velocitypowered/proxy/command/builtin/VelocityCommand.java
```diff
@@ -70,6 +71,7 @@
  */
 public final class VelocityCommand {
   private static final String USAGE = "/velocity <%s>";
+  private static final JoinConfiguration COMMA_JOINER = JoinConfiguration.commas(true);
```
masmc05: I feel like this became unused from an earlier version of the change bellow

proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/BossBarPacket.java
```diff
@@ -62,221 +64,52 @@ public class BossBarPacket implements MinecraftPacket {
...
-  private int overlay;
-  private short flags;
+
+  public UUID getUuid() {
+    return uuid;
+  }
+
+  public int getAction() {
+    return action;
+  }
```
masmc05: Why keeping both getter types is necessary?

proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/DisconnectPacket.java
```diff
@@ -21,36 +21,36 @@
...
+import com.velocitypowered.proxy.protocol.PacketCodec;
 import com.velocitypowered.proxy.protocol.ProtocolUtils;
 import com.velocitypowered.proxy.protocol.StateRegistry;
 import com.velocitypowered.proxy.protocol.packet.chat.ComponentHolder;
 import io.netty.buffer.ByteBuf;
 import net.kyori.adventure.text.Component;
-import org.checkerframework.checker.nullness.qual.Nullable;
 
-public class DisconnectPacket implements MinecraftPacket {
+public final class DisconnectPacket implements MinecraftPacket {
```
masmc05: Why some classes like this one aren't records? It has public constructor, all final fields and record style getters, so I can't think of a reason

astei: I don't think it's worth converting every class to a record, it just wastes time. This feels like a very unimportant nit.

proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/BossBarPacket.java
astei: Because the AI (Claude) wanted to make sure the code compiled and it sometimes chose to amend the existing code and sometimes add in getters. I didn't steer it in that direction.

Again, while it's redundant, it's a nit that someone else can fix.

proxy/src/main/java/com/velocitypowered/proxy/protocol/netty/MinecraftDecoder.java
```diff
@@ -95,14 +100,16 @@ private void tryDecode(ChannelHandlerContext ctx, ByteBuf buf) throws Exception
     }
   }
 
-  private void doLengthSanityChecks(ByteBuf buf, MinecraftPacket packet) throws Exception {
-    int expectedMinLen = packet.decodeExpectedMinLength(buf, direction, registry.version);
-    int expectedMaxLen = packet.decodeExpectedMaxLength(buf, direction, registry.version);
+  private void doLengthSanityChecks(ByteBuf buf,
+      com.velocitypowered.proxy.protocol.PacketCodec<? extends MinecraftPacket> codec)
```
novitpw, line 104: Why is the full name listed here?

Timongcraft:
It seems like this PR (`MultiVersionPacketRegistry`) is using fastutil classes excluded by the buildscript, which you can see when trying to ping the server.

https://github.com/PaperMC/Velocity/blob/8f65a8142029717d3af09ecf82792f7345dc873a/proxy/build.gradle.kts#L85

[latest.log](https://github.com/user-attachments/files/24271427/latest.log) - https://pastes.dev/92K4RMv1gM - https://mclo.gs/zdPdSDr

PR #1681, opened 2025-10-31
Increase TabComplete max char limit to respect command block character limit

ZoeWithTheE:
This change increases the TabComplete limit in order to fix the now broken pasting of larger commands into command blocks after Mojang's change to the way Tab Complete works. It has been increased to 32767 from 2048, and this increase has no negative impact on performance, stability, or protocol compliance.

Mojang clients and servers already allow command strings up to 32767 characters, and this update brings Velocity in line with the vanilla protocol. This prevents unnecessary disconnections when pasting valid command block contents while retaining full protection against malformed packets.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/TabCompleteRequestPacket.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/TabCompleteRequestPacket.java
@@ -31,7 +31,7 @@ import org.checkerframework.checker.nullness.qual.Nullable;
 
 public class TabCompleteRequestPacket implements MinecraftPacket {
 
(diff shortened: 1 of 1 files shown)


ZoeWithTheE:
No open issue for this bug. This is the error being observed: [PasteBin velocity logs](https://pastebin.com/1wE4vzbY)

electronicboy:
We and paper intentionally limited this in the past in order to restrict ongoing DoS attacks, I do ponder if it makes any sense to move this restriction elsewhere, I guess some limited adventage with brig is that it might be able to deal with this stuff a bit smarter than the older bukkit command system so maybe this wasn't too much of a problem for us, outside of network buffers

PR #1682, opened 2025-11-02
feat: PlayerServerBrandEvent

ProdPreva1l:
This event allows modification of the server brand that is relayed to the player via the api.


electronicboy:
Removal of attribution is not something we desire to support

PR #1683, opened 2025-11-02
Explicitly create the configuration file from the default

linsaftw:
This fixes a race-condition that affects Pterodactyl and similar panel hosted servers. It does not affect on Linux servers. This makes sure the configuration is explicitly created before doing any migrations or default setting.

This issue seems to have been tormeting newcomers for a while.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
@@ -33,6 +33,7 @@ import com.velocitypowered.proxy.config.migration.MotdMigration;
 import com.velocitypowered.proxy.config.migration.TransferIntegrationMigration;
 import com.velocitypowered.proxy.util.AddressUtil;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import java.io.InputStream;
 import java.io.IOException;
 import java.net.InetSocketAddress;
 import java.net.URL;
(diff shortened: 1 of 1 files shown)


linsaftw:
I fixed inline import and changed with proper import above

Emilxyz:
I don't think this actually fixes the issue since we are already doing this using our config library: https://github.com/PaperMC/Velocity/blob/b6b6b20fe97cd9cb0d6b4e817d3e7db72aca2d8d/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java#L495

To my understanding, the "real" issue is that the docker images used by those panels include those "broken" minimal configuration files and our config library lacks the functionality to populate missing fields. I think a switch to a different config library is planned in the future

linsaftw:
> I don't think this actually fixes the issue since we are already do this using our config library:
> 
> https://github.com/PaperMC/Velocity/blob/b6b6b20fe97cd9cb0d6b4e817d3e7db72aca2d8d/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java#L495
> 
> To my understanding the "real" issue is that the docker images used by those panels include those "broken" minimal configuration files and our config library lacks the functionality to populate missing fields. I think a switch to a different config library is planned in the future

I worked with this fix with a person having the same issue, he said doing a explicit copy fixes it. He made a different patch for a different project but it worked. Today he is available again and I will ask him to test again. I will confirm soon.

linsaftw:
Hey! Good news. This patch seems to completely patch the issue. I tested on a pterodactyl panel and now the file properly generates after applying this exact patch. You can safely merge. Thank you very much.

electronicboy:
The nature of the PR comment around aspects like a race condition make 0 sense here, the config file either exists or it doesn't unless there is some janky issue elsewhere, in which, I would rather deal with a solution for that rather than manually copying the configuration file to the disk to work around something which now sounds even more questionable than before.

Either way, given past interactions with this author, I do not feel comfortable merging code submitted by them into this repo due to the legal risks it could potentially entail.

PR #1688, opened 2025-11-11
feat: add server field to ServerPostConnectEvent

bemoty:
Since the [ServerConnectedEvent](https://jd.papermc.io/velocity/3.4.0/com/velocitypowered/api/event/player/ServerConnectedEvent.html) doesn't seem to support sending plugin messages to the newly connected server, it would be helpful to be able to access the new server in ServerPostConnectEvent, where plugin messaging works.

Sorry for the weird formatting changes, I couldn't get my formatter to stop indenting the constructor arguments... the checkstyle seems to pass though...


Emilxyz:
The "new" server is already available in `Player#getCurrentServer()` when this event is fired, so I don't think we to need add it to this event

bemoty:
True, not sure how I didn't realize that..

PR #1701, opened 2026-01-01
Implement loading of extra plugin directories via cli arg

theminecoder:
Same idea as PaperMC/Paper#13447

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/plugin/VelocityPluginManager.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/plugin/VelocityPluginManager.java
@@ -85,34 +85,54 @@ public class VelocityPluginManager implements PluginManager {
    * @param directory the directory to load from
    * @throws IOException if we could not open the directory
    */
+  @Deprecated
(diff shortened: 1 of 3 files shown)


proxy/src/main/java/com/velocitypowered/proxy/plugin/VelocityPluginManager.java
```diff
@@ -85,34 +85,54 @@ public void registerPlugin(PluginContainer plugin) {
    * @param directory the directory to load from
    * @throws IOException if we could not open the directory
    */
+  @Deprecated
+  public void loadPlugins(Path directory) throws IOException {
```
Emilxyz, line 89: Is there a reason we're keeping this and marking it as deprecated? Since it's an internal method, I think it would be okay to remove it

theminecoder: Mostly just didn't want to interrupt those doing strange things w/ reflection/forks straight away since it can't be as clean the paper impl

PR #1706, opened 2026-01-12
Pass register plugin messages through the serverbound configuration plugin message packet event

florianreuth:
In our plugin, I need to cancel a channel sent by Fabric's Particle API in order to prevent a confliction between Fabric API and our packet transformers. This patch allows me to intercept the register/unregister payloads and remove the channel.

I'm not familiar with the internals, so I wasn't sure where to put this best (as putting it into the registrar directly would probably introduce destructive API changes)

This patch is required for the top header commit on this tree: https://github.com/ViaVersion/ViaVersion/tree/next/fabric-particle-workaround

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientConfigSessionHandler.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientConfigSessionHandler.java
@@ -24,6 +24,7 @@ import com.velocitypowered.api.event.player.configuration.PlayerConfigurationEve
 import com.velocitypowered.api.event.player.configuration.PlayerFinishConfigurationEvent;
 import com.velocitypowered.api.event.player.configuration.PlayerFinishedConfigurationEvent;
 import com.velocitypowered.api.proxy.messages.ChannelIdentifier;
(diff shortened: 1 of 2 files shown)


electronicboy:
Velocity not exposing channels which are handled internally is intentional, I get the desire here, but I feel that this opens up a door I'd rather not

florianreuth:
It's not handled in the configuration state at all, without my patch Velocity will just cancel/forward the packet to the server. I in general don't get the purpose of the protocol / why it still exists, as it makes things just more complicated and doesn't offer any benefits. Is there any other way I can edit the packet with internals/raw API?

electronicboy:
If those where to be handled it would be to fit it into the existing path that is used in play, rather than exposing the ability to cancel this stuff; it's generally an intentional design of velocity that plugins can't intercept arbitary stuff, if you want to intercept a channel you should register it yourself so that the standard event handling works, and then you can deal with controlling if it forwards or not

florianreuth:
Pretty sure you can't register minecraft:register itself so that wouldn't really work

electronicboy:
I guess in this case where you care about messing with actual registrations that wouldn't be viable to expose in general, especially given that we don't really want to support stuff screwing over other stuff; For this you'd pretty much be delegated to using a packet library

PR #1708, opened 2026-01-14
feat(player-connection): add config option to disable errors from player while they cant connect to a backend server.

### LINKED ISSUE #1707: Add new setting to prevent connected player errors in console
### Requested Feature

Add a new config option to disable the error message when a connected player cannot connect.

https://github.com/PaperMC/Velocity/blob/372a3b28bd3bbaebf5934b4582dbb2feba70a93e/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java#L718

### Why is this needed?

Especially on big servers, the red error messages can spam the console, even when player connection messages are disabled (there is already a setting for logging connection messages)

### Alternative Solutions

- don’t add a new setting and only add the config check for the log-player-connections boolean
- ignore these messages in console, wich can be very annoying or even confusing

### Additional Information

The error is send here: https://github.com/PaperMC/Velocity/blob/372a3b28bd3bbaebf5934b4582dbb2feba70a93e/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java#L718

TheBjoRedCraft:
Fixes #1707 

This PR adds a new config option to the velocity config (refactoring the default toml) and using the configuration value to either prevent or allow sending the error message when a connected player fails to join the backend server.

### DIFF
--- a/proxy/src/main/resources/default-velocity.toml
+++ b/proxy/src/main/resources/default-velocity.toml
@@ -146,6 +146,9 @@ log-command-executions = false
 # and disconnecting from the proxy.
 log-player-connections = true
 
+# Enables logging of errors when players fail to connect to backend servers.
+logConnectionErrors = true
+
 # Allows players transferred from other hosts via the
 # Transfer packet (Minecraft 1.20.5) to be received.
 accepts-transfers = false
--- a/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
@@ -437,6 +437,10 @@ public class VelocityConfiguration implements ProxyConfig {
     return advanced.isLogPlayerConnections();
   }
 
+  public boolean isLogConnectionErrors() {
+    return advanced.isLogConnectionErrors();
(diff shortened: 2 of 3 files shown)


R00tB33rMan (changes requested):
Add a space in between the "if" and the configuration getter.

Although I'm pretty sure this isn't a feature Velocity intends to implement in this format. At the very "best," likely a way to suppress the stacktrace, but to retain the unable to connect indicator

Timongcraft:
IMO a config migration would be nice. Even though there is nothing to be migrated but to have the option show up for old users, unlike what happened with the rate limiter options where a at least few users where confused

proxy/src/main/resources/default-velocity.toml
```diff
@@ -146,6 +146,9 @@ log-command-executions = false
 # and disconnecting from the proxy.
 log-player-connections = true
 
+# Enables logging of errors when players fail to connect to backend servers.
+logConnectionErrors = true
```
Emilxyz, line 150:
```suggestion
log-connection-errors = true
```

Emilxyz (changes requested):

PR #1711, opened 2026-01-17
Add Thai localization for Velocity proxy messages

LIUZAZA007:


Leguan16:
Hey, nice work!

However, I think localization is generally handled through crowdin.
https://papermc-io.crowdin.com/velocity

electronicboy:
Cheers, however, as mentioned, localisations need to be done through crowdin otherwise they won't be tracked properly

Timongcraft:
Thai should now be available on Crowdin

PR #1712, opened 2026-01-19
Provide an IDE pattern hint for plugin IDs

Warriorrrr:
This would allow IDEs to show a warning when a plugin id isn't valid while you're writing it instead of during compile.


electronicboy:
checkstyle

Warriorrrr:
checkstyle is now happy

electronicboy (approved):

PR #1728, opened 2026-02-11
Introduce SchedulerBackend to fix VelocitySchedulerTest intermittent failure

WouterGritter:
As the title says.

I was having intermittent failures because of these tests when building a Velocity fork on GitHub actions.

The culprit was already identified when the tests were first written:
```java
// TODO: The timings here will be inaccurate on slow systems.
```

Note that `DeterministicSchedulerBackend` was written partially with the help of an LLM.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/scheduler/ExecutorSchedulerBackend.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/scheduler/ExecutorSchedulerBackend.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright (C) 2018-2026 Velocity Contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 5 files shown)


Warriorrrr:
Do you have any link to a workflow run that failed because of this? Velocity uses github actions too and does not have this issue.

electronicboy:
I think I managed to make it fail once when my system was on fire, having time based tests is often a bit of a nightmare, having testing envs work differently to real environments is also a bit of a wildcard too

astei (approved):
This seems fine to me - making the scheduler totally deterministic _is_ the canonical solution to the problem, I just didn't have an incentive to deal with it.

FWIW, I would expect those tests to be most flaky on lower-end hardware and the last time I touched this, it was mostly to improve reliability but not all of it was made deterministic.

WouterGritter:
For reference, a failed job before this merge: https://github.com/GemstoneGG/Velocity-CTD/actions/runs/21911993248/job/63268303163

But again, this build is for a velocity fork (though it didn't make any meaningful change to the scheduler).

PR #1729, opened 2026-02-11
Feature/health and filter

Eministar:
### Description
This PR introduces a robust Server Health Checking system and an advanced IP Filtering framework (Maintenance, Blacklist, Whitelist) to Velocity. It also adds several administrative commands to manage these systems in real-time.

#### 1. Server Health Checker
- **Periodic Pings**: Automatically monitors backend servers at a configurable interval.
- **Health States**: Servers are categorized as `HEALTHY`, `DEGRADED` (high latency), or `UNHEALTHY` (timeout/offline).
- **Events**: Fires a `ServerHealthChangeEvent` whenever a server's status changes, allowing plugins to react (e.g., for smart load balancing).
- **Configuration**: Preserves defaults with a new `[health-check]` section in `velocity.toml`.

#### 2. Advanced IP Filter Manager
- **Dynamic Blacklisting**: Support for individual IPs and CIDR ranges with optional reasons and durations (temporary bans).
- **Whitelisting**: Allows specific IPs/ranges to bypass maintenance mode.
- **Maintenance Mode**: A global toggle to restrict access, featuring customizable MiniMessage support for the kick message.
- **Early Handshake Filtering**: Connections are checked during the Handshake/Status phase to prevent MOTD leakage and save resources.

#### 3. New Administrative Commands
- `/maintenance <on|off> [message]`: Manage the global maintenance state.
- `/ipblacklist <add|remove|list> <pattern> [duration] [reason]`: Manage the blacklist.
- `/ipwhitelist <add|remove|list> <pattern>`: Manage the whitelist.
- `/vhelp`: A dynamic help command that only lists commands the sender has permission to use.

#### 4. Localization & UX
- Full integration with `messages.properties` for all new command outputs.
- Enhanced "Unknown command" messaging: Now provides a helpful hint to use `/vhelp` when a user enters an invalid command.

### Technical Details
- **Scheduler**: The health checker uses the internal `VelocityVirtualPlugin` to safely register scheduled tasks.
- **Brigadier**: All commands are implemented using the Brigadier API for full tab-completion support.
- **Performance**: IP lookups use `ConcurrentHashMap` and lazy expiration cleanup to ensure minimal overhead during the handshake.

### Testing
- Verified local build with `./gradlew build`.
- Ran all 202 proxy unit tests (all passing).
- Manual verification:
  - Confirmed health transitions in logs when stopping/starting backend servers.
  - Verified that blacklisted IPs are kicked before the MOTD is sent.
  - Verified CIDR matching for both IPv4 and IPv6.

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/event/proxy/server/ServerHealthChangeEvent.java
+++ b/api/src/main/java/com/velocitypowered/api/event/proxy/server/ServerHealthChangeEvent.java
@@ -1,0 +1,111 @@
+/*
+ * Copyright (C) 2018-2026 Velocity Contributors
+ *
+ * The Velocity API is licensed under the terms of the MIT License. For more details,
+ * reference the LICENSE file in the api top-level directory.
+ */
+
+package com.velocitypowered.api.event.proxy.server;
+
+import com.velocitypowered.api.proxy.server.RegisteredServer;
+import com.velocitypowered.api.proxy.server.ServerHealthInfo;
+import com.velocitypowered.api.proxy.server.ServerHealthStatus;
+import org.checkerframework.checker.nullness.qual.NonNull;
+
+/**
+ * Fired when a server's health status changes.
+ *
+ * @since 3.5.0
+ */
+public final class ServerHealthChangeEvent {
+
+  private final RegisteredServer server;
+  private final ServerHealthStatus previousStatus;
+  private final ServerHealthStatus newStatus;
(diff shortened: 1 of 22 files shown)


electronicboy:
This generally feels entirely out of scope for velocity itself and would be much better suited as a plugin, I'm not comfortable taking on this, either. Offering user commands but 0 persistence is flawed, but, we're also not really in a position to want to provide persistence, that should be on plugins to deal with.

PR #1733, opened 2026-02-15
Fix CallbackCommand permission

NeyrowZ:
Should allow disabling `/velocity:callback` auto complete by disabling the `velocity.command.callback` permission to the desired players.


Timongcraft:
Please read though what I've sent you.
> This is intentional, see https://github.com/PaperMC/Velocity/pull/1627 for details

Timongcraft:
And locking an API feature behind a permission would be really bad. >ou could hide it, that is not wanted on the API side though (see the PR) and you can hide the cmd yourself using the [PlayerAvailableCommandsEvent](<https://jd.papermc.io/velocity/3.5.0/com/velocitypowered/api/event/command/PlayerAvailableCommandsEvent.html>).

NeyrowZ:
Thanks i'll look into it.

PR #1734, opened 2026-02-15
Fix dimension reading in JoinGamePacket

xphorror:
Velocity中`JoinGamePacket.java`对于1.7.10维度ID的传递只使用1字节，这导致部分模组维度的维度ID传递过程中发生数据截断，传递了非法数据给客户端，导致客户端崩溃。使用4字节处理维度ID可以解决该问题。我使用了Claude来分析和修改（因为我不会编程），图片为Claude修改完成的总结。

我是在游玩GTNH 2.8.0整合包时发现的此问题，当时我的服务器使用Velocity转发，通过Velocity实现玩家登录验证，我在游玩途中发现在私人维度（维度ID为180）中下线再加入客户端会崩溃，而跳过Velocity转发则不会，报错日志提示非法维度ID-76。

---

In Velocity, the `JoinGamePacket.java` uses only 1 byte to transmit the dimension ID for 1.7.10. This causes data truncation during the transmission of dimension IDs for some modded dimensions, sending illegal data to the client and causing the client to crash. Using 4 bytes to handle the dimension ID can resolve this issue. I used Claude to analyze and modify the code (since I don't know how to program), and the image shows the summary of the modifications completed by Claude.

I discovered this issue while playing the GTNH 2.8.0 modpack. At the time, my server used Velocity for forwarding and implemented player login authentication through Velocity. During gameplay, I found that logging out and re-joining in a private dimension (dimension ID 180) would cause the client to crash, whereas bypassing Velocity forwarding did not. The error log indicated an illegal dimension ID of -76.
Snipaste_2026-02-15_22-35-01

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/JoinGamePacket.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/JoinGamePacket.java
@@ -246,7 +246,11 @@ public class JoinGamePacket implements MinecraftPacket {
     if (version.noLessThan(ProtocolVersion.MINECRAFT_1_9_1)) {
       this.dimension = buf.readInt();
     } else {
-      this.dimension = buf.readByte();
+      // Read as signed byte, then convert to unsigned for modded dimension IDs > 127
+      // (e.g. dimension 180 would be read as signed byte -76, but should be stored as 180).
+      // Preserve -1 (Nether) as-is since it is the only standard negative dimension ID.
+      int raw = buf.readByte();
+      this.dimension = (raw < -1) ? (raw & 0xFF) : raw;
     }
     if (version.noGreaterThan(ProtocolVersion.MINECRAFT_1_13_2)) {
       this.difficulty = buf.readUnsignedByte();


WouterGritter:
I think it'll be beneficial to write some tests for this behavior.

WouterGritter:
However, this logic seems to do functionally nothing as `dimension` is only ever written back to a `ByteBuf` using `ByteBuf#writeByte`, which will turn this (now signed) byte into an unsigned byte anyway.
See line 401 of this file https://github.com/xphorror/Velocity/blob/c7b83326182ed24c2edb462de04364f81ada5b0b/proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/JoinGamePacket.java#L401

Unless somehow the packet is read as a pre-1.9.1 packet and written as a post-1.9.1 packet? If this is the case please ignore this comment. Though that seems unlikely.

Are you sure this change solves your problem?

xphorror:
> However, this logic seems to do functionally nothing as `dimension` is only ever written back to a `ByteBuf` using `ByteBuf#writeByte`, which will turn this (now signed) byte into an unsigned byte anyway.
> See line 401 of this file https://github.com/xphorror/Velocity/blob/c7b83326182ed24c2edb462de04364f81ada5b0b/proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/JoinGamePacket.java#L401
> 
> Unless somehow the packet is read as a pre-1.9.1 packet and written as a post-1.9.1 packet? If this is the case please ignore this comment. Though that seems unlikely.
> 
> Are you sure this change solves your problem?
---
Yes,the problem fixed.Maybe I can record a video upload to youtube later.

xphorror:
https://www.youtube.com/watch?v=NHzHfsw4t64

WouterGritter:
I don't like this. But it works. RespawnPacket encodes the dimension as an int instead of a byte, which is where the actual bug occurs.

Reading this as an unsigned byte, then just writing it back as a byte in JoinGamePacket is a NOP. When switching servers, due to some weirdness with old clients we need to send a RespawnPacket that's generated off of this JoinGamePacket. That's when we write this previously signed byte as an int, resulting in a different value than the client expects.

This is a huge hack because we still need `-1` to work for the nether.
I suggest changing the comment for why exactly this is needed, for future readers:
```java
// Vanilla 1.7.10 uses a signed byte for dimension (-1 Nether, 0 Overworld, 1 End).
// Modded servers hack this to an unsigned byte to allow dim IDs up to 255.
// We must store the canonical int here because RespawnPacket.fromJoinGame copies this
// value into a packet that encodes dimension as a 4-byte int (RespawnPacket.encode,
// pre-1.16 logic). Sign-extending a modded byte like 0xB4 to int -76 sends an
// illegal dimension ID to the client and crashes it.
//
// The wire byte is ambiguous: 0xFF could be vanilla Nether (-1) or modded dim 255.
// We resolve in Nether's favor since vanilla 1.7.10 only ever uses -1/0/1.
```

bluegreensea:
I think this just hack the allowed range from -128 ~ 127 to -1~254 (255 is -1).
If modded servers register a dimension in -2 ~ -128, it may still broken.

And the second things, GTNH is not "hack this to an unsigned byte to allow dim IDs up to 255".
It hack to an int to completely fix this vanilla bug.
(https://github.com/GTNewHorizons/Hodgepodge/blob/master/src/main/java/com/mitchej123/hodgepodge/mixins/early/minecraft/packets/MixinS01PacketJoinGame_FixDimensionID.java)

WouterGritter:
> If modded servers register a dimension in -2 ~ -128, it may still broken.

Not sure if this is possible, in older versions anyway. `JoinGamePacket` handles the dimension as a byte, regardless of its sign, it will always encode/decode as the same hex value. Though in vanilla minecraft this is definitely expected to be signed, because the nether is `-1`. `RespawnPacket` just encodes this as an integer instead, which is also expected to be signed, as the vanilla protocol expects the values `-1`, `0` and `1`.

> GTNH is not "hack this to an unsigned byte to allow dim IDs up to 255".

IMO it is a hack. The nether being `-1` suggests that this value should be signed. Mods using a value over 127 would have needed to patch this bug also; `RespawnPacket` needs to return a signed `-1` only ever for the nether, otherwise it should treat the byte as unsigned (and have a range from 0 - 255). This patch makes the range for the dimension value essentially be -1 - 254 (inclusive), as 255 is effectively mapped to -1.

rieverholic:
I don't think Velocity should care how the backend server processes the dimension ID. In that sense, I think it's better to just forward whatever GTNH attaches to/expects from a login packet (see GTNewHorizons/Hodgepodge#341). This is the patch I use:
```
--- a/proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/JoinGamePacket.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/JoinGamePacket.java
@@ -53,6 +53,8 @@ public class JoinGamePacket implements MinecraftPacket {
   private int portalCooldown; // 1.20+
   private int seaLevel; // 1.21.2+
   private boolean enforcesSecureChat; // 1.20.5+
+  private int dimensionId; // GTNH
+  private boolean extraDimensionId;

   public int getEntityId() {
     return entityId;
@@ -259,6 +261,10 @@ public class JoinGamePacket implements MinecraftPacket {
     if (version.noLessThan(ProtocolVersion.MINECRAFT_1_15)) {
       this.showRespawnScreen = buf.readBoolean();
     }
+    if (buf.readableBytes() >= 4) {
+      this.extraDimensionId = true;
+      this.dimensionId = buf.readInt();
+    }
   }

   private void decode116Up(ByteBuf buf, ProtocolVersion version) {
@@ -410,6 +416,9 @@ public class JoinGamePacket implements MinecraftPacket {
     if (version.noLessThan(ProtocolVersion.MINECRAFT_1_15)) {
       buf.writeBoolean(showRespawnScreen);
     }
+    if (extraDimensionId) {
+      buf.writeInt(dimensionId);
+    }
   }

   private void encode116Up(ByteBuf buf, ProtocolVersion version) {
```

proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/JoinGamePacket.java
```diff
@@ -246,7 +246,11 @@ private void decodeLegacy(ByteBuf buf, ProtocolVersion version) {
     if (version.noLessThan(ProtocolVersion.MINECRAFT_1_9_1)) {
       this.dimension = buf.readInt();
     } else {
-      this.dimension = buf.readByte();
+      // Read as signed byte, then convert to unsigned for modded dimension IDs > 127
+      // (e.g. dimension 180 would be read as signed byte -76, but should be stored as 180).
+      // Preserve -1 (Nether) as-is since it is the only standard negative dimension ID.
+      int raw = buf.readByte();
+      this.dimension = (raw < -1) ? (raw & 0xFF) : raw;
```
WouterGritter, line 253:
```suggestion
      // Vanilla 1.7.10 uses a signed byte for dimension (-1 Nether, 0 Overworld, 1 End).
      // Modded servers hack this to an unsigned byte to allow dim IDs up to 255.
      // We must store the canonical int here because RespawnPacket.fromJoinGame copies this
      // value into a packet that encodes dimension as a 4-byte int (RespawnPacket.encode,
      // pre-1.16 logic). Sign-extending a modded byte like 0xB4 to int -76 sends an
      // illegal dimension ID to the client and crashes it.
      //
      // The wire byte is ambiguous: 0xFF could be vanilla Nether (-1) or modded dim 255.
      // We resolve in Nether's favor since vanilla 1.7.10 only ever uses -1/0/1.
      short raw = buf.readUnsignedByte();
      this.dimension = raw == 0xFF ? -1 : raw;
```

Explains the reason behind this "hack" more clearly, and makes the assignment explicit (only replace the value with `-1` if the byte is `xFF`)

Caedis: GTNH dev here that fixed the issue originally 

Just a heads up, because the dim id is actually an int, it can also have values larger than 256 (see [PersonalSpace](https://github.com/GTNewHorizons/PersonalSpace) dimensions with a large number of players on a public server for example)

The change I made in Hodgepodge keeps the original byte read for backwards compatibility and appends an raw int to the end of the login packet so that JourneyMap correctly reads the the player's current dim if they login in a dim outside of the byte range.

WouterGritter: It's stored as an int, but on this branch (as per the 1.7.10 protocol), it's read as an unsigned byte (so we need to read it into a short). This is actually where the discrepancy comes from; in another place 1.7.10 writes it as an integer instead, messing up the sign that some mods expect (see the comment).

Adding support for an additional field in the login packet is most likely not something we'd consider for Velocity.

WouterGritter:
Squashed manually in 2676520c6a54bac0529544793c982dd701b338d9

bluegreensea:
I can confirm that after https://github.com/PaperMC/Velocity/commit/2676520c6a54bac0529544793c982dd701b338d9, login in a negative number world (-2 ~ -128) it crash client
Test with dimension id -7 (and it was map to 249 by that commit):
```
java.lang.IllegalArgumentException: Could not get provider type for dimension 249, does not exist
	at Launch//net.minecraftforge.common.DimensionManager.getProviderType(DimensionManager.java:148)
	at Launch//net.minecraftforge.common.DimensionManager.createProviderFor(DimensionManager.java:302)
	at Launch//net.minecraft.world.WorldProvider.func_76570_a(WorldProvider.java:160)
	at Launch//net.minecraft.client.multiplayer.WorldClient.<init>(WorldClient.java:54)
	at Launch//net.minecraft.client.network.NetHandlerPlayClient.func_147280_a(NetHandlerPlayClient.java:871)
	at Launch//net.minecraft.network.play.server.S07PacketRespawn.func_148833_a(SourceFile:32)
	at Launch//net.minecraft.network.play.server.S07PacketRespawn.func_148833_a(SourceFile:13)
	at Launch//net.minecraft.network.NetworkManager.func_74428_b(NetworkManager.java:212)
	at Launch//net.minecraft.client.multiplayer.PlayerControllerMP.func_78765_e(PlayerControllerMP.java:273)
	at Launch//net.minecraft.client.Minecraft.func_71407_l(Minecraft.java:1602)
	at Launch//net.minecraft.client.Minecraft.func_71411_J(Minecraft.java:973)
	at Launch//net.minecraft.client.Minecraft.func_99999_d(Minecraft.java:898)
	at Launch//net.minecraft.client.main.Main.main(SourceFile:148)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
	at java.base/java.lang.reflect.Method.invoke(Method.java:565)
	at System//net.minecraft.launchwrapper.Launch.rfb$realLaunch(Launch.java:250)
	at System//net.minecraft.launchwrapper.Launch.launch(Launch.java:35)
	at System//net.minecraft.launchwrapper.Launch.main(Launch.java:60)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
	at java.base/java.lang.reflect.Method.invoke(Method.java:565)
	at com.gtnewhorizons.retrofuturabootstrap.Main.main(Main.java:207)
	at com.gtnewhorizons.retrofuturabootstrap.MainStartOnFirstThread.lambda$main$1(MainStartOnFirstThread.java:47)
	at java.base/java.lang.Thread.run(Thread.java:1474)
```

proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/JoinGamePacket.java
Caedis: yeah, its only the login packet that writes/reads it as a byte for whatever reason

PR #1736, opened 2026-02-22
Implemented plugin update folder

### LINKED ISSUE #809: Feature request: Add update folder mechanism similar to what Bukkit servers do

OakLoaf:
The goal of this PR was to make progress on a plugin update folder for Velocity.

Tasks:
- [x] Initial implementation
- [x] Discuss nullability annotation for `VelocityConfiguration#getUpdateFolderName`
- [x] Discuss default config value
- [x] Discuss what the config value should be to entirely disable the update folder
- [x] Test final implementation

closes https://github.com/PaperMC/Velocity/issues/809
based on https://github.com/PaperMC/Velocity/pull/842

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
@@ -423,18 +423,23 @@ public class VelocityServer implements ProxyServer, ForwardingAudience {
 
     try {
       Path pluginPath = Path.of("plugins");
+      Path updatePath = configuration.isUpdateFolderEnabled() ? pluginPath.resolve(configuration.getUpdateFolderName()) : null;
 
       if (!pluginPath.toFile().exists()) {
         Files.createDirectory(pluginPath);
-      } else {
-        if (!pluginPath.toFile().isDirectory()) {
-          logger.warn("Plugin location {} is not a directory, continuing without loading plugins",
-              pluginPath);
-          return;
-        }
-
-        pluginManager.loadPlugins(pluginPath);
       }
+
+      if (updatePath != null && !updatePath.toFile().exists()) {
+        Files.createDirectory(updatePath);
+      }
+
+      if (!pluginPath.toFile().isDirectory()) {
+        logger.warn("Plugin location {} is not a directory, continuing without loading plugins",
+            pluginPath);
+        return;
+      }
+
+      pluginManager.loadPlugins(pluginPath, updatePath);
     } catch (Exception e) {
       logger.error("Couldn't load plugins", e);
     }
--- a/proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
@@ -423,7 +423,8 @@ public class VelocityServer implements ProxyServer, ForwardingAudience {
 
     try {
       Path pluginPath = Path.of("plugins");
-      Path updatePath = configuration.isUpdateFolderEnabled() ? pluginPath.resolve(configuration.getUpdateFolderName()) : null;
+      String updateFolderName = System.getProperty("velocity.update-folder-name");
+      Path updatePath = updateFolderName != null ? pluginPath.resolve(updateFolderName) : null;
 
       if (!Files.exists(pluginPath)) {
         Files.createDirectory(pluginPath);
--- a/proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
@@ -425,15 +425,15 @@ public class VelocityServer implements ProxyServer, ForwardingAudience {
       Path pluginPath = Path.of("plugins");
       Path updatePath = configuration.isUpdateFolderEnabled() ? pluginPath.resolve(configuration.getUpdateFolderName()) : null;
 
-      if (!pluginPath.toFile().exists()) {
+      if (!Files.exists(pluginPath)) {
         Files.createDirectory(pluginPath);
       }
 
-      if (updatePath != null && !updatePath.toFile().exists()) {
+      if (updatePath != null && !Files.exists(updatePath)) {
         Files.createDirectory(updatePath);
       }
 
-      if (!pluginPath.toFile().isDirectory()) {
+      if (!Files.isDirectory(pluginPath)) {
         logger.warn("Plugin location {} is not a directory, continuing without loading plugins",
             pluginPath);
         return;
--- a/proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
@@ -418,12 +418,16 @@ public class VelocityServer implements ProxyServer, ForwardingAudience {
     }
   }
 
+  public String getUpdateFolderName() {
+    return System.getProperty("velocity.update-folder-name", "update");
+  }
+
   private void loadPlugins() {
     logger.info("Loading plugins...");
 
     try {
       Path pluginPath = Path.of("plugins");
-      String updateFolderName = System.getProperty("velocity.update-folder-name", "update");
+      String updateFolderName = getUpdateFolderName();
       Path updatePath = !updateFolderName.isEmpty() ? pluginPath.resolve(updateFolderName) : null;
 
       if (!Files.exists(pluginPath)) {
--- a/proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
@@ -423,8 +423,8 @@ public class VelocityServer implements ProxyServer, ForwardingAudience {
 
     try {
       Path pluginPath = Path.of("plugins");
-      String updateFolderName = System.getProperty("velocity.update-folder-name");
-      Path updatePath = updateFolderName != null ? pluginPath.resolve(updateFolderName) : null;
+      String updateFolderName = System.getProperty("velocity.update-folder-name", ".update");
+      Path updatePath = !updateFolderName.isEmpty() ? pluginPath.resolve(updateFolderName) : null;
 
       if (!Files.exists(pluginPath)) {
         Files.createDirectory(pluginPath);
--- a/proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
@@ -423,7 +423,7 @@ public class VelocityServer implements ProxyServer, ForwardingAudience {
 
     try {
       Path pluginPath = Path.of("plugins");
-      String updateFolderName = System.getProperty("velocity.update-folder-name", ".update");
+      String updateFolderName = System.getProperty("velocity.update-folder-name", "update");
       Path updatePath = !updateFolderName.isEmpty() ? pluginPath.resolve(updateFolderName) : null;
 
       if (!Files.exists(pluginPath)) {
--- a/proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
@@ -418,16 +418,12 @@ public class VelocityServer implements ProxyServer, ForwardingAudience {
     }
   }
 
-  public String getUpdateFolderName() {
-    return System.getProperty("velocity.update-folder-name", "update");
-  }
-
   private void loadPlugins() {
     logger.info("Loading plugins...");
 
     try {
       Path pluginPath = Path.of("plugins");
-      String updateFolderName = getUpdateFolderName();
+      String updateFolderName = System.getProperty("velocity.update-folder-name", "update");
       Path updatePath = !updateFolderName.isEmpty() ? pluginPath.resolve(updateFolderName) : null;
 
       if (!Files.exists(pluginPath)) {
--- a/proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
@@ -655,7 +655,7 @@ public class VelocityServer implements ProxyServer, ForwardingAudience {
   }
 
   /**
-   * Calls {@link #shutdown(boolean, Component)} with the default reason "Proxy shutting down.".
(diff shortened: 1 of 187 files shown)


OakLoaf:
Before marking this as ready for review I wanted to discuss the 3 topics mentioned in the tasks list, so any discussion regarding these would be greatly appreciated.

Toffikk:
i believe an empty string would be the best way to go about disabling it, same as it is done in other softwares

```
update-folder = ""
```

OakLoaf:
> i believe an empty string would be the best way to go about disabling it, same as it is done in other softwares
> 
> ```
> update-folder = ""
> ```

I completely agree, that's what my head initially went to. Wasn't sure if there was a different practice in toml as I've not worked with it before :)

I'll adjust that one

OakLoaf:
I also noticed that in most places in the config object it seems like the nullable annotation is not used so I'll remove it unless suggested otherwise for `VelocityConfiguration#getUpdateFolderName`

Toffikk:
> I also noticed that in most places in the config object it seems like the nullable annotation is not used so I'll remove it unless suggested otherwise for `VelocityConfiguration#getUpdateFolderName`

technically the name is not nullable, as it is an empty string (and should be checked against being empty, not null) in the `update-folder = ""` syntax, so marking it as `@Nullable` would be wrong.

OakLoaf:
> technically the name is not nullable, as it is an empty string (and should be checked against being empty, not null) in the `update-folder = ""` syntax, so marking it as `@Nullable` would be wrong.

That's true, currently it was requiring being set to null to disable it, I'll make these changes now. I am also going to opt for `.update` as default for now as that is my personal preference and what I think makes most sense.

proxy/src/main/java/com/velocitypowered/proxy/plugin/VelocityPluginManager.java
```diff
@@ -82,19 +83,19 @@ public void registerPlugin(PluginContainer plugin) {
   /**
    * Loads all plugins from the specified {@code directory}.
    *
-   * @param directory the directory to load from
+   * @param pluginDirectory the directory to load from
```
Toffikk, line 86: should also include a `@param` for the update dir

proxy/src/main/java/com/velocitypowered/proxy/plugin/VelocityPluginManager.java
```diff
@@ -82,19 +83,19 @@ public void registerPlugin(PluginContainer plugin) {
...
    * @throws IOException if we could not open the directory
    */
   @SuppressFBWarnings(value = "RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE",
       justification = "I looked carefully and there's no way SpotBugs is right.")
-  public void loadPlugins(Path directory) throws IOException {
-    checkNotNull(directory, "directory");
-    checkArgument(directory.toFile().isDirectory(), "provided path isn't a directory");
+  public void loadPlugins(Path pluginDirectory, @Nullable Path updateDirectory) throws IOException {
+    checkNotNull(pluginDirectory, "directory");
+    checkArgument(pluginDirectory.toFile().isDirectory(), "provided path isn't a directory");
```
Toffikk: why aren’t we doing this check for the update directory too?

proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
```diff
@@ -800,6 +810,7 @@ private Advanced(CommentedConfig config) {
         this.announceProxyCommands = config.getOrElse("announce-proxy-commands", true);
         this.logCommandExecutions = config.getOrElse("log-command-executions", false);
         this.logPlayerConnections = config.getOrElse("log-player-connections", true);
+        this.updateFolderName = config.getOrElse("update-folder", ".update");
```
Toffikk: imo either the `name` suffix should be removed or the config key should have the `name` suffix as well for consistency between those two

proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
```diff
@@ -866,6 +877,10 @@ public boolean isLogPlayerConnections() {
       return logPlayerConnections;
     }
 
+    public String getUpdateFolderName() {
```
Toffikk: same here, refer to the comment om the lines above

Toffikk: but thinking about this more; the getter can have the suffix for more clarity ig

proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
OakLoaf: I opted for the suffix for clarity in the code but didn't think that same clarity was necessary/suitable in the config file. Happy to adjust either if needed though

proxy/src/main/java/com/velocitypowered/proxy/plugin/VelocityPluginManager.java
```diff
@@ -35,6 +35,7 @@
 import com.velocitypowered.proxy.plugin.loader.VelocityPluginContainer;
 import com.velocitypowered.proxy.plugin.loader.java.JavaPluginLoader;
 import com.velocitypowered.proxy.plugin.util.PluginDependencyUtils;
+import edu.umd.cs.findbugs.annotations.Nullable;
```
Emilxyz: nit: I *think* we normally use nullability annotations from checkerframework. However, there are also several places using jspecify/jetbrains nullability annotations so idk if this really matters.

proxy/src/main/java/com/velocitypowered/proxy/plugin/VelocityPluginManager.java
```diff
@@ -182,6 +192,67 @@ protected void configure() {
...
+        }
+      }
+    }
+
+    for (PluginDescription availableUpdate : availableUpdates) {
+      PluginDescription potentialMatch = found.get(availableUpdate.getId());
+
+      // This should not happen but acts as a failsafe
+      if (availableUpdate.getSource().isEmpty()) {
+        logger.warn("No source found for plugin {} found", availableUpdate.getId());
```
Toffikk: duplicate `found` here

OakLoaf: Good spot, resolved :)

Timongcraft:
I think Velocity's update folder should mirror the Paper update folder's name (update without a dot prefix).
Also why does this need config option? I don't see why the average admin would need to change this and for advanced setups a system property is better suited IMO.
I also don't get why we would create the update folder by default or why you don't utilize Java's Files class and instead use #toFile.

OakLoaf:
> I think Velocity's update folder should mirror the Paper update folder's name (update without a dot prefix).

The dot prefix acts as a way of keeping the folder at the top of the directory list and separates it from being mixed in with the plugin folders. I personally think it makes the update folder much more accessible and easy to use, which is why I set it as the default. I would go as far to say that Paper could also benefit from changing the default to include the prefix.


> Also why does this need config option? I don't see why the average admin would need to change this and for advanced setups a system property is better suited IMO.

I personally think a config option is nice to have for people who prefer different formats or languages. It could use a system properpty but I don't really think this being a config option has any drawbacks


> I also don't get why we would create the update folder by default or why you don't utilize Java's Files class and instead use #toFile.

I am happy to change this if needed

Timongcraft:
> The dot prefix acts as a way of keeping the folder at the top of the directory list and separates it from being mixed in with the plugin folders. I personally think it makes the update folder much more accessible and easy to use, which is why I set it as the default. I would go as far to say that Paper could also benefit from changing the default to include the prefix.

That makes sense, thank you for explaining.

> I personally think a config option is nice to have for people who prefer different formats or languages. It could use a system properpty but I don't really think this being a config option has any drawbacks

I don't get how different languages/formats matter here. For the average admin it is that folder and the folder being standardized also simplifies debugging for other people.

> I am happy to change this if needed

Ofc it is not needed, IMO it's better to use the NOI though, as that is the more modern and cleaner solution.

OakLoaf:
> I don't get how different languages/formats matter here. For the average admin it is that folder and the folder being standardized also simplifies debugging for other people.

That's fair and very true. I guess it's almost like if Paper added a config option for `.paper-remapped` which would be unnecessary. With system properties the option would still be there for the update folder so I do get your point. Would we stick to a single `velocity.update-folder-name` where an empty string disables the update folder or do you think a second `velocity.disable-update-folder` would be suitable?

> Ofc it is not needed, IMO it's better to use the NOI though, as that is the more modern and cleaner solution.

Was just having a look through the changes in this PR and the only places I am using `toFile` are for `isDirectory` which was what was previously used in other parts of the code, I'll migrate `VelocityPluginManager` and `VelocityServer` over to use NIO instead

Timongcraft:
Well IMO we don't need to disable the update folder if we just don't auto create it, no?

OakLoaf:
> Well IMO we don't need to disable the update folder if we just don't auto create it, no?

I based the need to disable off of [this discussion](https://github.com/PaperMC/Velocity/pull/842#issuecomment-3675488124) in the other pr regarding this topic

Timongcraft:
> > Well IMO we don't need to disable the update folder if we just don't auto create it, no?
> 
> I based the need to disable off of discussion in the other pr regarding this topic

Well then I think we might want a config section to disable it for individual plugins, but it also more seems like something for plugin authors to handle. If they are truly malicious they can do anything and if not then users should either raise an issue or search for an alternative. Also this should only become an issue after this feature is introduced, no?

OakLoaf:
> Well then I think we might want a config section to disable it for individual plugins, but it also more seems like something for plugin authors to handle. If they are truly malicious they can do anything and if not then users should either raise an issue or search for an alternative. Also this should only become an issue after this feature is introduced, no?

In that case, I think sticking to just an empty string to disable is probably suitable enough as it doesn't require any extra options.

Timongcraft:
Should probably do a config migration to disable the update folder for older setups as there's the chance that some systems/forks implemented their own version which could break

OakLoaf:
> Should probably do a config migration to disable the update folder for older setups as there's the chance that some systems/forks implemented their own version which could break

How would we approach that for the system property route? 

Although, any forks should be aware of what they are merging so I would personally say responsibility falls on them to make sure they aren't merging duplicate features/incompatibilities.

OakLoaf:
Is there anything else that might hold this PR back? 
And is it possible to approve the PR workflow? Not quite sure how that works but thought I'd ask :)

4drian3d (commented):
I like the implementation
The only issue I see is the folder name, since on some operating systems, putting a `.` at the beginning of a folder/file name hides it, so I'd prefer to keep a default similar to the one in Paper

OakLoaf:
> I like the implementation The only issue I see is the folder name, since on some operating systems, putting a `.` at the beginning of a folder/file name hides it, so I'd prefer to keep a default similar to the one in Paper

Very good point, I hadn't considered that, I've just pushed that one. 
Only other thing that I have just thought - could it be worth adding a `getUpdateFolderName` API method to `VelocityServer`?

WouterGritter:
I don't like how a `velocity.update-folder-name` system property of `""` needs to be set in order to disable this feature. That sounds to me like an empty folder name, which is more likely to refer to the "current directory" (whatever this might be, I would assume the `plugins/` "root" folder with no prior knowledge), instead of "no directory" / "disable the feature", and would need to be documented elsewhere. Even still, it won't be clear to someone reading startup parameters what this feature does when set to an empty string, and that they should look up the docs.

> I think Velocity's update folder should mirror the Paper update folder's name (update without a dot prefix).

See [PluginInitializerManager.java#parse](https://github.com/PaperMC/Paper/blob/bd74bf6581ce81e59bdab07eadbfbe5d485eefa7/paper-server/src/main/java/io/papermc/paper/plugin/PluginInitializerManager.java#L52)

Paper also uses the "update" (no dot) folder, but it gets it from the configuration. If we really want to mirror how Paper does things (which is not a bad thing, I think), we should also get this value from the config. An advantage of this is that we can add a comment explaining the (imo) ambiguous behavior when this is set to `""`.

If we don't want this to be a configuration thing, I would prefer a second system property, read as a boolean defaulting to `true`, that disables this feature (instead of an empty string doing so).

For what it's worth, I'm surprised paper went the config route and I'd have wanted to see a system property solution there too. But afaik is this left over from Bukkit.

OakLoaf:
> I don't like how a velocity.update-folder-name system property of "" needs to be set in order to disable this feature. That sounds to me like an empty folder name, which is more likely to refer to the "current directory" (whatever this might be, I would assume the plugins/ "root" folder with no prior knowledge), instead of "no directory" / "disable the feature", and would need to be documented elsewhere. Even still, it won't be clear to someone reading startup parameters what this feature does when set to an empty string, and that they should look up the docs.

I think whether a second system property is added or not an empty string would have the same effect on the feature nonetheless. I personally am indifferent on whether a second system property is added or not. If someone wants to disable the feature they are likely to search the docs so will either find the system property and information regarding disabling it.

proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
```diff
@@ -418,12 +418,16 @@ private void doStartupConfigLoad() {
     }
   }
 
+  public String getUpdateFolderName() {
+    return System.getProperty("velocity.update-folder-name", "update");
+  }
```
4drian3d: Personally, I think this method is somewhat unnecessary. It would be an _oversimplification_ of a method call that’s already straightforward, and very few people would actually need it _(not to mention that it is exclusive to the proxy module, which is not normally accessible)_

WouterGritter: Most places where `Boolean#getBoolean(String)` (helper to convert a system variable to a boolean) is called, it's a static field.
```java
private static final String UPDATE_FOLDER_NAME =
        System.getProperty("velocity.update-folder-name", "update");
```

OakLoaf:
Was there anything left to do on my part for this? :)

4drian3d (approved):

R00tB33rMan (approved):

maxcom1 (approved):

twisti-dev:
Any updates on this?

PR #1744, opened 2026-03-18
delay teardown ownership until duplicate login check passes

### LINKED ISSUE #1740: `ConnectedPlayer#disconnected` called even when connection fails due to duplicate connection
### Expected Behavior

This method should'nt be called.

### Actual Behavior

This method does get called, possibly resulting in undesired `VelocityBossBarImplementation#viewerDisconnected` calls.
For reference, the entire `ConnectedPlayer#disconnected` method:
```java
public void disconnected() {
  for (final VelocityBossBarImplementation bar : this.bossBars) {
    bar.viewerDisconnected(this);
  }
}
```

### Steps to Reproduce

Add a printline/log statement in `ConnectedPlayer#disconnected`, join the proxy with minecraft account A, try to re-join the proxy with the same account.
The second connection will fail, and `ConnectedPlayer#disconnected` still gets called.

### Plugin List

n/a

### Velocity Version

Compiled from latest commit e0db25664fc82eabd9fde5aac22a2311a9765975 + the log statement in `disconnected()`.

### Additional Information

This currently isn't an issue, as bossbars won't be registered to the ConnectedPlayer. However if more logic ever gets added to this method this might be an issue.

My proposed fix is to gate the `connection.disconnected()` call in `VelocityServer#unregisterConnection`:
```java
public void unregisterConnection(ConnectedPlayer connection) {
  connectionsByName.remove(connection.getUsername().toLowerCase(Locale.US), connection);
  connectionsByUuid.remove(connection.getUniqueId(), connection);
  connection.disconnected(); // Only call when actually removed from the map(s).
}
```

However, this _may_ result in unwanted behavior due to `VelocityBossBarImplementation` storing a `Set<ConnectedPlayer> viewers`. If a player _is_ added to this list while they haven't been properly registered in `VelocityServer` yet, this will cause a memory leak. It could be worth it to make sure we only add viewers to bossbars when they've been properly registered (I don't think that's the case here though, because that would mean bossbar packets are being sent before the ServerLoginSuccessPacket, but this is untested).

ZECHEESELORD:
Fixes #1740.

Prevents `ConnectedPlayer` teardown from running for a rejected duplicate login attempt by only taking teardown ownership after the duplicate login gate passes.

Tests:
- Added a regression test to verify a rejected duplicate login does not reach teardown/unregister logic
- Added a test to verify owned players still teardown on disconnect

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/client/AuthSessionHandler.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/client/AuthSessionHandler.java
@@ -100,13 +100,16 @@ public class AuthSessionHandler implements MinecraftSessionHandler {
       ConnectedPlayer player = new ConnectedPlayer(server, profileEvent.getGameProfile(),
           mcConnection, inbound.getVirtualHost().orElse(null), inbound.getRawVirtualHost().orElse(null), onlineMode,
           inbound.getHandshakeIntent(), inbound.getIdentifiedKey());
-      this.connectedPlayer = player;
       if (!server.canRegisterConnection(player)) {
         player.disconnect0(
             Component.translatable("velocity.error.already-connected-proxy", NamedTextColor.RED),
             true);
         return CompletableFuture.completedFuture(null);
       }
+      // Only take teardown ownership after the duplicate login gate passes.
+      // Before this point, a rejected preregistration connection must not run
+      // ConnectedPlayer teardown.
+      this.connectedPlayer = player;
 
       if (server.getConfiguration().isLogPlayerConnections()) {
         logger.info("{} has connected", player);
--- a/proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/config/ClientboundCustomReportDetailsPacket.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/config/ClientboundCustomReportDetailsPacket.java
@@ -22,7 +22,6 @@ import com.velocitypowered.proxy.connection.MinecraftSessionHandler;
 import com.velocitypowered.proxy.protocol.MinecraftPacket;
 import com.velocitypowered.proxy.protocol.ProtocolUtils;
 import io.netty.buffer.ByteBuf;
-import java.util.HashMap;
 import java.util.Map;
 
 public class ClientboundCustomReportDetailsPacket implements MinecraftPacket {
--- a/proxy/src/test/java/com/velocitypowered/proxy/connection/client/AuthSessionHandlerTest.java
+++ b/proxy/src/test/java/com/velocitypowered/proxy/connection/client/AuthSessionHandlerTest.java
@@ -1,0 +1,122 @@
+/*
+ * Copyright (C) 2026 Velocity Contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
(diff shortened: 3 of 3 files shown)


WouterGritter:
I think this can be fixed even better by only ever instantiating `ConnectedPlayer` after `VelocityServer#canRegisterConnection` returns true. This avoid ever creating a BossBarManager/ChatQueue/whatever (this is done in ConnectedPlayer's constructor).

`VelocityServer#canRegisterConnection` currently consumes the `ConnectedPlayer`, but only ever checks its username and UUID. We can consume a `com.velocitypowered.api.proxy.player.PlayerInfo` instead, which is effectively a record of `String username, UUID uuid`, or just consume the username/UUID to perform the check.

Edit:
The only issue is that we won't have access to `ConnectedPlayer#disconnect0` then. This method only references the VelocityServer instance and the `connection` (`mcConnection` in AuthSessionHandler's context). Maybe we can pull this method into `MinecraftConnection`, and have `ConnectedPlayer#disconnect0` just call this method on it's own `connection` to avoid duplicate code?

Edit # 2: The canonical solution is to use `GameProfile`, which contains the player's name and uuid, as we already have access to this through `profileEvent.getGameProfile()` here. See https://github.com/PaperMC/Velocity/pull/1748/

ZECHEESELORD:
> I think this can be fixed even better by only ever instantiating `ConnectedPlayer` after `VelocityServer#canRegisterConnection` returns true. This avoid ever creating a BossBarManager/ChatQueue/whatever (this is done in ConnectedPlayer's constructor).
> 
> `VelocityServer#canRegisterConnection` currently consumes the `ConnectedPlayer`, but only ever checks its username and UUID. We can consume a `com.velocitypowered.api.proxy.player.PlayerInfo` instead, which is effectively a record of `String username, UUID uuid`, or just consume the username/UUID to perform the check.
> 
> Edit: The only issue is that we won't have access to `ConnectedPlayer#disconnect0` then. This method only references the VelocityServer instance and the `connection` (`mcConnection` in AuthSessionHandler's context). Maybe we can pull this method into `MinecraftConnection`, and have `ConnectedPlayer#disconnect0` just call this method on it's own `connection` to avoid duplicate code?
> 
> Edit # 2: The canonical solution is to use `GameProfile`, which contains the player's name and uuid, as we already have access to this through `profileEvent.getGameProfile()` here. See #1748


this is definitely cleaner than my version-

#1744 does add regression coverage. If you feel that it is necessary, feel free to cherrypick the test commit from here, or I can adapt the tests to #1748’s shape.

happy to defer to #1748 if that’s the preferred direction.

WouterGritter:
I'll leave it up to the maintainers whether or not we need unit tests in the second PR, or @ZECHEESELORD if you have the time feel free to cherry pick my commits into this PR and adapt the tests that you already have in place.

ZECHEESELORD:
> I'll leave it up to the maintainers whether or not we need unit tests in the second PR, or @ZECHEESELORD if you have the time feel free to cherry pick my commits into this PR and adapt the tests that you already have in place.

Thanks- I’ll fold your commits into #1744 when I get a chance (and then adapt the tests, so we have regression coverage)

4drian3d:
The problem with this is that disconnections caused by a duplicate login are represented in the API, specifically in the DisconnectEvent (https://jd.papermc.io/velocity/3.5.0/com/velocitypowered/api/event/connection/DisconnectEvent.LoginStatus.html#CONFLICTING_LOGIN),  so by preventing the creation of the player object and its teardown, the event would not be triggered

ZECHEESELORD:
That’s a fair point. I think I got a bit too focused on the internal cleanup path and not enough on the API surface.
I’m going to move this PR back to the original ownership based fix (which keeps that behavior intact and still fixes #1740).

WouterGritter:
> The problem with this is that disconnections caused by a duplicate login are represented in the API, specifically in the DisconnectEvent (https://jd.papermc.io/velocity/3.5.0/com/velocitypowered/api/event/connection/DisconnectEvent.LoginStatus.html#CONFLICTING_LOGIN), so by preventing the creation of the player object and its teardown, the event would not be triggered

The problem is now that we are already exposing the Player object to plugins through this event. Plugins can do whatever they want with the Player, including registering bossbars. Currently `disconnected()` only cleans up the bossbars, which could be omitted if we didn't ever expose the ConnectedPlayer object, but because we do we must consider a plugin registering bossbars on player disconnect. This sounds like weird behavior, but it's something that's possible nevertheless.

Also, it's interesting that a `DisconnectEvent` can be called on a player who was never connected, as far as the plugins go. Is this intentional?

PR #1748, opened 2026-03-22
Delay instantiating ConnectedPlayer to after Velocity#canRegisterConnection returns true

### LINKED ISSUE #1740: `ConnectedPlayer#disconnected` called even when connection fails due to duplicate connection

WouterGritter:
Fixes https://github.com/PaperMC/Velocity/issues/1740.
Inspired by https://github.com/PaperMC/Velocity/pull/1744, see https://github.com/PaperMC/Velocity/pull/1744#issuecomment-4105976714.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
@@ -692,16 +692,16 @@ public class VelocityServer implements ProxyServer, ForwardingAudience {
   /**
    * Checks if the {@code connection} can be registered with the proxy.
    *
-   * @param connection the connection to check
+   * @param gameProfile the {@link GameProfile} of the incoming connection
    * @return {@code true} if we can register the connection, {@code false} if not
    */
-  public boolean canRegisterConnection(ConnectedPlayer connection) {
+  public boolean canRegisterConnection(GameProfile gameProfile) {
     if (configuration.isOnlineMode() && configuration.isOnlineModeKickExistingPlayers()) {
       return true;
     }
-    String lowerName = connection.getUsername().toLowerCase(Locale.US);
+    String lowerName = gameProfile.getName().toLowerCase(Locale.US);
     return !(connectionsByName.containsKey(lowerName)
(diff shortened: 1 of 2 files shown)


WouterGritter:
Closing, see https://github.com/PaperMC/Velocity/pull/1744#issuecomment-4112646957

PR #1749, opened 2026-03-22
Mixed player-info-forwarding-mode

onePlaceholder:
I needed it for my server, so why not share it with the world.
Added the "secure_mix" option to "player-info-forwarding-mode" allowing the use of modern forwarding on modern servers and bungeeguard on legacy servers


Timongcraft:
Just so you're aware there are other PR's such as #1655 that implement a per-server forwarding mode, which IMO is preferable.

Also adding util methods to the enum would probably make it more readable

PR #1756, opened 2026-03-27
feat: add boss bar viewer

RealBauHD:


WouterGritter:
Why does `Set<BossBar> bossBars` need to be monotonically non-null? Seems like no real performance gain for less readability.

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
```diff
@@ -587,25 +590,41 @@ public void resetTitle() {
...
+          if (this.bossBars.isEmpty()) {
+            this.bossBars = null;
+          }
+        }
       }
     }
   }
 
+  @Override
+  public @UnmodifiableView @org.jspecify.annotations.NonNull Iterable<? extends BossBar> activeBossBars() {
```
WouterGritter, line 621:
```suggestion
  public @UnmodifiableView @NonNull Iterable<? extends BossBar> activeBossBars() {
```

`org.checkerframework.checker.nullness.qual.NonNull` is imported

Timongcraft:
> Why does `Set<BossBar> bossBars` need to be monotonically non-null? Seems like no real performance gain for less readability.

What are you talking about?
For me at least it is Nullable:
https://github.com/PaperMC/Velocity/pull/1756/changes#diff-1cd097c1de92fff8c43bc33eb4f52405eb5396f9e8cc922a4502e4cad12e1f88R189

WouterGritter:
> > Why does `Set<BossBar> bossBars` need to be monotonically non-null? Seems like no real performance gain for less readability.
> 
> What are you talking about? For me at least it is Nullable: https://github.com/PaperMC/Velocity/pull/1756/changes#diff-1cd097c1de92fff8c43bc33eb4f52405eb5396f9e8cc922a4502e4cad12e1f88R189

yes but you're treating it as a monotonically non-null value. we can just leave it as a
```java
private final Set<BossBar> bossBars = new HashSet<>();
```

and omit any null checks down the line. keeping this value null until it's used doesn't seem like it'll affect performance at all, and it makes the code that's using this field less readable (null -> either instantiate this set when modifying it, or assuming an empty set when querying it)

Timongcraft:
They don't though (https://github.com/PaperMC/Velocity/pull/1756/changes#diff-1cd097c1de92fff8c43bc33eb4f52405eb5396f9e8cc922a4502e4cad12e1f88R613), or am I missing something?

WouterGritter:
> They don't though (https://github.com/PaperMC/Velocity/pull/1756/changes#diff-1cd097c1de92fff8c43bc33eb4f52405eb5396f9e8cc922a4502e4cad12e1f88R613), or am I missing something?

oh yes my bad, its not monotonically non-null, but my point still holds. this can be a HashSet that's initiated once (at construct time of the object) and only updated/queried without null-checks. that'd get rid of null-checks in every method that's using this field

PR #1758, opened 2026-03-29
Fixing Hungarian translation

LitcherX:
Just translated the missing 3 lines that weren't yet


Timongcraft:
Localization is generally handled through Crowdin, see
https://papermc-io.crowdin.com/velocity

And escaping is only done by Crowdin and not necessary

LitcherX:
> Localization is generally handled through Crowdin, see https://papermc-io.crowdin.com/velocity
> 
> And escaping is only done by Crowdin and not necessary

Thanks, I didn't know that yet, I'll make sure to do it via crowdin!

PR #1743, opened 2026-03-18
Add more protocol safeguards

booky10:
Fixes checkstyle (broken by https://github.com/PaperMC/Velocity/commit/e8b64aa6c09edb0846dc06687307a4343637224e) and adds size checks to more config phase packets. Additionally, also implements some safeguards to ensure all packets during login/config states are known and the inbound packet queue doesn't grow too large in size.

Issue #1723, opened 2026-01-31
Velocity does not support re-entering the configuration phase from play.


mworzala:
### Expected Behavior

I expect the server to be able to send the client back to configuration and bring them back to play without issue.

### Actual Behavior

The server recieves an invalid packet and disconnects the client.

Error message from an unmodified vanilla server
```
[13:50:31] [Server thread/INFO]: notmattw (9f36238e-bee0-3e4e-9404-44efe6986143) lost connection: Internal Exception: io.netty.handler.codec.DecoderException: Received unknown packet id 12
```

Also error message from a Minestom server
```
java.lang.IllegalStateException: Packet id 0x1d isn't registered!
	at net.minestom.server.network.packet.PacketRegistry$PacketRegistryTemplate.packetInfo(PacketRegistry.java:482)
	at net.minestom.server.network.packet.PacketReading.readPayload(PacketReading.java:210)
	at net.minestom.server.network.packet.PacketReading.readFramedPacket(PacketReading.java:187)
```

### Steps to Reproduce

Start velocity pointing at a vanilla server with debug commands enabled, eg

```
java -DMC_DEBUG_ENABLED -DMC_DEBUG_DEV_COMMANDS -jar server.jar nogui
```

1. join the server through velocity
2. Force yourself into the configuration state, eg `debugconfig config <your_username>`
3. Force yourself back into the play state, eg `debugconfig unconfig <your_uuid>`
4. Observe above error message in server logs and disconnect
5. Become sad :-(

Note its easier to reproduce with some latency (eg 200ms), not sure if its required

### Plugin List

No plugins on either velocity or server (vanilla server)

### Velocity Version

running on commit 5320aae5d9a839e3a118d4ada8b1c0d1e901422c (from 26/1/26)

[13:58:54 INFO]: Velocity <unknown>
[13:58:54 INFO]: Copyright 2018-2026 Velocity Contributors. Velocity is licensed under the terms of the GNU General Public License v3.
[13:58:54 INFO]: PaperMC - GitHub

### Additional Information

I believe the issue is to do with the packet queue logic that occurs. When you enter config it starts queuing received play packets (most typically a tick end packet, 0xc). When the server sends a finish configuration packet, `MinecraftConnection#setState` is called which sends all those queued packets immediately. The result is the server receiving the play packets before the finish config ack from the client, and failing to parse them (it's still looking for play packets).

[...]

PR #1761, opened 2026-03-31
Add system property for packet direction validation

pop4959:
An unintentional side effect introduced as a byproduct of the additional validation added in #1743 (in particular, the check added for packet direction and connection state) is that it breaks servers running a proxy-in-proxy configuration. Granted of course this is a known grey area and this is not officially supported, however we have had numerous complaints from users claiming that after updating to Velocity for Minecraft 26.1, their server no longer works.

The aim is to keep this validation optional via an added system property, retaining support for proxy-in-proxy, as it currently works fine otherwise, without any other changes on the user-end.

If there was any specific reasoning or justification for the validation being added that I'm unaware of, please let me know, but for the most part my understanding is that it was added as a generic safeguard because this is an uncommon packet state which doesn't normally occur for a single proxy to server.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/protocol/netty/MinecraftDecoder.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/protocol/netty/MinecraftDecoder.java
@@ -22,6 +22,7 @@ import com.velocitypowered.api.network.ProtocolVersion;
 import com.velocitypowered.proxy.protocol.MinecraftPacket;
 import com.velocitypowered.proxy.protocol.ProtocolUtils;
 import com.velocitypowered.proxy.protocol.StateRegistry;
+import com.velocitypowered.proxy.util.VelocityProperties;
 import com.velocitypowered.proxy.util.except.QuietRuntimeException;
 import io.netty.buffer.ByteBuf;
 import io.netty.channel.ChannelHandlerContext;
@@ -37,7 +38,7 @@ public class MinecraftDecoder extends ChannelInboundHandlerAdapter {
   private static final QuietRuntimeException DECODE_FAILED =
       new QuietRuntimeException("A packet did not decode successfully (invalid data). For more "
           + "information, launch Velocity with -Dvelocity.packet-decode-logging=true to see more.");
-  private static final boolean DIRECTION_VALIDATION = Boolean.getBoolean("velocity.packet-direction-validation");
+  private static final boolean DIRECTION_VALIDATION = VelocityProperties.readBoolean("velocity.packet-direction-validation", true);
 
   private final ProtocolUtils.Direction direction;
   private StateRegistry state;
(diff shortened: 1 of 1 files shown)


proxy/src/main/java/com/velocitypowered/proxy/protocol/netty/MinecraftDecoder.java
```diff
@@ -37,6 +37,7 @@ public class MinecraftDecoder extends ChannelInboundHandlerAdapter {
   private static final QuietRuntimeException DECODE_FAILED =
       new QuietRuntimeException("A packet did not decode successfully (invalid data). For more "
           + "information, launch Velocity with -Dvelocity.packet-decode-logging=true to see more.");
+  private static final boolean DIRECTION_VALIDATION = Boolean.getBoolean("velocity.packet-direction-validation");
```
kashike:
```suggestion
  private static final boolean DIRECTION_VALIDATION = Boolean.parseBoolean(System.getProperty("velocity.packet-direction-validation", "true"));
```

kashike (changes requested):

proxy/src/main/java/com/velocitypowered/proxy/protocol/netty/MinecraftDecoder.java
RealBauHD: Velocity has a utility method for this (`VelocityProperties.readBoolean`)

pop4959: I've made the change to use `VelocityProperties` (no particular reason not to use this, although the existing property a few lines up used `Boolean.getBoolean` so I simply copied that), and set the default to true.

pop4959: I would advocate for the default not being true here though, as that would mean behavior is broken by default. Still better than not having any way to configure it at all however.

pop4959:
@booky10 Would love your input here if the original validation was necessary to for example fix some niche issue. If it's sensitive feel free to reach out privately.

I would love to preserve this validation if it serves some purpose, but definitely need more context as to what the default should be.

electronicboy:
The way that phase works kinda expects that we understand all of the packets used during that phase, and so not knowing stuff is a really weird situation, which can cause problems, such as not allowing a client to connect properly

The generic handling should work so this is probably not a huge concern, just, wouldn't be a setup we could provide support within

proxy/src/main/java/com/velocitypowered/proxy/protocol/netty/MinecraftDecoder.java
pop4959: After some more discussion, fine with keeping the default as suggested by kashike (defaulting to validation enabled). I don't have any further concerns, but let me know if the code as-is looks good or if there are any other changes code-style wise or naming wise

JayPeaSize:
Worth noting for those looking; this also fixes some issues with Forge/Fabric servers when connecting. That in combination with `velocity.max-plugin-message-payload-size` fixes any connection issues that have come up with the recent commits/changes.

electronicboy:
is the forge/fabric thing still an issue? The primary reported case this was trying to jump around was fixed by: https://github.com/PaperMC/Velocity/pull/1747

if so, would need means of reproduction

pop4959:
I believe otherwise this was just reported by #1723 but for our use case this is solved, so I'll go ahead and close this PR

Issue #1007, opened 2023-05-03
 A plugin tried to deny a command with signable component(s). This is not supported. Disconnecting player <nickname>


RealKomander:
Only a few players experience this issue, especially ones playing on TLauncher or versions 1.19.3+ (2 completely random factors, I know).

75% of the times they try joining, this message appears to them:
![image](https://user-images.githubusercontent.com/124439290/236015280-daba9abc-5254-4691-93ad-6b044557d395.png)

And this error is prompted in the console:
`[21:46:39 FATAL]: A plugin tried to deny a command with signable component(s). This is not supported. Disconnecting player deishoke`

Full logs are attached, sorry if it's not a velocity error, but it still happens without plugins.

[latest.log](https://github.com/PaperMC/Velocity/files/11389960/latest.log)

RealKomander:
I still don't understand how it blames a plugin, but this issue still happens without them. Unfortunately, I don't have time to test it right now, I'll send logs without plugins in the near future if you want them for some reason

electronicboy:
If this is happening without plugins, we'd need a log showing that, and
info of the command being executed

The only way message shows is if a command packet is denied by something in
the proxy when there is a signed component in the command

mi2sha:
это все из-за пупу каки

Nacioszeczek:
I'll be closing this as stale. If the issue still occurs please reply with the requested logs and it'll be reopened.

A248:
This is not stale. It is a real issue that is hard to track down, and it has affected multiple users.

Over the course of 2 years or so, 6 users have reported this problem to me.

This is not a matter of plugins or no plugins. A plugin that merely registers a command shouldn't cause players to be disconnected.

**A potential lead**
I received the first report of this bug in July 2024. A user was having trouble on Velocity build `#402`.

They performed their own debugging, and as soon as they downgraded to build `#400`, the problems stopped.

The user concluded that code changes surrounding commands with signable component caused this error. Due to a lack of communication neither of us filed a bug report, but this bug surfaced rarely and has been a shadowy menace since. It seems to be the product of nondeterministic factors as I can never reproduce it, but it comes back repeatedly to haunt users of my plugin (LibertyBans) which registers the /ban command.

**Logs**
Today I obtained logs from a user experiencing this bug.

[2026-04-11-6.log](https://github.com/user-attachments/files/26649046/2026-04-11-6.log)

electronicboy:
[...]

PR #1773, opened 2026-04-12
Improve debugging somewhat for signable command-related errors

A248:
Help with #1007


WouterGritter:
It would probably make more sense to keep `CommandHandler` an interface and give each implementing class it's own `logger` field (though that being said this class does contain a suspiciously large amount of default implemented methods). As the log messages are exactly the same, it would be helpful to see which class is logging the messages.

PR #1775, opened 2026-04-17
feat: Make version clickable in velocity info command

Emilxyz:
Make it easier for users to provide us the full version string when asked to in support channels, if they're running the command in game instead of in console.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/command/builtin/VelocityCommand.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/command/builtin/VelocityCommand.java
@@ -161,14 +161,17 @@ public final class VelocityCommand {
           .decoration(TextDecoration.BOLD, true)
           .color(VELOCITY_COLOR)
           .append(Component.text()
-                  .content(version.getVersion())
-                  .decoration(TextDecoration.BOLD, false))
+              .content(version.getVersion())
+              .decoration(TextDecoration.BOLD, false))
+          .hoverEvent(Component.translatable("velocity.command.version-offer-copy-version"))
+          .clickEvent(ClickEvent.copyToClipboard(version.getName() + " "
+              + version.getVersion()))
           .build();
       final Component copyright = Component
           .translatable("velocity.command.version-copyright",
               Argument.string("vendor", version.getVendor()),
-                  Argument.string("name", version.getName()),
-                  Argument.component("year", Component.text(LocalDate.now().getYear())));
+              Argument.string("name", version.getName()),
+              Argument.component("year", Component.text(LocalDate.now().getYear())));
       source.sendMessage(velocity);
       source.sendMessage(copyright);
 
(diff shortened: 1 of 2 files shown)


jonesdevelopment (approved):

kennytv (approved):

WouterGritter:
This is not an issue with this PR per-se, but there are no migrations in place for `messages.properties` or any similar file for a different languages. Updating velocity without removing `messages.properties` or manually adding this key will make the hover text show up as the translation key. Is this something that should be considered in the future? Message files migrations for adding keys, just like there's a migration system in-place for `velocity.toml`?

proxy/src/main/java/com/velocitypowered/proxy/command/builtin/VelocityCommand.java
```diff
@@ -161,14 +161,17 @@ public int run(final CommandContext<CommandSource> context) {
           .decoration(TextDecoration.BOLD, true)
           .color(VELOCITY_COLOR)
           .append(Component.text()
-                  .content(version.getVersion())
-                  .decoration(TextDecoration.BOLD, false))
+              .content(version.getVersion())
+              .decoration(TextDecoration.BOLD, false))
+          .hoverEvent(Component.translatable("velocity.command.version-offer-copy-version"))
+          .clickEvent(ClickEvent.copyToClipboard(version.getName() + " "
+              + version.getVersion()))
```
WouterGritter, line 168: would a

```java
private String getCopyableVersionString() {
  return version.getName() + " " + version.getVersion();
}
```

be more maintainable here?

Emilxyz: Could you explain more where you think this method should be added? I don't really see how this would improve maintainability.

Emilxyz:
> This is not an issue with this PR per-se, but there are no migrations in place for `messages.properties` or any similar file for a different languages. Updating velocity without removing `messages.properties` or manually adding this key will make the hover text show up as the translation key. Is this something that should be considered in the future? Message files migrations for adding keys, just like there's a migration system in-place for `velocity.toml`?

This problem also came to my attention when working on this pr. However, solving it is way out of scope of this pr and should be discussed in a separate issue.

R00tB33rMan (approved):

proxy/src/main/java/com/velocitypowered/proxy/command/builtin/VelocityCommand.java
WouterGritter: In the same `VelocityCommand.Info` record, right above or below `run()`.

As to why this should (could) be a separate helper method, it's not `run()`'s responsibility to format this string. Having it be a separate method is clearer and easier to find/modify at a later date.

Though that makes me think, maybe this better as a translatable message in `messages.properties`? That way this would be a `Component.translatable()` call with two arguments (name + version) and it's the translatable component responsibility to format this properly.

Uhuli (approved):

proxy/src/main/java/com/velocitypowered/proxy/command/builtin/VelocityCommand.java
Emilxyz: > As to why this should (could) be a separate helper method, it's not run()'s responsibility to format this string. Having it be a separate method is clearer and easier to find/modify at a later date.

Thanks for your input, however, I don't think this really matters that much, so I'm gonna leave this as-is for now.

> Though that makes me think, maybe this better as a translatable message in messages.properties?

As to making this message translatable, this doesn't make a lot of sense in my opinion. The translation string would essentially be `{0} {1}`, so I think this is also fine as-is.

WouterGritter (approved):

PR #1780, opened 2026-04-24
Add unix domain socket support

Beaness:
Deprecates ``ServerInfo#getAddress`` and errors if it is a unix domain socket, current velocity proxies setups should remain working. 
Adds a new method ``ServerInfo#getSocketAddress`` as alternative.

Based on https://github.com/PaperMC/Velocity/pull/991, https://github.com/PaperMC/Velocity/pull/422, https://github.com/PaperMC/Velocity/commit/b00389029f4c11ad34898b5298625f4fcfc03de5, https://github.com/PaperMC/Velocity/commit/b00389029f4c11ad34898b5298625f4fcfc03de5


proxy/src/main/java/com/velocitypowered/proxy/network/TransportType.java
```diff
@@ -54,35 +64,49 @@ public enum TransportType {
...
       EpollSocketChannel::new,
       EpollDatagramChannel::new,
+      EpollServerDomainSocketChannel::new,
+      EpollDomainSocketChannel::new,
       EpollIoHandler::newFactory),
   KQUEUE("kqueue", KQueueServerSocketChannel::new,
       KQueueSocketChannel::new,
       KQueueDatagramChannel::new,
+      null,
+      null,
```
calcastor:
```suggestion
      KQueueServerDomainSocketChannel::new,
      KQueueDomainSocketChannel::new,
```

WouterGritter:
I wonder what your usecase for this feature is? I think the 'modern' go-to solution for a sort of 'anonymous'/intermediate proxy not exposed to the world would be to use Docker and Docker's networking.

SirYwell:
> I wonder what your usecase for this feature is? I think the 'modern' go-to solution for a sort of 'anonymous'/intermediate proxy not exposed to the world would be to use Docker and Docker's networking.

Using Docker still allows using Unix Domain Sockets, and the performance will be (slightly) better most likely.

Beaness:
> I wonder what your usecase for this feature is? I think the 'modern' go-to solution for a sort of 'anonymous'/intermediate proxy not exposed to the world would be to use Docker and Docker's networking.

I think most people would use this for better performance when having multiple servers on the same machine. As you said there are different (or even better) alternatives for networking security.

proxy/src/main/java/com/velocitypowered/proxy/network/TransportType.java
```diff
@@ -91,6 +117,40 @@ public String toString() {
...
+   * Returns the channel factory to use to listen on the specified socket address.
+   *
+   * @param address the address we want to listen on
+   * @return the channel factory
+   */
+  public ChannelFactory<? extends ServerChannel> getServerChannelFactory(SocketAddress address) {
+    if (address instanceof DomainSocketAddress) {
+      if (this.domainServerSocketChannelFactory == null) {
+        throw new IllegalArgumentException(
+                "Domain sockets are not available for non-Linux platforms");
```
A248: Do you mean non-Unix? kqueue runs on Unix machines like BSD and MacOS.

SirYwell: Non-Unix systems might support unix domain sockets as well (https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/)

A248: This PR doesn't support them, however.

Beaness: Nice catch

Uhuli (approved):

PR #1781, opened 2026-04-24
Small optimization to prevent blocking netty threads on UUID.randomUUID()

Beaness:
At higher playercounts a lot of player tablist packets can be sent concurrently causing some small thread congestion (across all netty threads). This is a small optimization to prevent this congestion by just generating the uuids using threadlocal random.


electronicboy:
invalid UUIDs is generally a no-go in my book, using a non-secure random also introduces some risks around duplicate UUIDs being generated

Beaness:
> invalid UUIDs is generally a no-go in my book, using a non-secure random also introduces some risks around duplicate UUIDs being generated

Changed to a valid uuid v4, I don't think the duplicate uuid is a real issue as there are so many random bits I don't think it will ever collide

WouterGritter:
Have you measured the performance of java's random UUID generation and this proprietary implementation? Is it even worth it to roll our own, aside from possible security/duplicate UUID concerns?

Beaness:
> Have you measured the performance of java's random UUID generation and this proprietary implementation? Is it even worth it to roll our own, aside from possible security/duplicate UUID concerns?

Yes at higher player counts there is thread congestion on netty threads (this worsens with the higher the threads and all players getting the same player info packets at the same time)

The security UUID concern should not be applicable here as they are only used as identifiers internally.
I kept the UUID.randomUUID() in click callbacks to ensure the security.

WouterGritter:
Also, how "unique" do the UUIDs have to be in the tablist? Could we increment the lower or higher long to keep the UUIDs "locally" unique per session or per tablist? That would speed this up a lot. Kind of similar how anonymous player entries in the server player list packets are expected to have a uuid of `new UUID(0, 0)`. Though that does abuse the hell out of the _universally_ unique IDs here.

Beaness:
> Also, how "unique" do the UUIDs have to be in the tablist? Could we increment the lower or higher long to keep the UUIDs "locally" unique per session or per tablist? That would speed this up a lot. Kind of similar how anonymous player entries in the server player list packets are expected to have a uuid of `new UUID(0, 0)`. Though that does abuse the hell out of the _universally_ unique IDs here.

This should technically also work for the legacy tablist implementations. The uuid is used to have a unique id for a player name as UUIDs aren't send in legacy with the tablist packets.

WouterGritter:
https://github.com/f4b6a3/uuid-creator

> UUID Version 7: the Unix Epoch time-based UUID specified in RFC 9562.

Would be a perfect fit here (`UuidCreator#getTimeOrderedWithRandom` / `UuidCreator#getTimeOrderedEpoch[Fast]`).

Beaness:
> https://github.com/f4b6a3/uuid-creator
> 
> > UUID Version 7: the Unix Epoch time-based UUID specified in RFC 9562.
> 
> Would be a perfect fit here (`UuidCreator#getTimeOrderedWithRandom` / `UuidCreator#getTimeOrderedEpoch[Fast]`).

Is it really needed? The current implementation keeps the v4 validity and the probability of actually having a collision is astronomically low.

WouterGritter:
I think relying on an external library for this rather than manually fiddling with bits in our own FastRandomUuid is favorable. Its current name also suggests its perfectly safe to use in other cases where a random uuid is needed, which its really not due to relying on a non-secure random generator.

I think a better solution would be to get rid of this class and move it to VelocityTabListLegacy as a private static method (does the bossbar class actually re-generate the random uuids a bunch of time?)

Beaness:
> I think relying on an external library for this rather than manually fiddling with bits in our own FastRandomUuid is favorable. Its current name also suggests its perfectly safe to use in other cases where a random uuid is needed, which its really not due to relying on a non-secure random generator.
> 
> I think a better solution would be to get rid of this class and move it to VelocityTabListLegacy as a private static method (does the bossbar class actually re-generate the random uuids a bunch of time?)

The method is now only present in VelocityTabListLegacy. I don't think an additional library is really needed, the collision chances are basically zero and the uuid is structurally valid. It's also not a usable "uuid" as it's used to spoof a tablist entry for temporary binding to a tab name.

VidTu:
On the topic of v4-vs-v7, if Mojang [uses it](https://mcsrc.dev/1/26.1.2/net/minecraft/util/Mth#L320) to [generate Entity UUIDs](https://mcsrc.dev/1/26.1.2/net/minecraft/world/entity/Entity#L295), then it's alright enough.

On the performance topic, are the gains actually significant?

VidTu:
I actually got curious and wrote a [small benchmark](https://gist.github.com/VidTu/533ef124727e9144562e6f0de95954c8) for that and ran on my (admittedly ancient) desktop PC on both Windows (everyday usage) and Linux (installation ISO without background services):
```
> Windows
Benchmark                    Mode  Cnt       Score      Error   Units
UuidBenchmark.uuidInsecure  thrpt    6  151409.421 ± 1504.844  ops/ms
UuidBenchmark.uuidSecure    thrpt    6     698.752 ±   10.978  ops/ms

> Linux
Benchmark                    Mode  Cnt       Score     Error   Units
UuidBenchmark.uuidInsecure  thrpt    6  156989.602 ± 593.248  ops/ms
UuidBenchmark.uuidSecure    thrpt    6    2846.491 ±  35.525  ops/ms
```

So yeah, there are good magnitude-order speedups (less good on Linux, probably due to better reseeding from the kernel CSRNG). However, the baseline of 700 (or 2800) UUIDs/ms is not something that needs extreme urgent optimization.

SirYwell:
`UUID.randomUUID()` is known to scale rather bad with more threads (see https://github.com/openjdk/jdk/pull/14135), so this might make sense. If secure random should be used, an approach similar to that in the linked PR might be possible as well.

VidTu:
Fair point, I completely neglected the multithreaded-ness of Velocity and cas/mutex slowdown of Java rng/csrng implementations. The gains may be even better in a multithreaded environment

WouterGritter:
The referenced library also provides benchmarks: https://github.com/f4b6a3/uuid-creator/wiki/5.0.-Benchmark

Edit: This PR implements the exact behavior of `UuidCreator.getRandomBasedFast()`. Looking at these benchmarks (this method itself isn't directly measured, but `altGUIDv4` comes the closest in terms of implementation, also using two random longs from `ThreadLocalRandom.current()`), this should be about 130x faster than `UUID.randomUUID()`

kennytv (approved):

proxy/src/main/java/com/velocitypowered/proxy/tablist/VelocityTabListLegacy.java
```diff
@@ -148,9 +149,16 @@ public TabListEntry buildEntry(GameProfile profile, @Nullable Component displayN
     return new VelocityTabListEntryLegacy(this, profile, displayName, latency, gameMode);
   }
 
-  @Override
-  public TabListEntry buildEntry(GameProfile profile, @Nullable Component displayName, int latency,
-                                 int gameMode, @Nullable ChatSession chatSession, boolean listed, int listOrder, boolean showHat) {
-    return new VelocityTabListEntryLegacy(this, profile, displayName, latency, gameMode);
```
WouterGritter, line 154: Is dropping this method intentional? `KeyedVelocityTabList#buildEntry` (which this was previously overriding) just delegated this to one of the other `buildEntry()`'s that _is_ still overridden now (so right now this is safe), but this behavior might change in the future.

If it does, a change in `KeyedVelocityTabList` may make `VelocityTabListLegacy` return the wrong type here.

proxy/src/main/java/com/velocitypowered/proxy/tablist/VelocityTabListLegacy.java
```diff
@@ -148,9 +149,16 @@ public TabListEntry buildEntry(GameProfile profile, @Nullable Component displayN
...
+  /**
+   * Generates a random UUID v4 using {@link ThreadLocalRandom}. The result is a structurally valid
+   * UUID v4 but is not cryptographically secure
+   *
+   * @return a new random {@link UUID}
+   */
+  private static UUID generateInsecureRandomUuid() {
+    ThreadLocalRandom random = ThreadLocalRandom.current();
+    long msb = (random.nextLong() & 0xffffffffffff0fffL) | 0x0000000000004000L; // version 4
+    long lsb = (random.nextLong() & 0x3fffffffffffffffL) | 0x8000000000000000L; // IETF variant
```
WouterGritter, line 167: Could be overkill here, but we might consider "seeding" (xorring) the nextLong's with a value from SecureRandom, example & explanation from the referenced library: https://github.com/f4b6a3/uuid-creator/blob/master/src/main/java/com/github/f4b6a3/uuid/alt/GUID.java#L796

proxy/src/main/java/com/velocitypowered/proxy/tablist/VelocityTabListLegacy.java
Beaness: Not intentional, fixed.

proxy/src/main/java/com/velocitypowered/proxy/tablist/VelocityTabListLegacy.java
Beaness: I think this is a bit overkill for this usecase. The uuid is only used for a "fake" mapping to the name, theres no real use besides it so it would never "cross" JVMs.

WouterGritter (approved):

R00tB33rMan (approved):

proxy/src/main/java/com/velocitypowered/proxy/tablist/VelocityTabListLegacy.java
WouterGritter: I agree, worth considering if this is ever extracted into a utility class, for other callers that need non-secure random UUIDs.

PR #1783, opened 2026-04-28
Fix handshake hostname length limit and plugin message drop during join

### LINKED ISSUE #1766: Plugin messages silently dropped during player join when getConnectedServer() is null
### Expected Behavior

Plugin messages sent on registered channels during the player join process should be forwarded to the backend server via `getConnectionInFlight()` when `getConnectedServer()` is null, the same way Forge Legacy handshake messages are. Messages should not be silently discarded.
 

### Actual Behavior

`ClientPlaySessionHandler.handle(PluginMessagePacket)` at lines 303 to 310 only falls back to `getConnectionInFlight()` when the channel is `LegacyForgeConstants.FORGE_LEGACY_HANDSHAKE_CHANNEL`. For all other channels, when `getConnectedServer()` returns null during the join window, `serverConn` is set to null and the handler silently skips the message. No warning, no error, no exception, etc.

```java
VelocityServerConnection serverConn =
    (player.getConnectedServer() == null
        && packet.getChannel().equals(
        LegacyForgeConstants.FORGE_LEGACY_HANDSHAKE_CHANNEL))
        ? player.getConnectionInFlight() : player.getConnectedServer();
```
 
The null window exists between `player.setConnectedServer(null)` at `TransitionSessionHandler` line 100 and `player.setConnectedServer(serverConn)` at line 140 (inside a `thenRunAsync` callback). During this window the client connection continues processing packets, but any non-Forge plugin message is dropped.
 
This is the same null window documented in issue #569, where `ServerConnectedEvent` is described as "fatally flawed" because `Player#getCurrentServer()` returns null during it.

### Steps to Reproduce

1. Register a custom plugin message channel on Velocity
2. Have a plugin or client mod send a plugin message on that channel during the join process, while `getConnectedServer()` is still null (e.g. a fast async callback such as a cached response completing during the join window)
3. Observe that the message is silently dropped with no log output
 
The bug is intermittent. It only triggers when the message arrives during the null window between `setConnectedServer(null)` and `setConnectedServer(serverConn)`. A delayed retry reliably succeeds.

### Plugin List

N/A - reproducible with any plugin that sends messages on a custom channel during the join process.
 
Discovered via Geyser + Floodgate. When Floodgate's websocket subscriber fails to connect to the Global API (e.g. temporary unavailability during startup), Geyser's `FloodgateSkinUploader` falls back to sending skin data via plugin message on the `floodgate:skin` channel. If the Global API responds quickly (cached skin), this message arrives during the null window and is silently dropped. The player appears as default skins to Java players.

### Velocity Version

[18:08:06 INFO]: Velocity 3.5.0-SNAPSHOT (git-ab99bde9-b585)
[18:08:06 INFO]: Copyright 2018-2026 Velocity Contributors. Velocity is licensed under the terms of the GNU General Public License v3.
[18:08:06 INFO]: PaperMC - GitHub

### Additional Information

**Proposed fix:** Remove the Forge channel gate on the `getConnectionInFlight()` fallback:
 
```java
VelocityServerConnection serverConn = player.getConnectedServer();
if (serverConn == null) {
    serverConn = player.getConnectionInFlight();
}
```
 
This is safe because the downstream code already guards against edge cases:
- Backend state check (line 311) rejects writes if the backend isn't in PLAY
- Phase completeness checks (lines 352-353) queue messages if phases aren't complete
- The in-flight connection's backend is already in `StateRegistry.PLAY` (installed via `LoginSessionHandler` line 158 / `ConfigSessionHandler` line 252)

SendableMetatype:
- **Remove restrictive handshake hostname length limit:** The 261-character `MAXIMUM_HOSTNAME_LENGTH` in `HandshakePacket.decode()` silently disconnects clients when enough plugins inject forwarding data into the hostname field that the combined length exceeds the limit. `readString(buf)` now uses the default 65536 limit, matching every other string field and aligning with Paper/Spigot's equivalent fix (`readUtf(255)` to `readUtf(Short.MAX_VALUE)` in `ClientIntentionPacket`).

- **Fix plugin messages silently dropped during player join:** Fixes #1766

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientPlaySessionHandler.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientPlaySessionHandler.java
@@ -38,7 +38,6 @@ import com.velocitypowered.proxy.connection.MinecraftSessionHandler;
 import com.velocitypowered.proxy.connection.backend.BackendConnectionPhases;
 import com.velocitypowered.proxy.connection.backend.BungeeCordMessageResponder;
 import com.velocitypowered.proxy.connection.backend.VelocityServerConnection;
-import com.velocitypowered.proxy.connection.forge.legacy.LegacyForgeConstants;
 import com.velocitypowered.proxy.connection.player.resourcepack.ResourcePackResponseBundle;
 import com.velocitypowered.proxy.protocol.MinecraftPacket;
 import com.velocitypowered.proxy.protocol.StateRegistry;
@@ -357,13 +356,9 @@ public class ClientPlaySessionHandler implements MinecraftSessionHandler {
 
   @Override
   public boolean handle(PluginMessagePacket packet) {
-    // Handling edge case when packet with FML client handshake (state COMPLETE)
-    // arrives after JoinGame packet from destination server
-    VelocityServerConnection serverConn =
-        (player.getConnectedServer() == null
-            && packet.getChannel().equals(
-            LegacyForgeConstants.FORGE_LEGACY_HANDSHAKE_CHANNEL))
-            ? player.getConnectionInFlight() : player.getConnectedServer();
+    final VelocityServerConnection connected = player.getConnectedServer();
+    final VelocityServerConnection serverConn =
+        connected != null ? connected : player.getConnectionInFlight();
 
     MinecraftConnection backendConn = serverConn != null ? serverConn.getConnection() : null;
     if (serverConn != null && backendConn != null) {
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientPlaySessionHandler.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientPlaySessionHandler.java
@@ -37,7 +37,6 @@ import com.velocitypowered.proxy.connection.MinecraftSessionHandler;
 import com.velocitypowered.proxy.connection.backend.BackendConnectionPhases;
 import com.velocitypowered.proxy.connection.backend.BungeeCordMessageResponder;
 import com.velocitypowered.proxy.connection.backend.VelocityServerConnection;
-import com.velocitypowered.proxy.connection.forge.legacy.LegacyForgeConstants;
 import com.velocitypowered.proxy.connection.player.resourcepack.ResourcePackResponseBundle;
 import com.velocitypowered.proxy.protocol.MinecraftPacket;
 import com.velocitypowered.proxy.protocol.StateRegistry;
@@ -341,13 +340,9 @@ public class ClientPlaySessionHandler implements MinecraftSessionHandler {
 
   @Override
   public boolean handle(PluginMessagePacket packet) {
-    // Handling edge case when packet with FML client handshake (state COMPLETE)
-    // arrives after JoinGame packet from destination server
-    VelocityServerConnection serverConn =
-        (player.getConnectedServer() == null
-            && packet.getChannel().equals(
-            LegacyForgeConstants.FORGE_LEGACY_HANDSHAKE_CHANNEL))
-            ? player.getConnectionInFlight() : player.getConnectedServer();
+    final VelocityServerConnection connected = player.getConnectedServer();
+    final VelocityServerConnection serverConn =
+        connected != null ? connected : player.getConnectionInFlight();
 
     MinecraftConnection backendConn = serverConn != null ? serverConn.getConnection() : null;
     if (serverConn != null && backendConn != null) {
--- a/proxy/src/main/java/com/velocitypowered/proxy/protocol/netty/LegacyPingDecoder.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/protocol/netty/LegacyPingDecoder.java
@@ -28,6 +28,8 @@ import io.netty.handler.codec.ByteToMessageDecoder;
 import java.net.InetSocketAddress;
 import java.nio.charset.StandardCharsets;
 import java.util.List;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
 
 /**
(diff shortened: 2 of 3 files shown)


WouterGritter:
I think returning -1 in `HandshakePacket#decodeExpectedMaxLength` isn't desirable. No authentication is required to send this packet, anyone can open TCP connections & spam this packet possibly opening up DOS attacks again that were mitigated by some recent commits.

SendableMetatype:
> I think returning -1 in `HandshakePacket#decodeExpectedMaxLength` isn't desirable. No authentication is required to send this packet, anyone can open TCP connections & spam this packet possibly opening up DOS attacks again that were mitigated by some recent commits.

Good point. I was assuming the VarInt limit handled it, but 2MB is obviously still an exploitable size. The hostname is now capped at Short.MAX_VALUE characters, which matches Paper's readUtf(Short.MAX_VALUE) in ClientIntentionPacket.

WouterGritter:
See https://github.com/GemstoneGG/Velocity-CTD/pull/885

Is is true that this PR adds support for [SendableMetatype/EduGeyser](https://github.com/SendableMetatype/EduGeyser)? Is there any other reason why this PR is needed?

SendableMetatype:
These fixes have nothing strictly to do with EduGeyser support. They fix general Velocity bugs that any plugin setup can trigger under the right conditions. I just happened to run into them and bother diagnosing them.

electronicboy:
I mean, this looks like two distinct set of changes

1) The plugin message thing, this would be a bug

2) the hostname length limit, this isn't a bug fix, at least not as we consider it given that we cater to the vanilla protocol, increasing it beyond the vanilla length limit on the incoming side is pretty breaky behaviorally

THEMPGUYAlt:
> See [GemstoneGG#885](https://github.com/GemstoneGG/Velocity-CTD/pull/885)
> 
> Is is true that this PR adds support for [SendableMetatype/EduGeyser](https://github.com/SendableMetatype/EduGeyser)? Is there any other reason why this PR is needed?

Just cloning into velocity ctd before velocity, i fixed the title too to fix the misleading stuff

WouterGritter:
> increasing it beyond the vanilla length limit on the incoming side is pretty breaky behaviorally

I'd consider putting it behind a system property, with vanilla defaults. That could at least give the user the option to have support for things like EduGeyser, just like we have with many other limits.

SendableMetatype:
> I mean, this looks like two distinct set of changes
> 
> 1. The plugin message thing, this would be a bug
> 2. the hostname length limit, this isn't a bug fix, at least not as we consider it given that we cater to the vanilla protocol, increasing it beyond the vanilla length limit on the incoming side is pretty breaky behaviorally

Regarding the hostname limit: Paper/Spigot made the equivalent change years ago (readUtf(255) to readUtf(Short.MAX_VALUE) in ClientIntentionPacket) for the same reason - plugins inject forwarding data into the hostname field, which is standard practice in the BungeeCord forwarding model. Velocity's own encode() already writes hostnames with no length cap, so the decode side is just being made consistent. The limit is now Short.MAX_VALUE to match Paper, and decodeExpectedMaxLength is kept in place for DoS protection on the unauthenticated packet.

electronicboy:
Spigot made that change year ago purely for their forwarding mechanism and nothing else, it breaks compatability with the vanilla protocol in a way which prevents the proxy from being able to connect to vanilla servers entirely. The decode/encode inconsistency here is just a sad reality of how that mechanism is intended to work to keep compat with legacy forwarding.

That data inside of that packet is somewhat trusted due to the vanilla protocol limit on the decode side usually getting in the way of doing anything fancy there, effectively removing that limit stands out as a potential security risk at the very least; it's definetly not something I feel should just be changed overnight

SendableMetatype:
The incoming decode limit doesn't affect vanilla backend compatibility. Velocity controls what it sends to backends based on forwarding mode, and the encode side already has no length cap,  so anything that would break from a longer incoming hostname is already breakable through Velocity's own outgoing path. The decode/encode inconsistency doesn't protect anything if the encode side is already uncapped.

The security concern is fair, which is why the limit is Short.MAX_VALUE with decodeExpectedMaxLength kept in place, not removed entirely.

WouterGritter:
In my opinion it does not need to matter if upping the limit breaks vanilla backend compatibility if we put this limit behind a system property.

It's already possible to completely break vanilla (or any) connections with system properties by lowering packet size limits. Adding another system property that's a NOP by default, but allows a server admin to _break vanilla connections_ (and fix support for EduGeyser) by increasing this limit isn't something new in that regard.

SendableMetatype:
A system property that defaults to the broken behavior means every affected server admin has to independently diagnose a silent disconnect with no error output and somehow discover the flag. That's the same problem as the current bug, just with an extra step. The default should handle legitimate use cases without manual intervention. And to be clear, this has nothing to do with EduGeyser support. EduGeyser typically works fine without this change. These are general Velocity bugs that any plugin setup can hit under the right conditions.

electronicboy:
I'm not understanding how allowing clients to send arbitary data inside of the handshake packet is 
1) for plugins
2) a legitimate usecase

Mojang already offers a supported mechanism for plugins to sync arbitary data during early connections to the client, API here is lacking but I would much rather discuss proper solutions to that rather than having to maintain spec violations

SendableMetatype:
The change isn't about allowing external clients to send arbitrary data. The longer hostnames come from plugins running inside Velocity, sending through internal channels like LocalChannel. External clients still send normal short hostnames.

Some plugins use the hostname field to forward data before authentication. This is the same BungeeCord forwarding pattern that Velocity explicitly supports,  the limit was just set for DNS + Forge markers without accounting for forwarding payloads.
It's obviously not the intended use of the field, but I'm not the one introducing or picking that pattern.

THEMPGUYAlt:
Please push this, this would allow more plugins to work.

PR #1786, opened 2026-05-04
Add decompressed-bytes-per-second rate limit, update packet limiter defaults

WouterGritter:
The `packets-per-second` limit was introduced to mitigate compression bomb attacks, where an attacker floods the proxy with small packets that decompress to the maximum allowed size. It's the wrong metric for this, though: 500 packets/s at 2 MiB each is 1 GiB/s of decompressed data, which kills the proxy just as effectively. At the same time, legitimate clients with chatty mods (e.g. heavy plugin-message traffic) were getting disconnected despite consuming far less than any attacker would.

This PR adds a `decompressed-bytes-per-second` limit to `[packet-limiter]`, applied after decompression in `MinecraftCompressDecoder`. This directly measures what was causing OOM issues, rather than using a proxy metric. With a default of 5 MiB/s over a 7-second window (35 MiB of headroom per window), legitimate clients should never get close to this limit, while any compression bomb is caught well before it causes memory pressure.

Default config changes (config version `2.7` -> `2.8`):
- `packets-per-second`: 500 -> -1 (disabled)
- `bytes-per-second`: unchanged (-1)
- `decompressed-bytes-per-second`: new field, defaults to 5242880 (5 MiB/s)

A migration for this has also been added, overriding the previous defaults if the config version was `2.7` or older, including comments explaining each metric.

Ported from GemstoneGG/Velocity-CTD#871.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
@@ -31,6 +31,7 @@ import com.velocitypowered.proxy.config.migration.ForwardingMigration;
 import com.velocitypowered.proxy.config.migration.KeyAuthenticationMigration;
 import com.velocitypowered.proxy.config.migration.MiniMessageTranslationsMigration;
 import com.velocitypowered.proxy.config.migration.MotdMigration;
+import com.velocitypowered.proxy.config.migration.PacketLimiterMigration;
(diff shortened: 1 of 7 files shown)


R00tB33rMan (approved):

WouterGritter:
This PR also reverts ab8333d6944e844fb4152b561a0bee460a081216, see #1792. This commit is no longer needed as this PR resolves any threat of a decompression attack, this commit would kill legitimate connections way too soon.

electronicboy (approved):

PR #1787, opened 2026-05-07
Convert PlayerChooseInitialServerEvent into a resulted event

electronicboy:


proxy/src/main/java/com/velocitypowered/proxy/connection/client/AuthSessionHandler.java
```diff
@@ -262,14 +262,16 @@ private CompletableFuture<Void> connectToInitialServer(ConnectedPlayer player) {
...
-        player.disconnect0(
-            Component.translatable("velocity.error.no-available-servers", NamedTextColor.RED),
-            true);
+      final PlayerChooseInitialServerEvent.Result result = event.getResult();
+      if (event.getResult() instanceof PlayerChooseInitialServerEvent.Denied denied) {
+        player.disconnect0(Objects.requireNonNullElse(denied.reason(),
+          Component.translatable("velocity.error.no-available-servers", NamedTextColor.RED)),
+          true);
+        //noinspection UnnecessaryReturnStatement
         return;
```
WouterGritter, line 271:
```java
//noinspection UnnecessaryReturnStatement
return;
```

Seems unnecessary? I don't see why you would want a return here?
Maybe one of those fancy new switch "instanceof" statements would be better here anyway ([JEP 441](https://openjdk.org/jeps/441)), throwing on default (unknown/unexpected result type)

electronicboy: I'd originally wrote this without the else statement and then remembered I needed it so I could for the if anyways. I do know that IJ itself tended to get upset with shorter case switches and so I did tend to avoid them for stuff like this

api/src/main/java/com/velocitypowered/api/event/player/PlayerChooseInitialServerEvent.java
```diff
@@ -51,14 +60,62 @@ public Optional<RegisteredServer> getInitialServer() {
...
+    }
+
+    @Override
+    public boolean isAllowed() {
+      return true;
+    }
+  }
+
+  public record Denied(@Nullable Component reason) implements Result {
+    public static @NotNull Result of(@Nullable Component reason) {
```
WouterGritter, line 108: Inconsistent formatting (no whitespace after class declaration, `Allowed` does have this)

Is it an idea to force this with checkstyle? Whitespace after class declaration, or newline checking in general, idk if we currently have this. Iirc this is quite inconsistent in the codebase

proxy/src/main/java/com/velocitypowered/proxy/connection/client/AuthSessionHandler.java
WouterGritter: Fair enough, still I think line 270 and 271 can be deleted

api/src/main/java/com/velocitypowered/api/event/player/PlayerChooseInitialServerEvent.java
electronicboy: Yea, we don't really have a hard rule on this and it's pretty easy to miss when refactoring during working on something, need to update checkstyle and formalise rules for this stuff

PR #1789, opened 2026-05-08
Fix malformed NBT/JSON decode crashes in player list + server data packets

vishv843:
**Problem**
Fuzzing found two reproducible decode-time crashes:
- NegativeArraySizeException in ProtocolUtils.readBinaryTag(...) (bubbles up from Adventure NBT parsing).
- JsonSyntaxException/MalformedJsonException in LegacyPlayerListItemPacket.readOptionalComponent(...).
These exceptions were unhandled and could crash/abort decode paths instead of failing gracefully.

This vulnerability can be exploited by a DoS attack and this fix will improve code robustness

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/protocol/ProtocolUtils.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/protocol/ProtocolUtils.java
@@ -512,6 +512,9 @@ public enum ProtocolUtils {
     } catch (IOException thrown) {
       throw new DecoderException("Unable to parse BinaryTag, full error: " + thrown.getMessage());
     }
+    catch (NegativeArraySizeException e) {
+      throw new DecoderException("BinaryTag had a negative length, likely due to malformed input", e);
+    }
   }
 
   /**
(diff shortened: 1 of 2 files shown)


WouterGritter:
There are a couple issues with this PR:

- > This vulnerability can be exploited by a DoS attack and this fix will improve code robustness
  
  This path is not exploitable client-side. Both cases where the exceptions are caught are clientbound-only (readBinaryTag() isnt used serverbound). This could only be exploited by a malicious backend server.

- `readBinaryTag()` can throw way more than just the `NegativeArraySizeException` (most obvious example is an index out of bounds in ProtocolUtils#readBinaryTag line 506; `BINARY_TAG_TYPES[buf.readByte()]`)
- The server won't crash anyways as the uncaught exceptions will be caught upstream (`MinecraftDecoder#tryDecode`), and kill the connection instead of killing/DOSing the proxy


> Fuzzing found two reproducible decode-time crashes

Are you able to share the steps to actually reproduce a full on proxy crash by sending some malformed packets from the client?

PR #1798, opened 2026-05-21
Always close HttpClient (even on exception)

WouterGritter:


R00tB33rMan (approved):

proxy/src/main/java/com/velocitypowered/proxy/connection/client/InitialLoginSessionHandler.java
4drian3d: Now that we're using Java 21 throughout the project, I think we should take this opportunity to remove this try-catch block

4drian3d (approved):

electronicboy (approved):

PR #1799, opened 2026-05-21
Fix theoretical IOOBE race

WouterGritter:


R00tB33rMan (approved):

4drian3d (approved):
I don't think that's likely to happen, but since it's good practice, I don't see a problem with it

WouterGritter:
Completely agree on both counts, this effectively won't ever happen (worst case we'll abort the ping), but it's an eye sore & bad practice in its current state.

electronicboy (approved):

PR #1800, opened 2026-05-21
Cap pre-join plugin-message queue size (prevents arbitrary growth/OOM)

WouterGritter:
A client that stalls its FML/login handshake could enqueue plugin messages (each up to ~32 KiB serverbound) without bound in `ClientPlaySessionHandler#loginPluginMessages`, growing per-connection heap until an OOM kill. This PR caps the queue by both bytes (4 MiB) and count (1024), configurable via `velocity.max-queued-login-plugin-message-bytes` and `velocity.max-queued-login-plugin-messages`, and disconnects on overflow.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientPlaySessionHandler.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientPlaySessionHandler.java
@@ -86,6 +86,8 @@ import java.util.Queue;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
 import net.kyori.adventure.key.Key;
 import net.kyori.adventure.text.Component;
 import net.kyori.adventure.text.ComponentLike;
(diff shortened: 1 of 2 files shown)


R00tB33rMan (approved):

WouterGritter:
Depending on how long a client is actually able to stall the login phase for, this may be a scary OOM once again.

~Without this fix, and at the default rate limit of 500 pps, a client would be able to fill this queue with 160MB of plugin messages every second (at 32kb/PM). That's 10GB within a minute, currently exploitable.~
My math was wrong; this may fill the queue at a rate of 16MB/s. Still 1GB per minute.

Of course this PR fixes that, and I'm sorry to keep bumping this but, #1786 would have greatly crippled this attack. At a limit of 5MB/s of _decompressed packets per second_, it would take around half an hour to fill the same 10GB in this queue. Still an OOM, if the client is actually able to stall the login phase for 30 minutes, but way less instant.

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientPlaySessionHandler.java
```diff
@@ -635,6 +677,8 @@ public void handleBackendJoinGame(JoinGamePacket joinGame, VelocityServerConnect
     // If we had plugin messages queued during login/FML handshake, send them now.
     PluginMessagePacket pm;
     while ((pm = loginPluginMessages.poll()) != null) {
+      loginPluginMessagesBytes.addAndGet(-pm.content().readableBytes());
+      loginPluginMessagesCount.decrementAndGet();
```
kennytv: This seems unnecessary vs. just clearing it below. when login is sent, the client has already sent its finish configuration packet and cannot change this counter anymore. same below

kennytv (approved):

electronicboy (approved):

PR #1802, opened 2026-05-22
Fix `TitleActionbarPacket`'s default action type by setting it to `SET_ACTION_BAR`

WouterGritter:
Correct me if I'm wrong, but it seems like `SET_ACTION_BAR` is the right action for the `TitleActionbarPacket` packet.

This bug doesn't surface when creating this class from the factory:
```java
// in GenericTitlePacket
public static GenericTitlePacket constructTitlePacket(ActionType type, ProtocolVersion version) {
  GenericTitlePacket packet = null;
  if (version.noLessThan(ProtocolVersion.MINECRAFT_1_17)) {
    packet = switch (type) {
      case SET_ACTION_BAR -> new TitleActionbarPacket();
      case SET_SUBTITLE -> new TitleSubtitlePacket();
      case SET_TIMES -> new TitleTimesPacket();
      case SET_TITLE -> new TitleTextPacket();
      case HIDE, RESET -> new TitleClearPacket();
      default -> throw new IllegalArgumentException("Invalid ActionType");
    };
  } else {
    packet = new LegacyTitlePacket();
  }
  packet.setAction(type);
  return packet;
}
```

Namely through:
```java
case SET_ACTION_BAR -> new TitleActionbarPacket();
packet.setAction(type); // overrides the type as SET_ACTION_BAR (overrides wrong constructor value)
```

`StateRegistry` does use `TitleActionbarPacket`'s constructor directly, but it gets discarded and never stored/used because of all mappings being encode-only.

As far as I can see this has no implications in Velocity itself,the action type is only read by `TitleClearPacket` `LegacyTitlePacket` (the latter doesn't even distinguish between `SET_TITLE` and `SET_ACTION_BAR`)

For correctness sake, this should be fixed.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/title/TitleActionbarPacket.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/title/TitleActionbarPacket.java
@@ -28,7 +28,7 @@ public class TitleActionbarPacket extends GenericTitlePacket {
   private ComponentHolder component;
 
   public TitleActionbarPacket() {
-    setAction(ActionType.SET_TITLE);
+    setAction(ActionType.SET_ACTION_BAR);
   }
 
   @Override


WouterGritter:
We may even enforce this contract by adding this below the switch statement in the factory method:
```java
if (packet.getAction() != type) {
  throw new AssertionError("Title packet action type mismatch!");
}
```

But I don't think it's worth it to assert this at runtime (and this won't even work, `TitleClearPacket` can be created with both `HIDE` and `RESET`).

The enforcing of this is also quite inconsistent.  `TitleClearPacket` throws when setting its action type to anything but HIDE and RESET, but the other packets let you set the action type to anything you want. Maybe it's worth expanding this PR to enforcing this for all other packet's setAction methods as well?

---

I have worked this out in the referenced PR

kennytv (approved):

R00tB33rMan (approved):

PR #1803, opened 2026-05-22
Clean up Title packets and enforce action types

WouterGritter:
~~Closes #1802, this PR builds on-top of the fix proposed and explained in that PR.~~

~This PR drops the method visibility, enforces the action types (similar pattern to the already-existing enforcement in `TitleClearPacket`) and cleans up some whitespace formatting.~

After the third commit, this PR now does away with the `setAction()` method completely, which was only used within this package. It does not make sense to set the action type at all here; they must be exactly one action for most modern packets, except for `TitleClearPacket` which supports two types (which is still enforced), and except for `LegacyTitlePacket` which can still hold any action type.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/title/LegacyTitlePacket.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/title/LegacyTitlePacket.java
@@ -22,27 +22,40 @@ import com.velocitypowered.proxy.connection.MinecraftSessionHandler;
 import com.velocitypowered.proxy.protocol.ProtocolUtils;
 import com.velocitypowered.proxy.protocol.packet.chat.ComponentHolder;
 import io.netty.buffer.ByteBuf;
+import java.util.Objects;
 import org.checkerframework.checker.nullness.qual.Nullable;
+import org.jetbrains.annotations.NotNull;
 
 public class LegacyTitlePacket extends GenericTitlePacket {
+
+  private final ActionType action;
 
   private @Nullable ComponentHolder component;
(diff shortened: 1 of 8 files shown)


proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/title/LegacyTitlePacket.java
```diff
@@ -26,23 +26,33 @@
...
 
+  private final ActionType action;
+
   private @Nullable ComponentHolder component;
   private int fadeIn;
   private int stay;
   private int fadeOut;
 
+  public LegacyTitlePacket() {
+    this(null);
```
WouterGritter: I don't like that this is null here, but this is exactly the old behavior. Previously this was less explicit as the `setAction()` setter just never got called.

WouterGritter:
`HeaderAndFooterPacket` has a similar situation, where it's never expected to be decoded:
```java
public HeaderAndFooterPacket() {
  throw new UnsupportedOperationException("Decode is not implemented");
}

@Override
public void decode(ByteBuf buf, ProtocolUtils.Direction direction, ProtocolVersion version) {
  throw new UnsupportedOperationException("Decode is not implemented");
}
```

I'll update this PR to match this convention.

WouterGritter:
This PR cleans up the title packet implementations quite a bit, IMO it should have been looked at more closely during its merge. This gets rid of the `setAction` method which was only ever used internally in the classes that this PR touches (in the proxy implementation module), and it would never have made sense to call this anyway as you'd create a wrapper class with the wrong action type.

PR #1806, opened 2026-05-25
Fix "Invalid signature" when not forwarding UUIDs to `online-mode=false` backends

evan-goode:
Filter the ServerboundChatSessionUpdatePacket when player info forwarding is disabled, since the client's profile key signature will not be valid for the player's offline UUID.

See comment in `ClientPlaySessionHandler.java`. This is marked as draft since it's in a "works for me" state, though I'm happy to address feedback and get this merged. I'm unsure whether this behavior belongs in a separate option or whether `player-info-forwarding-mode = "NONE"` is a good enough heuristic for when it's needed.

My use case is a public, Mojang-authenticated gateway to an online-mode=false Vanilla server (no Paper/Fabric) that unauthenticated bots can connect to over localhost.


R00tB33rMan:
@evan-goode, I'd probably mark this as ready so reviewers know it's time to suggest changes

Issue #1369, opened 2024-07-04
Resource pack callbacks are not triggered


scrayos:
### Expected Behavior

I expect the callbacks to be triggered just like the `PlayerResourcePackStatusEvent` works. There should be invocations of the callback for any change.

### Actual Behavior

The callback is not invoked at all.

### Steps to Reproduce

1. Send a valid resource pack like below
2. Wait for it to load successfully
3. See that there is no log message

```java
e.getPlayer().sendResourcePacks(
                ResourcePackRequest.resourcePackRequest()
                        .packs(
                                ResourcePackInfo.resourcePackInfo()
                                        .id(UUID.randomUUID())
                                        .uri(URI.create("https://example.io/…"))
                                        .hash("54ff1a362a90915aa44a3b3a5a…")
                                        .build()
                        )
                        .replace(true)
                        .required(true)
                        .callback(ResourcePackCallback.onTerminal(
                                (uuid, audience) -> System.out.println("SUCCESSFULLY LOADED IT"),
                                (uuid, audience) -> System.out.println("FAILED TO LOAD IT")
                        ))
                        .build()
        );
```

### Plugin List

Only the test plugin

### Velocity Version

[23:27:39 INFO]: Velocity 3.3.0-SNAPSHOT (git-aa4e8780-b401)
[23:27:39 INFO]: Copyright 2018-2023 Velocity Contributors. Velocity ist lizenziert unter den Bedingungen der GNU General Public License v3.
[23:27:39 INFO]: velocitypowered.com - GitHub

PR #1809, opened 2026-05-26
Trigger resourcepack callbacks

R00tB33rMan:
This targets the long-lost https://github.com/PaperMC/Velocity/issues/1369 issue.

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/event/player/PlayerResourcePackStatusEvent.java
+++ b/api/src/main/java/com/velocitypowered/api/event/player/PlayerResourcePackStatusEvent.java
@@ -13,6 +13,7 @@ import com.velocitypowered.api.network.ProtocolVersion;
 import com.velocitypowered.api.proxy.Player;
 import com.velocitypowered.api.proxy.player.ResourcePackInfo;
 import java.util.UUID;
+import net.kyori.adventure.resource.ResourcePackStatus;
 import org.checkerframework.checker.nullness.qual.MonotonicNonNull;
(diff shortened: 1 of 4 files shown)


Timongcraft:
This would hold callbacks for packs that don't get a terminal client response for as long as the client is connected, right?

Timongcraft:
Could also be an idea to add a `status()` method for the event to promote the adventure status.

I also thought about marking the Velocity one as obsolete but the Adventure one for example doesn't use the same enum order as Velocity & Minecraft, which would make handling the packet more annoying if we'd replace the Velocity one in the future.

WouterGritter:
> This would hold callbacks for packs that don't get a terminal client response for as long as the client is connected, right?

Do you think it's better to add a timeout to this? What if the client is just really slow though, downloading a 50MB resource pack on a 1mbit/s connection

Timongcraft:
> > This would hold callbacks for packs that don't get a terminal client response for as long as the client is connected, right?
> 
> Do you think it's better to add a timeout to this? What if the client is just really slow though, downloading a 50MB resource pack on a 1mbit/s connection

Yea I'm not sure, one of the maintainers might have an idea

WouterGritter:
> Yea I'm not sure, one of the maintainers might have an idea

Are clients able to trigger this flow, of adding a resource pack & thus registering a callback? I don't think so but I might be wrong.

If we can only trigger a callback being registered through a plugin, I don't think this is really a cause for concern. The lifetime of the map (and its entries if the client never responds) is only as long as the client is connected, worst case we're storing a couple stale entries but that buys us having compatibility with slow clients.

Issue #774, opened 2022-06-25
Sending PluginMessage to player before he join a server


michcie:
As we talked on discord, 
When you want to send PluginMessage to player in PostLoginEvent like that:
` 
@Subscribe
  public void onLoginEvent(PostLoginEvent e) {
    byte[] data = getData(e.getPlayer());
    e.getPlayer().sendPluginMessage(channel, data);
  }
`
  You got error
  `exception encountered in com.velocitypowered.proxy.connection.client.InitialConnectSessionHandler@379f0a7f
java.lang.IllegalStateException: Not connected to server!
        at com.velocitypowered.proxy.connection.backend.VelocityServerConnection.ensureConnected(VelocityServerConnection.java:208) ~[velocity-proxy-3.1.2-SNAPSHOT-all.jar:3.1.2-SNAPSHOT (git-662fbc4e-bunknown)]
        at com.velocitypowered.proxy.connection.client.InitialConnectSessionHandler.handle(InitialConnectSessionHandler.java:49) ~[velocity-proxy-3.1.2-SNAPSHOT-all.jar:3.1.2-SNAPSHOT (git-662fbc4e-bunknown)]
        at com.velocitypowered.proxy.protocol.packet.PluginMessage.handle(PluginMessage.java:104) ~[velocity-proxy-3.1.2-SNAPSHOT-all.jar:3.1.2-SNAPSHOT (git-662fbc4e-bunknown)]
        at com.velocitypowered.proxy.connection.MinecraftConnection.channelRead(MinecraftConnection.java:142) ~[velocity-proxy-3.1.2-SNAPSHOT-all.jar:3.1.2-SNAPSHOT (git-662fbc4e-bunknown)]
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379) ~[velocity-proxy-3.1.2-SNAPSHOT-all.jar:3.1.2-SNAPSHOT (git-662fbc4e-bunknown)]
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365) ~[velocity-proxy-3.1.2-SNAPSHOT-all.jar:3.1.2-SNAPSHOT (git-662fbc4e-bunknown)]`
        
Which is wrong because we don't need server to send pluginMessage to player.

When fixing this error you have to remember that PluginMessageEvent firing has not been implemented inside InitialConnectSessionHandler

PR #1811, opened 2026-05-27
Fix inverted PluginMessageEvent source/target in InitialConnectSessionHandler

GoldenEdit:
- Fix `PluginMessageEvent` in `InitialConnectSessionHandler` so client -> backend plugin messages use `source = player` and `target = serverConn`, matching `ClientPlaySessionHandler` and `ClientConfigSessionHandler`.
- The handler previously used the backend -> client argument order from `BackendPlaySessionHandler` (introduced in #774), so client messages were exposed to plugins with `ServerConnection` as the source and `Player` as the target.


R00tB33rMan (approved):

WouterGritter (approved):
Should be merged asap

electronicboy (approved):

PR #1816, opened 2026-05-29
Add CLI options to load plugins from extra jars and directories

vjh0107:
Hi! This PR adds two cli options which allow loading additional plugin jars and directories in addition to the default plugins folder.
  
The data directory is intentionally kept fixed rather than tied to the plugin jar location. Paper takes the same approach, and since this option is likely to be used together with container volumes, which are often mounted readonly, keeping mutable data separate from the plugin binaries seems like the safer choice.
  
  https://github.com/PaperMC/Paper/blob/216388dfdf25b365e0d3ec24748def4dcaa30a76/paper-server/src/main/java/io/papermc/paper/plugin/entrypoint/classloader/PaperPluginClassLoader.java#L176
  
  For this reason, loadPlugins still keeps the directory argument so it can be used for the data directory.
  
  https://github.com/vjh0107/Velocity/blob/feat/cli-extra-plugins/proxy/src/main/java/com/velocitypowered/proxy/plugin/VelocityPluginManager.java#L91~L94

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/ProxyOptions.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/ProxyOptions.java
@@ -21,6 +21,7 @@ import com.velocitypowered.api.proxy.server.ServerInfo;
 import com.velocitypowered.proxy.util.AddressUtil;
 import java.io.IOException;
 import java.net.InetSocketAddress;
+import java.nio.file.Path;
 import java.util.Arrays;
 import java.util.List;
 import joptsimple.OptionParser;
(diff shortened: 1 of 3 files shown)


WouterGritter:
> [...] container volumes, which are often mounted readonly [...]

How will this work if #1736 gets merged? Mounting a read-only volume for plugin jars would break the plugin update feature this PR is attempting to implement.

vjh0107:
The readonly thing here is only about the extra plugin paths from the new CLI options (immutable jars, e.g. mounted from a container image). #1736's update folder lives under the normal plugins/ dir, which stays writable like it is now.
So you'd just keep updatable plugins in plugins/ and put the immutable ones behind the readonly mounts. The readonly mount never touches the update folder, so nothing breaks!

vjh0107:
@WouterGritter If there's anything else I need to do, please feel free to tell me!

WouterGritter:
> @WouterGritter If there's anything else I need to do, please feel free to tell me!

I'm not a maintainer with any rights on this repository, just a contributor sharing their thoughts :)

PR #1818, opened 2026-05-30
Snoop the backend's player list header/footer so the proxy's values stay in sync

WouterGritter:
Right now this kinda a half-implemented feature. We can send player list header/footer `Component`s to the player from proxy plugins, and we can also get the ones last sent by the proxy through plugins, but this is misleading. It's not the current header/footer that the client is displaying, as a backend may have overridden these.

Would it be worth it to introduce an event for this as well? One that would fire both when the backend sends this packet, and also when a plugin sets these manually with `ConnectedPlayer#sendPlayerListHeaderAndFooter`, allowing the `Component`s to be manipulated?

~This has an added bonus (or regression?) of keeping a client's player list header/footer when they switch servers now. `ConfigSessionHandler` line 247 currently sends these values right after switching to PLAY.~


NEZNAMY (approved):

Timongcraft:
IMO it is ok as is but I think this could be nice especially with an event to block backend changes.
But currently it is not expected to carry over the backends header and footer on switch which could be confusing for existing plugins. Holding the backends state only for the server connection could be an ugly solution to that.
Seems like 1.8 also already encoded it as component, but I didn't test it.

Small nitpicks:
I believe we don't need the JD for internal silent methods as they should be self explanatory, same with the `// forward` comment and well the comment above that too.

Toffikk (approved):

WouterGritter:
> ~This has an added bonus (or regression?) of keeping a client's player list header/footer when they switch servers now. ConfigSessionHandler line 247 currently sends these values right after switching to PLAY.~

So I was wrong: currently, the playerlist header & footer _do not_ survive a server switch. Looks like this is by design, as we do re-send the header/footer after server switch (as per the PR description), but before that, we clear the header/footer (see `ClientPlaySessionHandler#doSwitch`, line 606.

This means that right now (before and after this PR), eventhough `playerListHeader` and `playerListFooter` are stored in `ConnectedPlayer`, they are reset on server switch. This PR doesn't change this, it only ensures the stored values are now up to date with what the client actually sees (if the backend changes it)

Apart from an event (should we add this?), this PR should be good to go.

R00tB33rMan (approved):

PR #1820, opened 2026-06-02
Reduce default `read-timeout`

### LINKED ISSUE #1819: `read-timeout` is too high, effectively does nothing in some scenarios
```toml
# Specify a read timeout for connections here. The default is 30 seconds.
read-timeout = 30000
```

The default `read-timeout` of 30 seconds matches the (modern) client value exactly. When a backend server dies/stalls completely (simulate with `kill -STOP <pid>`), the client disconnects itself after 30 seconds. Instead of Velocity considering this a read timeout (and sending them to a fallback server, or gracefully disconnecting the client), the client hard disconnects with "Timed out".

Setting the default `read-timeout` to some delta below the client's timeout would be the solution here, e.g. `25000` (25 seconds). I'd argue this could be even lower, but 25 seconds seems like a reasonable value, keeping the client's timeout in mind.

One issue here is that a PR updating the default to 25 seconds won't fix existing setups. Could this warrant a migration?

WouterGritter:
Fixes #1819

Reduces the default `read-timeout` to 25 seconds, avoiding the client disconnecting itself (tested & verified), and adds a config migration for an unchanged old value.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
@@ -32,6 +32,7 @@ import com.velocitypowered.proxy.config.migration.KeyAuthenticationMigration;
 import com.velocitypowered.proxy.config.migration.MiniMessageTranslationsMigration;
 import com.velocitypowered.proxy.config.migration.MotdMigration;
 import com.velocitypowered.proxy.config.migration.PacketLimiterMigration;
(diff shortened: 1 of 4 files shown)


electronicboy (approved):
needs rebasing/config migration version updating

WouterGritter:
> config migration version updating

This would bump the toml version to 3.0 (we can't have 2.10 as it's read as a double), I'd rather merge this after we do the yaml migration to a whole integer instead (then update 3 -> 4).

PR #1825, opened 2026-06-18
Support multi-line MOTDs natively via a string list in the config

WouterGritter:
As the title says.

As a bonus, we get rid of the ugly FQDN instead of an import for `Component` in `ProxyConfig` and `VelocityConfig`, and `ProxyConfig#getMotd`'s javadoc is updated to correctly reflect where the MOTD is actually being used.

I had first used a regex in `MotdMultilineMigration` to split on `literal newline`, `\n` `<newline>` and `<br>` ([MM's newline tags](https://docs.papermc.io/adventure/minimessage/format/#newline)) when migrating, but this seemed too intrusive. In its current state, this PR just migrates the existing string to a string list of size 1 and leaves it up to the user to split out into multiple lines.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
@@ -31,6 +31,7 @@ import com.velocitypowered.proxy.config.migration.ForwardingMigration;
 import com.velocitypowered.proxy.config.migration.KeyAuthenticationMigration;
 import com.velocitypowered.proxy.config.migration.MiniMessageTranslationsMigration;
 import com.velocitypowered.proxy.config.migration.MotdMigration;
+import com.velocitypowered.proxy.config.migration.MotdMultilineMigration;
 import com.velocitypowered.proxy.config.migration.PacketLimiterMigration;
 import com.velocitypowered.proxy.config.migration.TransferIntegrationMigration;
 import com.velocitypowered.proxy.util.AddressUtil;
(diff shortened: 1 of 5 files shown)


R00tB33rMan (approved):

proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
```diff
@@ -545,7 +551,7 @@ public static VelocityConfiguration read(Path path) throws IOException {
         }
       }
       final byte[] forwardingSecret = forwardingSecretString.getBytes(StandardCharsets.UTF_8);
-      final String motd = config.getOrElse("motd", "<#09add3>A Velocity Server");
+      final List<String> motd = config.getOrElse("motd", ImmutableList.of("<#09add3>A Velocity Server"));
```
4drian3d, line 554: I think it would be a good idea to add a validation check in case someone tries to add more than 2 lines

proxy/src/main/java/com/velocitypowered/proxy/config/migration/MotdMultilineMigration.java
```diff
@@ -0,0 +1,41 @@
...
+  public boolean shouldMigrate(CommentedFileConfig config) {
+    return configVersion(config) < 2.9;
+  }
+
+  @Override
+  public void migrate(CommentedFileConfig config, Logger logger) {
+    String motd = config.get("motd");
+    config.set("motd", List.of(motd));
+
+    config.set("config-version", "2.9");
```
4drian3d, line 39: Since the migration will only happen once, it might be a good idea to migrate the use of `\n` or `<newline>`/`<br>` to a new line in the configuration

WouterGritter: I agree, though it'll be best-effort and we probably dont want to match every case. Splitting on those 3 delimiters seems reasonable (case insensitive) but afaik minimessage also supports self-closing tags (?), possibly spaces between the tag name and the self-closing `/`. Emulating the minimessage parsing engine seems out of scope, and deserializing+serializing would be too intrusive imo.

PR #1826, opened 2026-06-22
Add wildcard forced hosts (fix #1587)

### LINKED ISSUE #1587: [Feature Request] Wildcard Forced Host Support
### Requested Feature

Make the `[forced-hosts]` domains support wildwards (ie. `*.example.com`) rather than having to statically define subdomains (`play.example.com`)

### Why is this needed?

Most forced host systems in software allow for wildcard support already and this allows for networks looking to bundle a vast majority of subdomains for different IPs into singular wildcards.

For example, this is my current setup without wildcards:
```toml
[forced-hosts]
# Configure your forced hosts here.

# Soul Realms
"play.soulrealms.net" = ["soulrealms-smp-1" ]
"yt.soulrealms.net" = [ "soulrealms-smp-1" ]
"ip.soulrealms.net" = [ "soulrealms-smp-1" ]
"soulrealms.net" = ["soulrealms-smp-1" ]
"soulrealms.minehut.gg" = [ "soulrealms-smp-1" ]

# Misc PvP
"play.miscpvp.org" = [ "miscpvp-gens-1" ]
"yt.miscpvp.org" = [ "miscpvp-gens-1" ]
"miscpvp.org" = [ "miscpvp-gens-1" ]
"ip.miscpvp.org" = [ "miscpvp-gens-1" ]
"miscpvp.minehut.gg" = [ "miscpvp-gens-1" ]
```

This discourages IP-based join tracking and forces players to be sent to a hub server rather than forwarded directly if their IP differs from the configured list. With wildcards, this would become:
```toml
[forced-hosts]
# Configure your forced hosts here.

# Soul Realms
"*.soulrealms.net" = [ "soulrealms-smp-1" ]
"soulrealms.net" = ["soulrealms-smp-1" ]
"soulrealms.minehut.gg" = [ "soulrealms-smp-1" ]

# Misc PvP
"*.miscpvp.org" = [ "miscpvp-gens-1" ]
"miscpvp.org" = [ "miscpvp-gens-1" ]
"miscpvp.minehut.gg" = [ "miscpvp-gens-1" ]
```

Allowing users to use enhanced subdomain tracking without sending someone to an arbitrary hub server due to not knowing where to forward the player.

### Alternative Solutions

Only alternative solution is to manually define each and every potential subdomain a player could join with, which can be clunky and bolsters line count significantly.

### Additional Information

_No response_

lllincoln:
I added wildcard support to `[forced-hosts]`, resolving #1587. Unit tests pass and checkstyle passes and I've tested this against my own running proxy. I'd appreciate feedback or additional testing if needed, especially from those on #1587 who submitted the issue.

### What this does

Consider the example from #1587

```toml
[forced-hosts]
# Soul Realms
"play.soulrealms.net" = [ "soulrealms-smp-1" ]
"yt.soulrealms.net"   = [ "soulrealms-smp-1" ]
"ip.soulrealms.net"   = [ "soulrealms-smp-1" ]
"soulrealms.net"      = [ "soulrealms-smp-1" ]
```

Now this can be changed to:

```toml
[forced-hosts]
# Soul Realms
"*.soulrealms.net" = [ "soulrealms-smp-1" ]
"soulrealms.net"   = [ "soulrealms-smp-1" ]
```

### How matching works

I added `AddressUtil.isHostMatchingPattern(pattern, host)`. Patterns are matched on each label where `*` matches exactly one label. To be matched, patterns must have the same number of labels as the host. Matching is case insensitive. For example:

- `*.example.com` matches `a.example.com` but not `example.com` (fewer labels) or `b.a.example.com` (more labels).
- `b.*.example.com` matches `b.a.example.com`

### Matching order

In both `ConnectedPlayer` and `ServerListPingHandler`, forced-host lookup now follows a hierarchy where exact matches always win over wildcards, so adding a `*.example.com` entry won't change routing for any host you've already defined explicitly.

1. **Exact match** (unchanged).
2. **Wildcard match** (my addition).
3. **Default** (`getAttemptConnectionOrder()` in `ServerListPingHandler`, or an empty collection in `ConnectedPlayer`. this are both unchanged from what was before).

### Tests

I added `ForcedHostsTest` covering `isHostMatchingPattern` cases for single and mid-pattern wildcards, label-count mismatches, and a non-matching domain.

### Potential issue

If two different wildcard patterns could match the same host, the one chosen depends on config map iteration order. Would like feedback on what the behavior should be here


proxy/src/test/java/com/velocitypowered/proxy/connection/client/ForcedHostsTest.java
```diff
@@ -0,0 +1,35 @@
...
+class ForcedHostsTest {
+  @Test
+  void testIsHostMatchingPattern() {
+    Assertions.assertTrue(AddressUtil.isHostMatchingPattern("*.miscpvp.org", "play.miscpvp.org"));
+    Assertions.assertTrue(AddressUtil.isHostMatchingPattern("*.miscpvp.org", "yt.miscpvp.org"));
+    Assertions.assertTrue(AddressUtil.isHostMatchingPattern("*.miscpvp.org", "ip.miscpvp.org"));
+    Assertions.assertTrue(AddressUtil.isHostMatchingPattern("test.*.miscpvp.org", "test.example.miscpvp.org"));
+    Assertions.assertFalse(AddressUtil.isHostMatchingPattern("*.miscpvp.org", "test.example.miscpvp.org"));
+    Assertions.assertFalse(AddressUtil.isHostMatchingPattern("*.miscpvp.org", "miscpvp.org"));
+    Assertions.assertFalse(AddressUtil.isHostMatchingPattern("miscpvp.minehunt.gg", "ip.miscpvp.org"));
```
underscore11code: These should be using fictious example domains (i.e. `example.com`), not real domains.

underscore11code (changes requested):

proxy/src/test/java/com/velocitypowered/proxy/connection/client/ForcedHostsTest.java
lllincoln: Updated tests to use fictitious examples.

underscore11code (approved):

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
```diff
@@ -890,8 +892,17 @@ private Optional<RegisteredServer> getNextServerToTry(@Nullable RegisteredServer
...
+      serversToTry = server.getConfiguration().getForcedHosts().getOrDefault(
+              virtualHostStr,
+              server.getConfiguration()
+                      .getForcedHosts()
+                      .entrySet()
+                      .stream()
+                      .filter(entry -> AddressUtil.isHostMatchingPattern(entry.getKey(), virtualHostStr))
+                      .map(Map.Entry::getValue)
+                      .findFirst()
+                      .orElse(Collections.emptyList())
```
WouterGritter: Three nits:

1 - This is a pretty large method chain that's duplicated twice, it would probably make sense to extract this out into its own method.

2 - the "or default" value is always computed, even in the case where the exact `virtualHostStr` mapping is present. And the `.orElse(Collections.emptyList())` also always returns an empty list. Fortunately this doesn't comput or allocate anything, but this means that all three cases - exact match, pattern match, empty list - are always computed, even when an exact match could abort the latter two, which is a smell

3 - Essentially this is a giant double if-else statement:
```java
if (/* exact forced host defined */) {
  serversToTry = /* exact forced host */;
} else if (/* isHostMatchingPattern matches */) {
  serversToTry = /* isHostMatchingPattern result */;
} else {
  serversToTry = Collections.emptyList();
}
```
However this is disguised in `getOrDefault` (first two `if` and `if else`) and `.orElse(Collections.emptyList()` (last `else`). I would probably make this more explicit, either something similar to the above pseudo code or using an optional chain with 2x `.orElseGet()`, keeping it all in optional-land.

proxy/src/test/java/com/velocitypowered/proxy/connection/client/ForcedHostsTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+package com.velocitypowered.proxy.connection.client;
+
+import com.velocitypowered.proxy.util.AddressUtil;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+class ForcedHostsTest {
+  @Test
+  void testIsHostMatchingPattern() {
```
WouterGritter: Every comment distinguishes different behavior. Placing each one in their own function is the canonical way to go about this. A bonus: you can drop the comment in favor of a descriptive method name

proxy/src/main/java/com/velocitypowered/proxy/util/AddressUtil.java
```diff
@@ -74,4 +74,37 @@ public static InetSocketAddress parseAndResolveAddress(String ip) {
...
     return new InetSocketAddress(uri.getHost(), port);
   }
+
+  /**
+   * Tests whether a host matches a pattern whose labels may be the
+   * wildcard {@code "*"}. Each {@code "*"} matches exactly one label; all other
+   * labels match case-insensitively.
+   *
+   * @param pattern the pattern to match against, for example, {@code *.soulrealms.net}
+   * @param host the virtual host to test, for example, {@code play.soulrealms.net}
```
WouterGritter: Instead of naming an actual domain here, I would rather use `example.com` or a similar domain name

proxy/src/main/java/com/velocitypowered/proxy/util/AddressUtil.java
```diff
@@ -74,4 +74,37 @@ public static InetSocketAddress parseAndResolveAddress(String ip) {
...
+   * @return true if the host matches the pattern, false otherwise
+   */
+  public static boolean isHostMatchingPattern(String pattern, String host) {
```
WouterGritter: As this is a utility method that may be used elsewhere and both parameters are nullable, it's good practice to annotate both with `@Nullable`

proxy/src/main/java/com/velocitypowered/proxy/util/AddressUtil.java
```diff
@@ -74,4 +74,37 @@ public static InetSocketAddress parseAndResolveAddress(String ip) {
...
+    }
+
+    String[] patternDomains = pattern.split("\\.");
+    String[] strDomains = host.split("\\.");
+
+    if (patternDomains.length != strDomains.length) {
+      return false;
+    }
+
+    for (int i = 0; patternDomains.length > i; i++) {
```
WouterGritter: `patternDomains.length > i` here is odd, I have never seen it written this way, I'd change this to `i < patternDomains.length`

WouterGritter (changes requested):

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
lllincoln: Good call, I was thinking about this too. Where do you think the right place to hold this new method could be? I would have put it in `ConnectedPlayer`, but this same pattern is in `ServerListPingHandler` as well, which are two separate things from each other. Thoughts?

proxy/src/test/java/com/velocitypowered/proxy/connection/client/HostPatternMatchingTest.java
```diff
@@ -0,0 +1,71 @@
...
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package com.velocitypowered.proxy.connection.client;
+
+import com.velocitypowered.proxy.util.AddressUtil;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+class HostPatternMatchingTest {
```
WouterGritter: Another nit: try to match the test class name with the class it's actually testing. In this case it would be `AddressUtilTest`. If this test class already exists and/or these test methods would seem out of place, consider adding a nested class, e.g.

```java
class AddressUtilTest {

  class HostMatchingPatternTest {
    /* isHostMatchingPattern test methods */
  }
}
```

Some people are against this, but when doing this, I would also consider statically importing the `AddressUtil.isHostMatchingPattern` method.

proxy/src/main/java/com/velocitypowered/proxy/connection/client/ConnectedPlayer.java
WouterGritter: `AddressUtil` seems like the right place, as you have already introduced the new `isHostMatchingPattern` there.

Consider making it return an `Optional<String>`, allowing the caller to change an empty result to an empty list (`ConnectedPlayer`) or `server.getConfiguration().getAttemptConnectionOrder()` (`ServerListPingHandler`).  (with `orElseGet()` to avoid computing when the optional is present)

lllincoln: See [54d6816](https://github.com/PaperMC/Velocity/pull/1826/commits/54d6816e17d0c352024368f60479f6cfb5845c2b). I've implemented it using `Optional<String>` but not 2x `.orElseGet()` since I find using if/else and a for loop a bit more readable, but let me know if you think otherwise

proxy/src/main/java/com/velocitypowered/proxy/connection/util/ServerListPingHandler.java
```diff
@@ -174,8 +175,8 @@ public CompletableFuture<ServerPing> getInitialPing(VelocityInboundConnection co
       String virtualHostStr = connection.getVirtualHost().map(InetSocketAddress::getHostString)
           .map(str -> str.toLowerCase(Locale.ROOT))
           .orElse("");
-      List<String> serversToTry = server.getConfiguration().getForcedHosts().getOrDefault(
-          virtualHostStr, server.getConfiguration().getAttemptConnectionOrder());
+      List<String> serversToTry = AddressUtil.resolveForcedHostServers(server, virtualHostStr)
+          .orElseGet(() -> server.getConfiguration().getAttemptConnectionOrder());
```
WouterGritter:
```suggestion
          .orElseGet(server.getConfiguration()::getAttemptConnectionOrder);
```

WouterGritter:
This would most likely make #1730 redundant.

lllincoln:
This would likely make #1730 redundant. However, this fix is more flexible because it adds pattern matching and this fix doesn't need a config migration. That said, #1730 does consider SRV records, which I didn't fully consider yet. I can integrate this though.

PR #1834, opened 2026-06-28
Resolve backend DNS on a bounded thread pool to avoid head-of-line blocking

### LINKED ISSUE #1833: Backend DNS resolution runs on a single thread, so one slow lookup blocks connections to other servers
### Expected Behavior

A slow DNS lookup for one backend shouldn't delay connecting to other backends. Sending a player to a server whose hostname is fine (cached or fast to resolve) should be quick regardless of other lookups happening at the same time.

### Actual Behavior

`SeparatePoolInetNameResolver` runs every backend hostname lookup on a single thread, with a 30s cache:

https://github.com/PaperMC/Velocity/blob/dev/3.0.0/proxy/src/main/java/com/velocitypowered/proxy/network/netty/SeparatePoolInetNameResolver.java#L57-L90

If a lookup is slow, every other uncached resolution queues behind it on that one thread until it returns. Cached names are fine (they skip the executor), but anything not currently cached has to wait. There's also no timeout on the lookup, so one hung resolve pins the resolver thread for the full JDK resolver duration.

We hit this on a fallback. A player gets kicked and is redirected to the fallback server (defined by hostname). The connect can't start until the name resolves, and that resolve gets stuck behind an unrelated slow lookup, so the player sits on no server for 10-18s. The fallback hostname itself resolves instantly when looked up on a separate thread during the same window, so it's the shared resolver thread that's blocked, not the name or the network.

Backend connects go through `VelocityServerConnection.connect()` -> `VelocityServer.createBootstrap()` -> `ConnectionManager.createWorker()`, which sets `.resolver(this.resolver.asGroup())`, so all of them share this one resolver.

### Steps to Reproduce

1. Define a backend by hostname in velocity.toml (e.g. the fallback: `lobby = "lobby:30066"`).
2. Run somewhere lookups can intermittently hang or be slow. Easy to hit with Docker's embedded DNS while containers come and go, but any flaky/slow resolver works.
3. Trigger a resolve of a slow name (any connect or ping to a slow-resolving backend).
4. While that's in flight, send a player to the fallback after its 30s cache entry has expired.
5. The fallback connect hangs for several seconds even though the name resolves fine on its own.

The stall is after ServerPreConnectEvent finishes and before the backend socket opens, i.e. in the resolve step.

### Plugin List

Proxy: floodgate 2.2.5, Geyser-Velocity 2.10.1, LuckPerms 5.5.32, PacketEvents 2.12.3, Skoice 3.3.1, spark 1.10.172, plus a few of our own plugins.
Backend: Paper with our own plugins.

### Velocity Version

[17:03:09 INFO]: Velocity 3.5.0-SNAPSHOT (git-a7581821-b605)
[17:03:09 INFO]: Copyright 2018-2026 Velocity Contributors. Velocity is licensed under the terms of the GNU General Public License v3.
[17:03:09 INFO]: PaperMC - GitHub

### Additional Information

Workaround: point the backend at an IP instead of a hostname so it never goes through the resolver.

carlodrift:
Fixes #1833.

`SeparatePoolInetNameResolver` runs every backend hostname lookup on a single thread (`Executors.newSingleThreadExecutor`). Because the lookups are blocking, one slow resolve serializes all the others behind it: an unrelated stalled DNS query delays connecting players to otherwise-healthy servers until it returns or the JDK resolver times out. We saw fallback connections to a hostname-defined server hang for 10–18s while the same name resolved instantly on another thread.

This swaps the single-thread executor for a small bounded pool (`ThreadPoolExecutor`, 0–8 threads, 60s keep-alive, `SynchronousQueue`), so a slow lookup only ties up one thread instead of blocking all resolution. Threads are created on demand and reaped when idle, so it costs nothing at rest.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/network/netty/SeparatePoolInetNameResolver.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/network/netty/SeparatePoolInetNameResolver.java
@@ -32,16 +32,19 @@ import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.util.List;
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
 import java.util.concurrent.RejectedExecutionException;
+import java.util.concurrent.SynchronousQueue;
+import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 
 /**
  * An implementation of {@code InetNameResolver} that performs blocking DNS name lookups
- * in a separate thread, avoiding blocking the Netty threads for an extended period of time
- * and without the downsides of Netty's native DNS resolver.
+ * on a small bounded pool of separate threads, avoiding blocking the Netty threads for an
+ * extended period of time and without the downsides of Netty's native DNS resolver.
  */
 public final class SeparatePoolInetNameResolver extends InetNameResolver {
+
+  private static final int MAX_RESOLVE_THREADS = 8;
 
   private final ExecutorService resolveExecutor;
   private final InetNameResolver delegate;
@@ -56,9 +59,10 @@ public final class SeparatePoolInetNameResolver extends InetNameResolver {
    */
   public SeparatePoolInetNameResolver(EventExecutor executor) {
     super(executor);
-    this.resolveExecutor = Executors.newSingleThreadExecutor(
+    this.resolveExecutor = new ThreadPoolExecutor(0, MAX_RESOLVE_THREADS,
+        60L, TimeUnit.SECONDS, new SynchronousQueue<>(),
         new ThreadFactoryBuilder()
-            .setNameFormat("Velocity DNS Resolver")
+            .setNameFormat("Velocity DNS Resolver #%d")
             .setDaemon(true)
             .build());
     this.delegate = new DefaultNameResolver(executor);
--- a/proxy/src/main/java/com/velocitypowered/proxy/network/netty/SeparatePoolInetNameResolver.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/network/netty/SeparatePoolInetNameResolver.java
@@ -32,8 +32,8 @@ import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.util.List;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.LinkedBlockingQueue;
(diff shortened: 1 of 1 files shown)


WouterGritter:
Since we use java 21, aren't virtual threads a better fit for this?

carlodrift:
`InetAddress.getByName` resolves in a native call, which pins the carrier on 21, so a stuck lookup still holds a real OS thread and virtual threads bring back the same serialization. A dedicated bounded pool stays predictable.

WouterGritter:
Interestingly, we did use Netty's async DNS resolver here in the past: 6e7c0298de14c1af2ad5aa4161f6bde054f6bb74

But this seems to come with some downsides (as the javadoc comment on `SeparatePoolInetNameResolver` also suggests). Wanted to suggest this as an alternative but that seems like a no-go.

One nit:
When 8 concurrent DNS requests occur, we will now throw `RejectedExecutionException` on the next request (`SynchronousQueue` has no size).
Something like
```java
ThreadPoolExecutor executor = new ThreadPoolExecutor(
    MAX_RESOLVE_THREADS, MAX_RESOLVE_THREADS,
    60L, TimeUnit.SECONDS,
    new LinkedBlockingQueue<>(), // unbounded -> never rejects
    new ThreadFactoryBuilder()
        .setNameFormat("Velocity DNS Resolver #%d")
        .setDaemon(true)
        .build());
executor.allowCoreThreadTimeOut(true);
```
Wouldn't throw and has the same queueing behavior as before, just with multiple threads (`corePoolSize` must also be `MAX_RESOLVE_THREADS`, otherwise the queue would always accept an incoming request without scaling up the amount of threads, resulting in the same single-thread behavior).

carlodrift:
Good catch, pushed a fix.

WouterGritter (approved):

electronicboy (approved):

electronicboy:
Netty DNS just turned out to be fairly unreliable for us in the past, worked fine for 99% of people but the 1% who'd have the odd issues with it just wasn't really acceptable; Felt kinda sad because it was fairly nice when it wasn't broken, especially as it defaulted to google DNS or something rather than relying on hosting providers DNS servers

WouterGritter:
> it defaulted to google DNS or something rather than relying on hosting providers DNS servers

Yeah that sounds bad actually. If the hosting provider uses split-horizon DNS for backends or you do it manually with docker, you really need to respect the native DNS resolver

PR #1835, opened 2026-07-02
Store historical console commands in .console_history (enables command history between restarts)

WouterGritter:
[Mirrors how Paper does things](https://github.com/PaperMC/Paper/blob/main/paper-server/src/main/java/com/destroystokyo/paper/console/PaperConsole.java#L24). Allows the user to press up to get previously executed commands, now persisting between reboots. Very useful for (plugin) development.


R00tB33rMan (approved):
Always liked this idea. Looks ready for merge!

jonesdevelopment (approved):

electronicboy (approved):

PR #1836, opened 2026-07-03
Strip pre-java-9 version check in `Metrics` as Velocity targets 21+

WouterGritter:
As Velocity builds to target java 21, there's really no need to support legacy version parsing.


kennytv (dismissed):
`Runtime.version()` contains the feature version and extras

WouterGritter:
> `Runtime.version()` contains the feature version and extras

`Runtime.version().toString()` will produce an output like `21.0.1+12-LTS`, not the `21.0.1` that `System.getProperty("java.version")` provides.

I'd propose the following refactor:
```java
metrics.addCustomChart(new DrilldownPie("java_version", () -> {
  Runtime.Version version = Runtime.version();

  return Map.of(
      "Java " + version.feature(),
      Map.of(System.getProperty("java.version"), 1));
}));
```

but it seems a bit odd to mix the old "api" (system property) and the new Runtime.Version API here. Thoughts?

Edit: I think it goes without saying that we shouldn't mess up the output to divide bStats metrics after this PR.

Edit 2: See latest commit, it's pretty easy to reconstruct the `java.version` system property exactly given a `Runtime.Version`

kennytv:
Mixing is fine I'd say

WouterGritter:
> Mixing is fine I'd say

Better than the `javaVersion` helper method? I'm leaning more towards that approach (the current state of the PR)

electronicboy (approved):

PR #1840, opened 2026-07-06
feat: add stable session UUID to MinecraftConnection

xfy2412:
Add a stable connection identifier to MinecraftConnection

A unique connectionId (AtomicLong) is held in MinecraftConnection,
the shared underlying instance across all InboundConnection wrappers,
surfaced via InboundConnection.getSessionId().

This allows plugins to reliably track the same TCP connection across
PreLoginEvent, GameProfileRequestEvent, and LoginEvent, even when
the InboundConnection wrapper object differs between events.

Tested with both online-mode and offline-mode connections:
- [x] Same connection across all login-phase events: same connId
- [x] Different players: different connId
- [x] Premium (Mojang-authenticated) connections: connId stable

### DIFF
--- a/api/src/main/java/com/velocitypowered/api/proxy/InboundConnection.java
+++ b/api/src/main/java/com/velocitypowered/api/proxy/InboundConnection.java
@@ -74,7 +74,7 @@ public interface InboundConnection {
    * Returns the unique, stable session UUID for this connection, generated when
    * the client initially connects to the proxy. The session id remains consistent
    * across all login-phase events and all proxy-to-backend connections for the
-   * same player session.
+   * same player session. This should not be confused with Mojang's server-wide play session ID introduced in Minecraft 26.2.
    *
    * @return the session UUID, never {@code null}
    */
--- a/api/src/main/java/com/velocitypowered/api/proxy/InboundConnection.java
+++ b/api/src/main/java/com/velocitypowered/api/proxy/InboundConnection.java
@@ -12,6 +12,7 @@ import com.velocitypowered.api.network.ProtocolState;
 import com.velocitypowered.api.network.ProtocolVersion;
 import java.net.InetSocketAddress;
 import java.util.Optional;
+import java.util.UUID;
 
 /**
  * Represents an incoming connection to the proxy.
@@ -70,11 +71,13 @@ public interface InboundConnection {
   HandshakeIntent getHandshakeIntent();
 
   /**
-   * Returns a unique, stable identifier for this connection.
+   * Returns a unique, stable UUID for this connection, generated at
+   * connection creation time. The UUID remains consistent across all
+   * login-phase events for the same TCP connection.
    *
-   * @return the connection id
+   * @return the connection UUID, or {@code null} if not supported
    */
-  default long getConnectionId() {
-    return -1;
+  default UUID getConnectionId() {
+    return null;
   }
 }
--- a/api/src/main/java/com/velocitypowered/api/proxy/InboundConnection.java
+++ b/api/src/main/java/com/velocitypowered/api/proxy/InboundConnection.java
@@ -68,4 +68,13 @@ public interface InboundConnection {
    * @return the intent of the connection
    */
   HandshakeIntent getHandshakeIntent();
+
+  /**
+   * Returns a unique, stable identifier for this connection.
+   *
+   * @return the connection id
+   */
+  default long getConnectionId() {
+    return -1;
+  }
 }
--- a/api/src/main/java/com/velocitypowered/api/proxy/InboundConnection.java
+++ b/api/src/main/java/com/velocitypowered/api/proxy/InboundConnection.java
@@ -71,13 +71,12 @@ public interface InboundConnection {
   HandshakeIntent getHandshakeIntent();
 
   /**
-   * Returns a unique, stable UUID for this connection, generated at
-   * connection creation time. The UUID remains consistent across all
-   * login-phase events for the same TCP connection.
+   * Returns the unique, stable session UUID for this connection, generated when
+   * the client initially connects to the proxy. The session id remains consistent
+   * across all login-phase events and all proxy-to-backend connections for the
+   * same player session.
    *
-   * @return the connection UUID, or {@code null} if not supported
+   * @return the session UUID, never {@code null}
    */
-  default UUID getConnectionId() {
-    return null;
-  }
+  UUID getSessionId();
 }
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/MinecraftConnection.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/MinecraftConnection.java
@@ -88,7 +88,7 @@ public class MinecraftConnection extends ChannelInboundHandlerAdapter {
 
   private static final Logger logger = LogManager.getLogger(MinecraftConnection.class);
 
-  private final UUID connectionId = UUID.randomUUID();
+  private final @Nullable UUID sessionId;
   private final Channel channel;
   public boolean pendingConfigurationSwitch = false;
   private SocketAddress remoteAddress;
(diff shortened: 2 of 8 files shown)


WouterGritter:
You're using FQDN's in the code instead of importing, and I'd rather like to see the use of UUIDs instead of an incrementing ID here. This way the connection IDs are also unique across reboots (useful contract to have when exposing this to plugins)

electronicboy:
UUIDs also generally have the advantage that they're more useful should plugins want to externalise that identity for other systems, especially in a network

xfy2412:
I initially chose long (AtomicLong) thinking it could serve dual purpose as both a connection identifier and a rough connection counter for server operators who use such metrics. But your point about UUIDs being more useful for external systems and surviving reboots makes sense, I'm going to change it

xfy2412:
@electronicboy Done, please review again

WouterGritter:
I think we should shift from a per-connection id to a session id: one UUID that's stable for the player's entire session on the proxy, and shared across the client<->proxy connection and every proxy<->backend connection they open. So I'd name it `sessionId` rather than `connectionId`.

Right now that shared/stable property doesn't actually hold. Since every `MinecraftConnection` generates its own `UUID.randomUUID()` in the constructor:
- each serverbound connection gets a different id, regenerated on every server switch, so it's neither stable nor shared with the client side
- outgoing server-list pings (`VelocityRegisteredServer`) allocate an id they never use

It happens to work for the login-phase events only because the `InboundConnection` wrappers all delegate to the one clientbound `MinecraftConnection`, but that's incidental, not a contract we could promise to plugins.

What I'd propose:
- `@Nullable UUID sessionId` on `MinecraftConnection`, passed in via the constructor
- incoming client connections get a fresh id in `ServerChannelInitializer`: `new MinecraftConnection(ch, server, UUID.randomUUID())`
- serverbound connections copy the player's id in `VelocityServerConnection`: `new MinecraftConnection(channel, server, proxyPlayer.getConnection().getSessionId())`
- outgoing pings pass `null`

The null case is never reachable through `InboundConnection`, so the API can stay non-null by construction:

```java
// VelocityInboundConnection
default UUID getSessionId() {
  return Objects.requireNonNull(getConnection().getSessionId(), "sessionId");
}
```

`LoginInboundConnection` just delegates to `delegate.getSessionId()`, `InitialInboundConnection`/`ConnectedPlayer` get it for free, and `InboundConnection#getSessionId` can be abstract + documented as non-null.

One thing to note is that this shouldn't be confused with Minecraft 26.2's new session ID (a shared "play session" UUID across every connected player that resets when the server goes from empty->non-empty). I'm not too sure if "session id" would be the right name here, but this could probably also be solved by a javadoc comment on `InboundConnection#getSessionId` (maybe `proxySessionId`? seems a bit too verbose).

xfy2412:
Hmmm… indeed, this is a scenario I didn't test before. I only tested multiple players joining the same backend server, but I didn't consider the case of switching between backend servers. I will test this scenario and make the necessary adjustments accordingly. Thanks for pointing this out!

xfy2412:
I also wrote a plugin for test
```
[13:42:36 INFO]: Booting up Velocity 3.5.0-SNAPSHOT (git-48e7519e)...
[13:42:36 INFO]: Connections will use NIO channels, Java compression, Java ciphers
[13:42:36 INFO]: Loading localizations...
[13:42:36 INFO]: Loading plugins...
[13:42:36 INFO]: Loaded plugin cidtest 3.5.0-SNAPSHOT by xfy2412
[13:42:36 INFO]: Loaded 2 plugins
[13:42:36 INFO] [cidtest]: CIDTest Loaded — Testing the stability of getSessionId()
[13:42:36 INFO]: Listening on /[0:0:0:0:0:0:0:0]:25565
[13:42:36 INFO]: Done (0.62s)!
[13:42:54 INFO] [cidtest]: [CIDTEST] PreLogin: xfy2102 connId=e1e309f8-e7c3-424c-8b02-56153af8633f hash=1834913037
[13:42:56 INFO] [cidtest]: [CIDTEST] GameProfileRequest: xfy2102 connId=e1e309f8-e7c3-424c-8b02-56153af8633f hash=1834913037
[13:42:56 INFO]: [connected player] xfy2102 (/192.168.110.1:4564) has connected
[13:42:56 INFO] [cidtest]: [CIDTEST] Login: xfy2102 connId=e1e309f8-e7c3-424c-8b02-56153af8633f uuid=2439bef3-b313-424b-a22c-662e894ffca3
[13:42:56 INFO] [cidtest]: [CIDTEST] ServerPreConnect: xfy2102 connId=e1e309f8-e7c3-424c-8b02-56153af8633f server=hub
[13:42:56 INFO]: [server connection] xfy2102 -> hub has connected
[13:42:59 INFO] [cidtest]: [CIDTEST] ServerPreConnect: xfy2102 connId=e1e309f8-e7c3-424c-8b02-56153af8633f server=login
[13:42:59 INFO]: [server connection] xfy2102 -> login has connected
[13:42:59 INFO]: [server connection] xfy2102 -> hub has disconnected
[13:43:04 INFO] [cidtest]: [CIDTEST] ServerPreConnect: xfy2102 connId=e1e309f8-e7c3-424c-8b02-56153af8633f server=hub
[13:43:04 INFO]: [server connection] xfy2102 -> hub has connected
[13:43:04 INFO]: [server connection] xfy2102 -> login has disconnected
[13:43:06 INFO] [cidtest]: [CIDTEST] Disconnect: xfy2102 connId=e1e309f8-e7c3-424c-8b02-56153af8633f
[13:43:06 INFO]: [connected player] xfy2102 (/192.168.110.1:4564) has disconnected
[13:43:06 INFO]: [server connection] xfy2102 -> hub has disconnected
```

xfy2412:
Let me rename this PR

proxy/src/main/java/com/velocitypowered/proxy/connection/util/VelocityInboundConnection.java
```diff
@@ -19,11 +19,17 @@
...
 /**
  * Base internal interface for a {@link InboundConnection}.
  */
 public interface VelocityInboundConnection extends InboundConnection {
 
   MinecraftConnection getConnection();
+
+  @Override
+  default UUID getSessionId() {
+    return getConnection().getSessionId();
```
WouterGritter: The session ID is ultimately coming from `MinecraftConnection#getSessionId`, which _is_ nullable (pings). `InboundConnection`s javadoc guarantees this returning `never {@code null}`, which is currently the case by circumstances, but a regression in the future might change this. I'd recommend wrapping this with `requireNonNull(...)` to surface a future regression early

api/src/main/java/com/velocitypowered/api/proxy/InboundConnection.java
```diff
@@ -68,4 +69,14 @@ public interface InboundConnection {
    * @return the intent of the connection
    */
   HandshakeIntent getHandshakeIntent();
+
+  /**
+   * Returns the unique, stable session UUID for this connection, generated when
+   * the client initially connects to the proxy. The session id remains consistent
+   * across all login-phase events and all proxy-to-backend connections for the
+   * same player session.
```
WouterGritter: I would like to see a sentence here mentioning this is not to be confused with Mojang's server-wide session ID that's been introduced in 26.2 (im still not entirely sure about the naming here, too)

proxy/src/main/java/com/velocitypowered/proxy/connection/MinecraftConnection.java
```diff
@@ -322,6 +325,17 @@ public Channel getChannel() {
     return channel;
   }
 
+  /**
+   * Returns the session UUID for this connection. Client connections have a
+   * unique session UUID; backend connections share the client's session UUID;
+   * ping connections have {@code null}.
```
WouterGritter: You're kinda sharing implementation details at a point in the code where you don't necessarily know or control this. Since it's an internal method, I'd drop the javadoc

xfy2412:
Maybe call it ”clientConnectionId” will be better?

xfy2412:
My plugin is planning to rely on this PR for development, but it has been pending for four days already. I'd like to confirm whether this feature will be merged in the future, then I could continue developing my plugin (by sessionId or other way). If there are any new comment here, I will still keep updating

xfy2412:
@electronicboy @WouterGritter Should this PR be retargeted to the dev/4.0.0 branch? I saw the latest few PRs submitted to that branch.

PR #1845, opened 2026-07-12
Expose signed player chat metadata through PlayerChatEvent

airgalaxie:
Velocity already parses and validates signed Minecraft player chat internally, but PlayerChatEvent currently exposes only the player and plaintext message to plugins.

Carry the original signed-message metadata across the public API boundary so ecosystem plugins can inspect modern Minecraft signed chat without depending on internal packet classes.

This change is backward compatible and does not alter validation, forwarding, cancellation or rewrite behavior.

### DIFF
--- a/api/build.gradle.kts
+++ b/api/build.gradle.kts
@@ -59,14 +59,14 @@ tasks {
 
         val o = options as StandardJavadocDocletOptions
         o.encoding = "UTF-8"
-        o.source = "21"
+        o.source = "25"
 
         o.use()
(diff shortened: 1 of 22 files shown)


airgalaxie:
This pull request targeted the previous development branch.

The work has since evolved significantly and has been rebased onto the current dev/4.0.0 branch.

I'm closing this PR in favor of the updated implementation, which will continue in a new pull request against dev/4.0.0.

Thank you for your time and previous review.

PR #1849, opened 2026-07-20
Fix GS4 query crash and validate login usernames

### LINKED ISSUE #1827: Offline-mode login accepts arbitrary Unicode in username — no charset validation
### Expected Behavior

Usernames accepted during offline-mode login should be validated against the vanilla character set (alphanumerics and underscore, length 1–16) to prevent log injection and identity confusion.

### Actual Behavior

`ServerLoginPacket.java` (line 84–88) only checks for emptiness:

```java
username = ProtocolUtils.readString(buf, 16);
if (username.isEmpty()) {
    throw EMPTY_USERNAME;
}
```

`readString(buf, 16)` permits any UTF-8 content up to 16 code units. In offline mode, this name becomes the profile name and the offline UUID seed, and is forwarded to the backend as-is. Spaces, control characters (`\n`, `\r`), section signs (`§`), and Unicode homoglyphs are all accepted.

This only affects offline-mode or `forwarding=none`/`legacy` setups. Online-mode names come from Mojang and are already safe, and modern forwarding wraps the name in an HMAC.

### Steps to Reproduce

1. Run Velocity with `online-mode=false` (or `forwarding-mode=none`).
2. Connect with a modified client using a username containing e.g. newlines, `§`, or spaces.
3. The proxy accepts the login and forwards the name to the backend.

### Plugin List

N/A — this is in the core proxy login path.

### Velocity Version

Checked at dev/3.0.0@a7581821 (MC 26.2 era).

### Additional Information

`HandshakeSessionHandler.cleanVhost` (line 192–207) already strips `\0` from the handshake hostname to prevent BungeeCord-style injection, so there's precedent for input sanitization in the login path.

A configurable filter (defaulting to `[A-Za-z0-9_]{1,16}`) would close this gap for standard offline setups while allowing proxies that intentionally support exotic names to opt out. At a minimum, control characters should be stripped before logging.

### LINKED ISSUE #1828: GS4 query handler throws on short/malformed datagrams instead of dropping them
### Expected Behavior

Malformed or too-short UDP datagrams sent to the GS4 query port should be silently dropped.

### Actual Behavior

`GameSpyQueryHandler.channelRead0` (around line 107–119) reads magic bytes, type, and sessionId via `readUnsignedByte()`/`readByte()`/`readInt()` without first checking `readableBytes()`. A datagram shorter than 7 bytes throws `IndexOutOfBoundsException` on every packet, caught and logged at WARNING level:

```
Error whilst handling query packet from <sender>
```

The challenge-token validation for `0x00` stat responses (line 139–143) correctly prevents reflection of the larger response — a spoofed source can't obtain a valid token. So this is just log noise and minor CPU overhead, not a reflection amplification issue.

### Steps to Reproduce

1. Enable `enable-query=true` in `velocity.toml`.
2. Send a short UDP datagram to the query port:
   ```
   echo -n "x" | nc -u <proxy-ip> <query-port>
   ```
3. Each datagram produces a WARNING-level exception in the log.

### Plugin List

N/A — core proxy code with `enable-query=true`.

### Velocity Version

Checked at dev/3.0.0@a7581821

### Additional Information

The fix is straightforward: check `in.readableBytes() >= 7` (2 magic + 1 type + 4 sessionId) at the top of `handleMessage` and return early if the datagram is too short. This matches how the stat path already validates `readableBytes()` at line 146.

nezxenka:
## Summary

Two small defensive fixes for the proxy's network layer.

### 1. Guard GameSpyQueryHandler against short datagrams

GameSpyQueryHandler.channelRead0() reads magic bytes, type, and sessionId
without first checking readableBytes(). A malformed UDP datagram shorter
than 7 bytes causes IndexOutOfBoundsException on every packet, spamming
the log at WARNING level.

This adds a readableBytes() >= 7 guard at the start so short datagrams
are silently dropped instead.

### 2. Validate usernames in ServerLoginPacket

In offline mode, ServerLoginPacket.decode() accepts any UTF-8 content
as a username. This username becomes the offline UUID seed and is
forwarded to the backend as-is, enabling log injection and identity
confusion.

This adds a [A-Za-z0-9_]+ regex check (the vanilla Minecraft character
set), rejecting non-conforming usernames with a QuietDecoderException.

## Testing

./gradlew :velocity-proxy:check passes all checks (compile, Checkstyle,
SpotBugs, tests).

Closes #1828
Closes #1827

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/protocol/netty/GameSpyQueryHandler.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/protocol/netty/GameSpyQueryHandler.java
@@ -109,7 +109,8 @@ public class GameSpyQueryHandler extends SimpleChannelInboundHandler<DatagramPac
     InetAddress senderAddress = msg.sender().getAddress();
 
     // Verify query packet magic
-    if (queryMessage.readUnsignedByte() != QUERY_MAGIC_FIRST
+    if (queryMessage.readableBytes() < 7
+        || queryMessage.readUnsignedByte() != QUERY_MAGIC_FIRST
         || queryMessage.readUnsignedByte() != QUERY_MAGIC_SECOND) {
       return;
     }


electronicboy:
These should be seperate PRs, and the 2nd fix breaks stuff like geyser

nezxenka:
Done — removed the username validation. This PR is now just the GS4 query fix.

WouterGritter:
Please update the PR description accordingly. And throwing the second part of this PR away could be a shame, if you think it's still relevant, I'd encourage you to open it under a second PR (but do see my comment on that PR.. this would likely be a breaking change so I doubt it'd be useful).

Other than "testing" it by compiling & running unit tests, were you able to verify that this indeed fixes the referenced issue? The issue contains a MRE in the form of a command to run.

One issue still:
"7" is a weird magic number here. Is there any way we could calculate this value? Most likely not, but at the very least I'd like this to be a private static field with a short javadoc explaining this is the shorted expected message and which branch reads this shortest possible message (`QUERY_TYPE_HANDSHAKE` / `QUERY_TYPE_STAT`).

PR #1850, opened 2026-07-23
fix: avoid re-fetching connected server in handle(ClientSettingsPacket)

qventymr:
## What
`ClientPlaySessionHandler#handle(ClientSettingsPacket)` fetches
`player.getConnectedServer()`, null-checks it, then discards the local
variable and calls `player.getConnectedServer()` a second time before
calling `.ensureConnected()` on the result.

Between these two calls the player's connected server can change on
another thread (e.g. mid server-switch), so the second call can return
`null` (NPE on `.ensureConnected()`) or a connection whose
`ensureConnected()` throws `IllegalStateException`, crashing/kicking
the session — the same class of bug fixed for the chat/command packet
handlers in f6fbd25 ("Downgrade severity of handling several incoming
user input packet states"), which however did not touch this handler.

## Fix
Reuse the already null-checked local `serverConnection` reference
instead of re-querying `player.getConnectedServer()`.

## Testing
Built locally with `./gradlew build`; no behavioral change for the
non-racy path, since the second `getConnectedServer()` call previously
returned the same object in the common case.

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientPlaySessionHandler.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/connection/client/ClientPlaySessionHandler.java
@@ -237,7 +237,7 @@ public class ClientPlaySessionHandler implements MinecraftSessionHandler {
       // No server connection yet, probably transitioning.
       return true;
     }
-    player.getConnectedServer().ensureConnected().write(packet);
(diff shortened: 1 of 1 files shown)


electronicboy:
I don't think this is an actual improvement outside of flow correction, this stuff, along with server transfer invocations, should occur on the event loop and so there should be no risk of something side-sweeping in here

PR #1853, opened 2026-07-26
Add provides API

Novampr:
Allows plugins to "provide" for other plugins, similar to how it is in Paper


electronicboy:
as is, this would cause getPlugins() to return duplicate instances, should probably migrate that to using the instance map

PR #1854, opened 2026-07-26
refactor: remove outdated messages + fix \! randomly appearing in som…

Real6dq:
This just fixes an annoying bug in the Finnish language, and almost all other languages where \! would appear in a specific language. The messages.properties for localization does not even have \! for the English language, which was weird.
My IDE was also complaining about some unused variables in other languages, so I just removed them.

I just also realized there's a crowdin for velocitypowered, my bad, but this seems faster.

before:
image
after:
image


electronicboy:
Modifications need to be handled via crowdin otherwise they'll just be lost

PR #1859, opened 2026-07-30
Fix dimension ID sign-extension in legacy JoinGamePacket decode for modded 1.7.10

nezxenka:
Fixes #1734 (supersedes it with tests and proper comments).

## Problem

When a modded 1.7.10 server (e.g. GTNH) sends dimension IDs above 127 as
unsigned bytes, Velocity stored them as signed bytes (e.g. 0xB4 → -76).
`RespawnPacket.fromJoinGame()` copies this value and encodes it as a 4-byte
int for pre-1.16 clients, sending an illegal negative dimension ID and
crashing the client.

## Fix

In `JoinGamePacket.decodeLegacy()`, convert unsigned byte values to their
canonical int representation while preserving -1 for the vanilla Nether:

- byte 0xB4 → 180 (not -76)
- byte 0x80 → 128
- byte 0xFE → 254
- byte 0xFF → -1 (Nether takes priority)

The comment explains why this matters — see `RespawnPacket.fromJoinGame()`
and the server-switch flow in `ClientPlaySessionHandler`.

## Tests

`JoinGamePacketTest` — 4 cases:
- Vanilla dimensions (-1, 0, 1)
- Unsigned modded dimensions (180, 128, 254)
- Ambiguous 0xFF → resolves to -1
- `RespawnPacket.fromJoinGame()` copies canonical dimension

## Validation

- [x] `./gradlew :velocity-proxy:build` — compile, checkstyle, spotless: all pass

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/JoinGamePacket.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/JoinGamePacket.java
@@ -245,7 +245,17 @@ public class JoinGamePacket implements MinecraftPacket {
     if (version.noLessThan(ProtocolVersion.MINECRAFT_1_9_1)) {
       this.dimension = buf.readInt();
     } else {
-      this.dimension = buf.readByte();
+      // Vanilla 1.7.10 uses a signed byte for dimension (-1 Nether, 0 Overworld, 1 End).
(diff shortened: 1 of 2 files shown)


WouterGritter:
Squashed #1734 manually in favor of this PR

PR #1860, opened 2026-07-31
Various dependency bumps

WouterGritter:
Bumps all dependencies to their latest versions, except for `checkstyle` which either needs accompanying formatting changes in the source or a modification of the config, as solely bumping the version generates some new style violations.

Fastutil's version bump breaks things with NoClassDefFoundError's as its latest release creates some internal references to classes we exclude due to some compilation magic (8.5.19 is compiled using Java 25, the older 8.5.18 using Java 21, both targeting Java 8 but with different outcomes). Getting rid of the `it/unimi/dsi/fastutil/objects/*ObjectArray*` exclusion alone would fix this. Though rather than maintaining a hacky list of exclusions, we will now be shipping the full fastutil library, not excluding any of its classes. This means plugins will now have access to the full fastutil library, and it will also mean that the Velocity jar grows in size roughly twofold, from ~20mb to ~40mb.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -2,24 +2,24 @@
 configurate3 = "3.7.3"
 configurate4 = "4.2.0"
 flare = "2.0.1"
-log4j = "2.26.0"
+log4j = "2.26.1"
 netty = "4.2.16.Final"
 
 [plugins]
 fill = "io.papermc.fill.gradle:1.0.12"
-shadow = "com.gradleup.shadow:9.5.1"
-spotless = "com.diffplug.spotless:8.2.0"
+shadow = "com.gradleup.shadow:9.6.1"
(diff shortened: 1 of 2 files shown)


kashike (approved):

onebeastchris:
Velocity shipping all of fastutil is super nice! Is that going to be part of API as a stable dependency, or just "expected" to be there via proxy's own depends?;

R00tB33rMan (approved):

WouterGritter:
> Is [fastutil] going to be part of API as a stable dependency, or just "expected" to be there via proxy's own depends?

Paper does something [very similar](https://github.com/PaperMC/Paper/blob/main/paper-api/build.gradle.kts#L45-L55), we will probably adopt that here as well. Out of scope for this PR but definitely something I'd like to add after!

electronicboy (approved):

PR #1861, opened 2026-08-03
Export JSpecify annotations at runtime

jpenilla:
## Summary

Expose `org.jspecify:jspecify` through the API variant using the JSpecify-recommended `api` scope. JSpecify annotations have runtime retention, so `compileOnlyApi` does not model the runtime contract correctly.

## Validation

- `./gradlew check`

### DIFF
--- a/api/build.gradle.kts
+++ b/api/build.gradle.kts
@@ -18,7 +18,7 @@ java {
 }
 
 dependencies {
-    compileOnlyApi(libs.jspecify)
+    api(libs.jspecify)
 
     api(libs.gson)
     api(libs.guava)


R00tB33rMan (approved):

WouterGritter (approved):
It should be noted that we use both jetbrains and checkerframework annotations in the `api` module as well without declaring or shipping them.

For context, it seems like 0993ce2f added this annotation lib, which only needed it for the `@NullMarked` annotation in the new `Ordered` interface. Since then we've used it in one more place internally (`com.velocitypowered.proxy.adventure.RegisteredCallback`).

~~It would be cleaner to get rid of this dependency in favor of jetbrains or checkerframework, but of course this has implications for plugins since we exposed it using `api()` so I don't think this is feasible.~~ It would be out of scope for this PR but I'd really like to see if we can reduce the amount of contract-style annotation dependencies we use.

jpenilla:
The ecosystem is converging on JSpecify, a cleanup should move towards it not away from it.

electronicboy (approved):

PR #1863, opened 2026-08-03
Fix negative ID Dimension in JoinGamePacket

bluegreensea:
This fix negative ID bug by #1734 (https://github.com/PaperMC/Velocity/commit/2676520c6a54bac0529544793c982dd701b338d9) and completly support GTNH (hodgepodge with B:fixLoginDimensionIDOverflow=true)

> I can confirm that after https://github.com/PaperMC/Velocity/commit/2676520c6a54bac0529544793c982dd701b338d9, login in a negative number world (-2 ~ -128) it crash client
> Test with dimension id -7 (and it was map to 249 by that commit):
> ```
> java.lang.IllegalArgumentException: Could not get provider type for dimension 249, does not exist
> 	at Launch//net.minecraftforge.common.DimensionManager.getProviderType(DimensionManager.java:148)
> 	at Launch//net.minecraftforge.common.DimensionManager.createProviderFor(DimensionManager.java:302)
> 	at Launch//net.minecraft.world.WorldProvider.func_76570_a(WorldProvider.java:160)
> 	at Launch//net.minecraft.client.multiplayer.WorldClient.<init>(WorldClient.java:54)
> 	at Launch//net.minecraft.client.network.NetHandlerPlayClient.func_147280_a(NetHandlerPlayClient.java:871)
> 	at Launch//net.minecraft.network.play.server.S07PacketRespawn.func_148833_a(SourceFile:32)
> 	at Launch//net.minecraft.network.play.server.S07PacketRespawn.func_148833_a(SourceFile:13)
> 	at Launch//net.minecraft.network.NetworkManager.func_74428_b(NetworkManager.java:212)
> 	at Launch//net.minecraft.client.multiplayer.PlayerControllerMP.func_78765_e(PlayerControllerMP.java:273)
> 	at Launch//net.minecraft.client.Minecraft.func_71407_l(Minecraft.java:1602)
> 	at Launch//net.minecraft.client.Minecraft.func_71411_J(Minecraft.java:973)
> 	at Launch//net.minecraft.client.Minecraft.func_99999_d(Minecraft.java:898)
> 	at Launch//net.minecraft.client.main.Main.main(SourceFile:148)
> 	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
> 	at java.base/java.lang.reflect.Method.invoke(Method.java:565)
> 	at System//net.minecraft.launchwrapper.Launch.rfb$realLaunch(Launch.java:250)
> 	at System//net.minecraft.launchwrapper.Launch.launch(Launch.java:35)
> 	at System//net.minecraft.launchwrapper.Launch.main(Launch.java:60)
> 	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
> 	at java.base/java.lang.reflect.Method.invoke(Method.java:565)
> 	at com.gtnewhorizons.retrofuturabootstrap.Main.main(Main.java:207)
> 	at com.gtnewhorizons.retrofuturabootstrap.MainStartOnFirstThread.lambda$main$1(MainStartOnFirstThread.java:47)
> 	at java.base/java.lang.Thread.run(Thread.java:1474)
> ```
https://github.com/PaperMC/Velocity/pull/1734#issuecomment-5163023103

### DIFF
--- a/proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/JoinGamePacket.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/JoinGamePacket.java
@@ -245,17 +245,7 @@ public class JoinGamePacket implements MinecraftPacket {
     if (version.noLessThan(ProtocolVersion.MINECRAFT_1_9_1)) {
       this.dimension = buf.readInt();
     } else {
-      // Vanilla 1.7.10 uses a signed byte for dimension (-1 Nether, 0 Overworld, 1 End).
-      // Modded servers hack this to an unsigned byte to allow dim IDs up to 255.
-      // We must store the canonical int here because RespawnPacket.fromJoinGame copies this
-      // value into a packet that encodes dimension as a 4-byte int (RespawnPacket.encode,
-      // pre-1.16 logic). Sign-extending a modded byte like 0xB4 to int -76 sends an
-      // illegal dimension ID to the client and crashes it.
-      //
-      // The wire byte is ambiguous: 0xFF could be vanilla Nether (-1) or modded dim 255.
-      // We resolve in Nether's favor since vanilla 1.7.10 only ever uses -1/0/1.
-      short raw = buf.readUnsignedByte();
-      this.dimension = raw == 0xFF ? -1 : raw;
+      this.dimension = buf.readByte();
     }
     if (version.noGreaterThan(ProtocolVersion.MINECRAFT_1_13_2)) {
       this.difficulty = buf.readUnsignedByte();
--- a/proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/JoinGamePacket.java
+++ b/proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/JoinGamePacket.java
@@ -264,6 +264,12 @@ public class JoinGamePacket implements MinecraftPacket {
     if (version.noLessThan(ProtocolVersion.MINECRAFT_1_15)) {
       this.showRespawnScreen = buf.readBoolean();
     }
+
+    // Hodgepodge
+    if (version.lessThan(ProtocolVersion.MINECRAFT_1_9_1)
+        && buf.readableBytes() >= 4) {
+      this.dimension = buf.readInt();
+    }
   }
 
   private void decode116Up(ByteBuf buf, ProtocolVersion version) {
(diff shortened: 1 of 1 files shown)


WouterGritter:
Thanks for finding this so quickly. If some mods expect a dim id `>= 128` and some use `< 0`, there's really no reliable way to tell the two apart in the `JoinGamePacket` due to both being encoded as a byte.

I'm undecided on whether we should provide explicit support for something that a modpack adds. Is it only this GTNH modpack that adds this protocol-hack fix, or is it more widely used than that?

Regardless, this would need a better comment than `// Hodgepodge` explaining what's going on here (shortest comment would be a single sentence + a reference to this PR). The `dimension < Byte.MIN_VALUE || dimension > Byte.MAX_VALUE` check could also be fragile, a better approach would be to store a boolean flag whether or not the extended/GTNH-specific additional dimension ID was read, and if true, write it back on encode.

electronicboy:
Generally always been against maintaing mod specific hacks because
1) We cater to the vanilla protocol
2) We have 0 concept of mods existing on the wire and this opens up a rabbit hole of stuff which just isn't really maintainable to us

bluegreensea:
> If some mods expect a dim id >= 128 and some use < 0

I think more mods with `​​< 0` are used, because they are supported by vanilla protocol and do not require `Hodgepodge` (or similar mods) to provide support.

> Is it only this GTNH modpack that adds this protocol-hack fix, or is it more widely used than that?

I apologize for that, i am not familiar to writing docs. `Hodgepodge` is the mod that adds this protocol-hack fix, I'm not sure how many modpacks used it, but it can use outside GTNH. I know the check is fragile, but i think `readByte()` won't out the bounds?
edit: Oh, I forgot the `readInt()` value may in byte range.

> We cater to the vanilla protocol

I agree that, but i think current https://github.com/PaperMC/Velocity/commit/2676520c6a54bac0529544793c982dd701b338d9 already break vanilla protocol (vanilla range -128 ~ 127, it was map to -1 ~ 254)

rieverholic:
Even though I mostly play GTNH these days running behind my own Velocity fork, I don't think this fix should be natively supported by Velocity because this is too GTNH specific at this point. This hack isn't really used elsewhere to my knowledge.

bluegreensea:
Closed due to reverted, if someone want int dim id support can build self

PR #1243, opened 2024-02-13
Create a new HttpClient per connection

4drian3d:
This would solve a rare ["too many concurrent streams"](https://discord.com/channels/289587909051416579/908507886420910101/1203719351136550962) error.

PR #1866, opened 2026-08-06
Create a HTTP Client per worker thread instead of per connection

Beaness:
This uses one HttpClient per worker thread instead of creating one for every connection or sharing a single client globally.

This should still avoid the concurrent stream issue from #1243, since requests are distributed across the worker threads. At the same time, each worker can reuse its client and connections, which reduces overhead and avoids repeatedly setting up new TCP connections and TLS handshakes. This saves some network round trips to Mojang and should reduce login times.


electronicboy:
The problem we had is that the connections would die and the http client wouldn't restore them, this IS supposed to be thread safe but the client is so fragile that we end up just recreating them instead. I don't think I could merge this without reputable people testing it in their environments for a long period of time.

PR #1867, opened 2026-08-07
Minecraft 26.3

RealBauHD:


electronicboy:
The way the configuration phase is wired up kinda expects that we understand all of the packets there last I knew, if not stuff gets a bit droppy, I think I had a fix for that somewhere but we should probably add the new post effects class? (we'll probably want to expose this anyways)

RealBauHD:
> The way the configuration phase is wired up kinda expects that we understand all of the packets there last I knew, if not stuff gets a bit droppy, I think I had a fix for that somewhere but we should probably add the new post effects class? (we'll probably want to expose this anyways)

You're right, I will add the post effect packet later and would implement the api inside in another pr

riku6460:
gamemode and previousGamemode in both JoinGamePacket and RespawnPacket are now VarInt.
https://mcsrc.dev/2/diff/26.2/26.3-pre-2/net/minecraft/network/protocol/game/CommonPlayerSpawnInfo

RealBauHD:
> gamemode and previousGamemode in both JoinGamePacket and RespawnPacket are now VarInt. https://mcsrc.dev/2/diff/26.2/26.3-pre-2/net/minecraft/network/protocol/game/CommonPlayerSpawnInfo

Thanks for pointing that out, in fact for all vanilla game modes the var int would write an normal byte, but I will change that to mach vanilla behavior and to not conflict with some hacky mods

proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/ClientboundPostEffectsPacket.java
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  public ClientboundPostEffectsPacket(List<Key> postEffects) {
+    this.postEffects = postEffects;
+  }
+
+  @Override
+  public void decode(ByteBuf buf, Direction direction, ProtocolVersion protocolVersion) {
+    int size = ProtocolUtils.readVarInt(buf);
+    this.postEffects = new ArrayList<>(size);
```
electronicboy, line 44: Mojang's decoder silently caps this at `65536` elements; we should probably mirror that. Annoyingly, we can't entirely trust servers in all environments

electronicboy:
not sure if the next PR wants to be rc2 for allowing people to test but I think this should be bumped to the release protocol version and merged this weekend

proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/ClientboundPostEffectsPacket.java
RealBauHD: What are your thoughts on how I should implement this? Currently, there's only a limit like this for strings; I could just create a ProtocolUtils#readKeyArray(buf, cap) method.

electronicboy: That would work, only real thing to remember is that we would still need to parse all of the elements in the stream if we want to carry on, that way we can actually read the entire packet to the end

R00tB33rMan (approved):

electronicboy (approved):

proxy/src/main/java/com/velocitypowered/proxy/protocol/StateRegistry.java
```diff
@@ -234,31 +236,44 @@ public enum StateRegistry {
           map(0x06, MINECRAFT_1_20_2, false),
           map(0x07, MINECRAFT_1_20_3, false),
           map(0x09, MINECRAFT_1_20_5, false));
+      clientbound.register(
+          ClientboundPostEffectsPacket.class, ClientboundPostEffectsPacket::new,
+          map(0xA, MINECRAFT_26_3, false));
```
WouterGritter:
```suggestion
          map(0x0A, MINECRAFT_26_3, false));
```

style nit: all other places pad this with a zero

RealBauHD: good catch

proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/ClientboundPostEffectsPacket.java
WouterGritter: > Mojang's decoder silently caps this at 65536 elements

https://mcsrc.dev/2/diff/26.3-rc-2/26.2/net/minecraft/network/codec/ByteBufCodecs

It seems like 65536 is just the maximum initial size of the collection, not the cap.

https://mcsrc.dev/2/diff/26.3-rc-2/26.2/net/minecraft/network/codec/ByteBufCodecs#R419
only throws when the size is more than the default limit of `Integer.MAX_VALUE`

WouterGritter: I think this is fine as-is as a defensive limit, so merging for now. It's currently only shared with `ActiveFeaturesPacket`. If it ever causes issues this needs to be increased of course.

PR #1870, opened 2026-08-22
Adding more options for configuring ping passthrough.

TheMiningTeamYT:
This pull request replaces pull request #1519. The existing code has had its commit history cleaned up, brought up to compatibility with the current head of the new active branch dev/4.0.0, and had its configuration options moved into their own section as was previously requested.
What follows is the description from the original pull request.
When using Velocity, I have been frustrated by the lack of fine control over ping passthrough, specifically the lack of support for passing through favicons without passing through the whole ping request. I've also seen people on line asking how to pass through favicons, so I decided to add the ability to control which aspects of server list ping requests are passed through from the back end server, and which aspects aren't, through 5 new boolean configuration options in the config file.


proxy/src/main/java/com/velocitypowered/proxy/config/PingPassthroughMode.java
```diff
@@ -18,11 +18,36 @@
...
 /**
- * Supported passthrough modes for ping passthrough.
+ * Object to contain all the things that can be toggled for ping passthrough.
  */
-public enum PingPassthroughMode {
-  DISABLED,
-  MODS,
-  DESCRIPTION,
-  ALL
+public class PingPassthroughMode {
```
electronicboy: make this a record, and put the default value as a static default inside of here

proxy/src/main/java/com/velocitypowered/proxy/config/migration/PingPassthroughMigration.java
```diff
@@ -0,0 +1,87 @@
...
+ * Migrate the old ping passthrough entry to separate config entries.
+ */
+public final class PingPassthroughMigration implements ConfigurationMigration {
+  @Override
+  public boolean shouldMigrate(final CommentedFileConfig config) {
+    return configVersion(config) < 2.9;
+  }
+
+  @Override
+  public void migrate(final CommentedFileConfig config, final Logger logger) {
```
electronicboy, line 34: This method would probably be cleaner if you just had the existing switch form flip a few booleans in the header

TheMiningTeamYT: Could you elaborate please?

proxy/src/main/java/com/velocitypowered/proxy/config/PingPassthroughMode.java
TheMiningTeamYT: What do you mean "put the default value as a static default"? I'm not too familiar with Java in general or records in particular (I had to google them to even know what you were talking about).

proxy/src/main/java/com/velocitypowered/proxy/config/migration/PingPassthroughMigration.java
electronicboy:
```java
bool a;
bool b;

case (ALL):
a = true
Case(MODS):
b = true
```

And then have the config set calls use the booleans rather than the ternaries

proxy/src/main/java/com/velocitypowered/proxy/config/PingPassthroughMode.java
electronicboy:
```java
public static final PingPassthroughMode DEFAULT = new Ping....
```

That way the default is contained alongside the definitions of all of those booleans

TheMiningTeamYT: I think I understand; you want an instance of ping passthrough mode in the file just to indicate what the defaults are, yes? Or are you wanting me to *use* it in some specific way?

proxy/src/main/java/com/velocitypowered/proxy/config/migration/PingPassthroughMigration.java
TheMiningTeamYT: Ah, thanks for explaining. Makes sense; will do.

proxy/src/main/resources/default-velocity.toml
```diff
@@ -87,6 +84,7 @@ bytes-per-second = -1
 # -1 disables this check.
 decompressed-bytes-per-second = 5242880
 
+
```
R00tB33rMan: This is an incredibly minor nit but this white space should probably be removed

R00tB33rMan (changes requested):

proxy/src/main/resources/default-velocity.toml
TheMiningTeamYT: Done

proxy/src/main/java/com/velocitypowered/proxy/config/migration/PingPassthroughMigration.java
TheMiningTeamYT: Alright done and committed.

proxy/src/main/java/com/velocitypowered/proxy/config/migration/PingPassthroughMigration.java
```diff
@@ -0,0 +1,106 @@
+/*
+ * Copyright (C) 2024 Velocity Contributors
```
WouterGritter: Copyright 2026

proxy/src/main/java/com/velocitypowered/proxy/config/LegacyPingPassthroughMode.java
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright (C) 2018-2023 Velocity Contributors
```
WouterGritter: Copyright 2026

proxy/src/main/java/com/velocitypowered/proxy/config/LegacyPingPassthroughMode.java
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Legacy modes for ping passthrough.
+ */
+public enum LegacyPingPassthroughMode {
+  DISABLED,
+  MODS,
+  DESCRIPTION,
+  ALL
+}
```
WouterGritter: Missing newline

proxy/src/main/java/com/velocitypowered/proxy/config/migration/PingPassthroughMigration.java
```diff
@@ -0,0 +1,106 @@
...
+    );
+
+    // Create ping passthrough entry for the favicon
+    config.set("ping-passthrough.favicon", favicon);
+    config.setComment(
+        "ping-passthrough.favicon",
+        " Should Velocity pass the favicon (also known as the server icon) from the backend server when responding to server list ping requests?"
+    );
+
+    // Create ping passthrough entry for the mods info
```
WouterGritter: I don't think these `// Create ping passthrough entry for the X` comments are adding much value, they can be omitted

proxy/src/main/java/com/velocitypowered/proxy/config/PingPassthroughMode.java
```diff
@@ -18,11 +18,23 @@
...
+
+  public boolean enabled() {
+    return this.version || this.players || this.description || this.favicon
+      || this.modinfo;
+  }
+
+  // Not used, just here to state what the defaults are.
+  public static final PingPassthroughMode DEFAULT = new PingPassthroughMode(
+      false, false, false, false, false);
+}
```
WouterGritter: Missing newline

proxy/src/main/java/com/velocitypowered/proxy/config/PingPassthroughMode.java
```diff
@@ -18,11 +18,23 @@
...
-}
+public record PingPassthroughMode(boolean version, boolean players,
+    boolean description, boolean favicon, boolean modinfo) {
+
+  public boolean enabled() {
+    return this.version || this.players || this.description || this.favicon
+      || this.modinfo;
+  }
+
+  // Not used, just here to state what the defaults are.
```
WouterGritter: > Not used, just here to state what the defaults are.

This is odd, why not use this in VelocityConfiguration line 83?

WouterGritter: ~~Or I'd actually just overload the record constructor with the defaults (all false) and use it in VelocityConfiguration.~~

This doesn't work well with using `DEFAULT` in the `getOrElse`s, but you can ofc do both - `public static final PingPassthroughMode DEFAULT = new PingPassthroughMode();`, either is fine

proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
```diff
@@ -555,9 +557,12 @@ public static VelocityConfiguration read(Path path) throws IOException {
...
               "player-info-forwarding-mode", PlayerInfoForwarding.NONE);
-      final PingPassthroughMode pingPassthroughMode = config.getEnumOrElse("ping-passthrough",
-              PingPassthroughMode.DISABLED);
-
+      final PingPassthroughMode pingPassthrough = new PingPassthroughMode(
+              config.getOrElse("ping-passthrough.version", false),
+              config.getOrElse("ping-passthrough.players", false),
+              config.getOrElse("ping-passthrough.description", false),
+              config.getOrElse("ping-passthrough.favicon", false),
+              config.getOrElse("ping-passthrough.modinfo", false));
```
WouterGritter: More of a nit/fine as is, but here you're splitting the defaults in multiple places. I'd rather be explicit about it and call `PingPassthroughMode.DEFAULT.xxx()` here instead of all the `false`s.

proxy/src/main/java/com/velocitypowered/proxy/connection/util/ServerListPingHandler.java
```diff
@@ -100,60 +101,66 @@ private CompletableFuture<ServerPing> attemptPingPassthrough(VelocityInboundConn
...
-          if (response == fallback) {
-            continue;
-          }
-          Optional<ModInfo> modInfo = response.getModinfo();
-          if (modInfo.isPresent()) {
-            return fallback.asBuilder().mods(modInfo.get()).build();
-          }
+
+        // Why the special handling of the description?
+        // Why not just response.getDescriptionComponent.orElse(null)?
```
WouterGritter: This needs to be resolved (we can't leave this as a comment)

proxy/src/main/java/com/velocitypowered/proxy/config/PingPassthroughMode.java
TheMiningTeamYT: I'm getting conflicting requests here. I'd appreciate it if y'all could decide on what you want me to implement (or just deal with it among yourselves after accepting the pull request).

electronicboy: Not really conflicting, the comment was that the default instance should exist here rather than being an anonymous default instance created elsewhere, this can then be used for the defaults elsewhere

proxy/src/main/java/com/velocitypowered/proxy/connection/util/ServerListPingHandler.java
TheMiningTeamYT: I figured it out. It's a (partial) re-implementation of [this](https://github.com/PaperMC/Velocity/blob/dev/4.0.0/proxy/src/main/java/com/velocitypowered/proxy/connection/util/ServerListPingHandler.java#L111-L115) and [this](https://github.com/PaperMC/Velocity/blob/dev/4.0.0/proxy/src/main/java/com/velocitypowered/proxy/connection/util/ServerListPingHandler.java#L140-L142) I'm guessing the client or the proxy doesn't handle null descriptions all too well.
I'll admit the behavior I implemented isn't identical, so I'll get that fixed..

proxy/src/main/java/com/velocitypowered/proxy/config/PingPassthroughMode.java
TheMiningTeamYT: Actually it might be a good idea to do something like the new [packet limiter code](https://github.com/PaperMC/Velocity/blob/dev/4.0.0/proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java#L1014-L1035)

proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
```diff
@@ -79,7 +80,7 @@ public class VelocityConfiguration implements ProxyConfig {
   @Expose
   private boolean onlineModeKickExistingPlayers = false;
   @Expose
-  private PingPassthroughMode pingPassthrough = PingPassthroughMode.DISABLED;
+  private PingPassthroughMode pingPassthrough = new PingPassthroughMode(false, false, false, false, false);
```
electronicboy:
```java
= PingPassthroughMode.DEFAULT
```

was the primary change I suggested rather than having this randomly created instance;

the blow block of getOrElse() can also use the default instance, that way it's all defined in a singular place

TheMiningTeamYT: Ah okay I understand now. I honestly forgot that was there. That being said, I think change this to function more like the PacketLimiterConfig. I like how it puts all the code for parsing the sections in one place. Unless you disagree.

electronicboy: The limiter config was just a quick toss together for the sake of getting this out, if I wanted everything in one place  I would just have `PingPassthroughMode.of(CommentedConfig)` for that

TheMiningTeamYT: > The limiter config was just a quick toss together for the sake of getting this out
>
Let me ask then, is there a different way you would prefer me to do this?
(Sorry if I come of as a bit annoying, I'm not the most experienced Java developer and I really wanna make sure I do this right so this PR can finally be merged).

proxy/src/main/java/com/velocitypowered/proxy/config/PingPassthroughMode.java
TheMiningTeamYT: Ah okay.

TheMiningTeamYT:
By the way, out of the scope this pull request, but while working on it I discovered that your gitignore doesn't include VS Code's workspace config directory (.vscode). Just thought y'all should know. I may open a pull request for this at some point.

proxy/src/main/java/com/velocitypowered/proxy/config/migration/PingPassthroughMigration.java
TheMiningTeamYT: Done.

proxy/src/main/java/com/velocitypowered/proxy/config/LegacyPingPassthroughMode.java
TheMiningTeamYT: Done.

proxy/src/main/java/com/velocitypowered/proxy/config/LegacyPingPassthroughMode.java
TheMiningTeamYT: Done.

proxy/src/main/java/com/velocitypowered/proxy/config/migration/PingPassthroughMigration.java
TheMiningTeamYT: I'll be honest, I agree with you. They have now been removed.

proxy/src/main/java/com/velocitypowered/proxy/config/PingPassthroughMode.java
TheMiningTeamYT: Fixed.

proxy/src/main/java/com/velocitypowered/proxy/config/PingPassthroughMode.java
TheMiningTeamYT: Alright, done. I ended up implementing both the overloaded constructor and the default PingPassthroughMode as @WouterGritter suggested. However, I kept the initialization of pingPassthrough in VelocityConfiguration.java line 83 as `private PingPassthroughMode pingPassthrough = PingPassthroughMode.DEFAULT;` since that's more in line with how the rest of the code is written.

proxy/src/main/java/com/velocitypowered/proxy/config/VelocityConfiguration.java
TheMiningTeamYT: Done in the new config reading code in PingPassthroughMode.java.

proxy/src/main/java/com/velocitypowered/proxy/connection/util/ServerListPingHandler.java
TheMiningTeamYT: It's been resolved, to my satisfaction at least.

proxy/src/main/java/com/velocitypowered/proxy/config/PingPassthroughMode.java
```diff
@@ -16,13 +16,49 @@
...
 
 /**
- * Supported passthrough modes for ping passthrough.
+ * Object to contain all the things that can be toggled for ping passthrough.
+ *
+ * @param version Whether the version should be passed through.
+ * @param players Whether the player count should be passed through.
+ * @param description Whether the description should be passed through.
+ * @param favicon Whether the favicon should be passed through.
+ * @param modinfo Whether the modinfo should be passed through.
```
WouterGritter:
```suggestion
 * @param version     Whether the version should be passed through.
 * @param players     Whether the player count should be passed through.
 * @param description Whether the description should be passed through.
 * @param favicon     Whether the favicon should be passed through.
 * @param modinfo     Whether the modinfo should be passed through.
```

proxy/src/main/java/com/velocitypowered/proxy/config/PingPassthroughMode.java
```diff
@@ -16,13 +16,49 @@
...
-  DESCRIPTION,
-  ALL
+public record PingPassthroughMode(boolean version, boolean players,
+    boolean description, boolean favicon, boolean modinfo) {
+  public static final PingPassthroughMode DEFAULT = new PingPassthroughMode();
+  
+  /**
+   * Creates a default PingPassthroughMode.
+   */
+  PingPassthroughMode() {
```
WouterGritter:
```suggestion
  private PingPassthroughMode() {
```

proxy/src/main/resources/default-velocity.toml
```diff
@@ -43,29 +43,14 @@ forwarding-secret-file = "forwarding.secret"
 # Announce whether or not your server supports Forge. If you run a modded server, we
 # suggest turning this on.
 # 
-# If your network runs one modpack consistently, consider using ping-passthrough = "mods"
+# If your network runs one modpack consistently, consider using ping-passthrough.modinfo = true
```
WouterGritter, line 46: This comment could also be changed in the `PingPassthroughMigration` migration.

proxy/src/main/java/com/velocitypowered/proxy/config/migration/PingPassthroughMigration.java
```diff
@@ -0,0 +1,100 @@
...
+  @Override
+  public boolean shouldMigrate(final CommentedFileConfig config) {
+    return configVersion(config) < 2.9;
+  }
+
+  @Override
+  public void migrate(final CommentedFileConfig config, final Logger logger) {
+    // Get legacy ping passthrough value
+    final LegacyPingPassthroughMode legacyMode = config.getEnumOrElse("ping-passthrough",
+        LegacyPingPassthroughMode.DISABLED);
```
WouterGritter: The `LegacyPingPassthroughMode` only exists to aid this migration class. I would rather get rid of this enum and use raw strings here. This maps cleanly in your switch statement.

proxy/src/main/java/com/velocitypowered/proxy/config/LegacyPingPassthroughMode.java
```diff
@@ -0,0 +1,28 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package com.velocitypowered.proxy.config;
+
+/**
+ * Legacy modes for ping passthrough.
+ */
+public enum LegacyPingPassthroughMode {
```
WouterGritter: Can be removed, [see my other comment](https://github.com/PaperMC/Velocity/pull/1870/changes#r3903934676).

proxy/src/main/java/com/velocitypowered/proxy/config/migration/PingPassthroughMigration.java
```diff
@@ -0,0 +1,100 @@
...
+        modinfo = true;
+        break;
+      case DISABLED:
+        break;
+      default:
+        break;
+    }
+
+    config.removeComment("ping-passthrough");
+    config.remove("ping-passthrough");
```
WouterGritter, line 64: ~~I'd imagine calling `config.removeComment()` is redundant here, please confirm and if so line 65 can be omitted.~~

Tested, not redundant as `ping-passthrough` existed previously.

proxy/src/main/java/com/velocitypowered/proxy/config/PingPassthroughMode.java
```diff
@@ -16,13 +16,49 @@
...
+  }
+
+  /**
+   * Returns a PingPassthroughMode from a config section, or the default if the section is null.
+   * Based on the code for PacketLimiterConfig.
+   *
+   * @param config The configuration object to parse.
+   * @return The PingPassthroughMode, or the default if {@code config} is null.
+   */
+  public static PingPassthroughMode fromConfig(CommentedConfig config) {
```
WouterGritter: `CommentedConfig` can be reduced to its parent `Config` here.

WouterGritter (approved):

Issue #1876, opened 2026-09-03
Very unintuitive error message for old configs with `ping-passthrough`


TopchetoEU:
### Expected Behavior

Old configs with a string value for `ping-passthru` should throw a more intuitive error or use the string value for a preset instead.

### Actual Behavior

After #1870, old configs with `ping-passthru` break with the very cryptic error message of:

java.lang.ClassCastException: class java.lang.String cannot be cast to class com.electronwill.nightconfig.core.Config

### Steps to Reproduce

Make a config 

### Plugin List

Not relevant

### Velocity Version

Latest, commit SHA 21bbf35d1da1d8a74aad897ecb4e6915f121898d

### Additional Information

This is to be expected, as we're trying to pass a String to a parameter, that accepts Config, but it took me a good while of digging to find the culprit.

If it were up to me to fix this, I'd define some presets, which a string value can correlate to, so that old configs don't break and backwards compatibility is retained

electronicboy:
Can you provide such a config? This stuff should be handled with the migration logic, so you shouldn't see this as an error unless you're trying to add that config value to a new config?

TopchetoEU:
```
config-version = "3.0"
bind = "0.0.0.0:25565"
online-mode = false
force-default-server = false
show-motd = false
ping-passthrough = "all"
player-info-forwarding-mode = "modern"
forwarding-secret-file = "/var/run/secrets/velocity-secret"
motd = ""
show-max-players = 0
enable-player-address-logging = true

[servers]
survival = "mc-survival"
creative = "mc-creative"
try = []

[forced-hosts]
"survival.[REDACTED]" = ["survival"]
"creative.[REDACTED]" = ["creative"]
```

The error I get, specifically, is this:

```
[13:38:01 INFO]: Booting up Velocity 4.1.2-SNAPSHOT (git-21bbf35d-b27)...
[13:38:01 INFO]: Connections will use epoll channels, libdeflate (Linux x86_64) compression, OpenSSL 3.x.x (Linux x86_64) ciphers
[13:38:01 ERROR]: Unable to read/load/save your velocity.toml. The server will shut down.
java.lang.ClassCastException: class java.lang.String cannot be cast to class com.electronwill.nightconfig.core.Config (java.lang.String is in module java.base of loader 'bootstrap'; com.electronwill.nightconfig.core.Config is in unnamed module of loader 'app')
[...]

PR #1878, opened 2026-09-04
Refactor/migrate velocity.toml to velocity.yml

WouterGritter:
This PR takes a different angle than #1824, focusing on keeping the configuration library abstract rather than fixed. This is done with two interfaces of our own:

- `Configuration`, a simple key-value store that maps string keys to primitive types (and `String`/`List`). A nested section is itself a `Configuration`, so it is read the same way a whole file is.
- `ConfigurationDocument extends Configuration`, adding methods to modify the configuration and to read and write comments, both of which migrations need.

This also introduces `LayeredConfiguration`, which layers multiple configurations on top of each other, returning the first config that defines a value when reading from it. This allows us to have only a single place to put our defaults: in `default-velocity.yml`. Previously we had **three** places where defaults were defined: `VelocityConfiguration`'s fields, `VelocityConfiguration`'s `read()` calls and `default-velocity.toml`, leaving discrepancies for some fields. Five of them disagreed: `announce-forge`, `command-rate-limit`, `kick-after-rate-limited-commands`, `tab-complete-rate-limit` and `kick-after-rate-limited-tab-completes`.

Together with the `Configuration` interface, `LayeredConfiguration` allows us to implement an environment-variable backed configuration in the future, where users can choose to override config values from the environment. This is useful for users running Velocity in Docker, who might want to configure only a minimal set of properties, leaving everything else at Velocity's default (inspired by https://github.com/PaperMC/Velocity/issues/1876#issuecomment-5526656965).

### Migrating
An existing `velocity.toml` is converted to `velocity.yml` on first start and kept as `velocity.toml.bak`. All seven legacy TOML migrations run first (if required), so a configuration from any version converts. `config-version` is an integer from now on, starting at `3` (superseding the last toml version `2.9`).

### YAML library
YAML is read with `org.spongepowered:configurate-yaml` from [GeyserMC's Configurate fork](https://github.com/GeyserMC/Configurate), which supports comments. It needs `https://repo.opencollab.dev/maven-snapshots/` and moves the project to that version of Configurate 4; it relocates snakeyaml internally, so it does not collide with the one the API exposes. `YamlDocument` is the only class that names it, allowing easy future swapping of the config lib.


WouterGritter:
One thing I'm concerned about is people running Velocity in Docker, exposing `velocity.toml` as a volume. We need to make sure this won't delete or empty it, then store everything in `velocity.yml`, only for it to be destroyed when the container shuts down.

We also need to give Docker users instructions on how to update, we can't have them run Velocity outside of Docker just for this migration. It might be as straight forward as asking users to create an empty `velocity.yml` file and pass that to Velocity as a volume. Moving `velocity.toml` -> `velocity.toml.bak` would most likely fail (if the file itself is passed through as a volume instead of the parent dir), and `velocity.yml` would already exist as an empty file, so we need to handle both.

If this is too much hassle, we could introduce an argument:
```bash
java -jar velocity.jar --migrate-config [source (defaults to velocity.toml or velocity.yml, whichever it finds)] [target (default velocity.yml)]
```
which just migrates and exists. Could be useful outside of this YML migration for any future migration.

PR #1879, opened 2026-09-07
feat: add write timeout handling

dwarslooper:
Adds a write timeout for proxy connections, alongside the existing read timeout.

This closes connections that remain unable to write for too long, preventing stalls.
(timeout defaults to 30 seconds, same as existing read timeout)


Uhuli (approved):

WouterGritter:
The `WriteTimeoutHandler` placement worries me on the hot path. `MinecraftConnection#delayedWrite` deliberately passes `channel.voidPromise()`, and every backend -> client packet goes through it (`BackendPlaySessionHandler#handleGeneric`/`handleUnknown`). Netty's handler opens with `promise = promise.unvoid()`, which allocates a fresh `DefaultChannelPromise`, plus a `WriteTimeoutTask` and a `ScheduledFutureTask` that gets inserted into the timer queue and cancelled again _per packet_. On top of that the outbound buffer now has to notify a real promise on every flush, which `voidPromise()` short-circuits today.

What I'm mostly stuck on though is what this buys us. As far as I can tell the stuck client case already dies: the client channel goes unwritable -> `ClientPlaySessionHandler#writabilityChanged` disables autoread on the backend -> `ReadTimeoutHandler` isn't autoread-aware, so the backend trips its own read timeout ~30s later. Same kill, same timing, the only thing that's actually wrong is the log line, which blames a perfectly healthy backend for a client that stopped draining its socket.

If that's the real problem, it seems fixable where it's reported, instead of with a second timeout plus a new `ProxyConfig` method and config key:
```java
// BackendPlaySessionHandler#exception
if (throwable instanceof ReadTimeoutException && !playerConnection.getChannel().isWritable()) {
  // backend is fine, we stopped reading it because the player isn't draining
}
```

Am I missing a case where nothing closes the connection today? The only one I can come up with is writes that have no paired connection to backpressure (a plugin looping writes to a stuck player), which is unbounded, but also needs a plugin bug to hit.

PR #1881, opened 2026-09-13
Forward plugin messages during server transitions

HP-network:
When switching servers, `connectedServer` is cleared before the new backend connection is installed. Client plugin messages received in that window were only sent to the in-flight connection for the legacy Forge handshake channel; other channels were dropped.

Use the in-flight connection whenever there is no current connected server. The existing backend state and connection-phase checks still decide whether the packet is forwarded or queued.

The regression test covers an ordinary plugin channel during the transition window.

Tests:
- `./gradlew :velocity-proxy:test --tests com.velocitypowered.proxy.connection.client.ClientPlaySessionHandlerTest --no-daemon --console=plain`
- `./gradlew :velocity-proxy:test --no-daemon --console=plain`
- `./gradlew :velocity-proxy:check --no-daemon --console=plain`


WouterGritter:
Some bookkeeping:
- Issue #1766 describes this exact bug (the `getConnectionInFlight()` fallback being gated on `FORGE_LEGACY_HANDSHAKE_CHANNEL`) and proposes the same fix.
- PR #1783 (fixing this issue) contains a functionally identical change to `ClientPlaySessionHandler.handle(PluginMessagePacket)`.

The differences are that #1783 also bundles an unrelated handshake hostname length change, while this PR is scoped to just the plugin-message fix and adds a regression test.

HP-network:
Thanks for checking. I agree that #1783 covers the same behavior. This PR is scoped to the plugin-message fix and includes a regression test, but I do not want to maintain a duplicate. Please keep whichever version is preferred and close this one if #1783 is the better path.

HP-network:
Closing this in favor of #1783, which covers the same plugin-message transition fix.

PR #1882, opened 2026-09-13
Fix config reload not removing deleted servers

Rainnny7:
When reloading velocity.toml, unregister servers that were removed from the config and evacuate any connected players to fallback servers.


proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
```diff
@@ -488,9 +488,29 @@ public boolean reloadConfiguration() throws IOException {
       return false;
     }
 
-    // Re-register servers. If a server is being replaced, make sure to note what players need to
-    // move back to a fallback server.
+    // Re-register servers. If a server is being replaced or removed, make sure to note what
+    // players need to move back to a fallback server.
     Collection<ConnectedPlayer> evacuate = new ArrayList<>();
+    for (Map.Entry<String, String> entry : configuration.getServers().entrySet()) {
+      if (isServerConfigured(newConfiguration.getServers(), entry.getKey())) {
```
WouterGritter:
```java
newConfiguration.getServers().keySet().stream()
    .anyMatch(name -> name.equalsIgnoreCase(entry.getKey()))
```

in-line looks neat enough and gets rid of the helper method

proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
```diff
@@ -488,9 +488,29 @@ public boolean reloadConfiguration() throws IOException {
...
+        continue;
+      }
+
+      Optional<RegisteredServer> rs = servers.getServer(entry.getKey());
+      if (rs.isEmpty()) {
```
WouterGritter: If `rs.get()` gets called multiple times like is the case here, I generally prefer a pattern like the following more:
```java
RegisteredServer rs = servers.getServer(entry.getKey()).orElse(null);
if (rs == null) {
  continue;
}
```

proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
```diff
@@ -488,9 +488,29 @@ public boolean reloadConfiguration() throws IOException {
...
+        continue;
+      }
+
+      for (Player player : rs.get().getPlayersConnected()) {
+        if (!(player instanceof ConnectedPlayer)) {
+          throw new IllegalStateException("ConnectedPlayer not found for player " + player
+              + " in server " + rs.get().getServerInfo().getName());
```
WouterGritter: This exception message seems a bit misleading, the error isn't "a (connected) player cant be found for player X", it's that somehow a non-ConnectedPlayer Player made it into a RegisteredServer's player-list, which should never be the case.

You can probably throw just a bare ISE here, but a message like "Expected ConnectedPlayer" would be a bit cleaner.

Though at line 502 / 522, the same misleading exception is thrown. I'll leave it up to you to either keep the same message for a shorter diff, or clean up the existing exception too.

Rainnny7: appreciate the feedback. will do these changes for you shortly and commit

WouterGritter:
One minor edge case found by Claude (pre-existing pattern, not blocking). Dropping it here for future reference, it's out of scope for this PR. Not verified but seems legit.

Evacuation runs before `this.configuration = newConfiguration`, so `getNextServerToTry()` (`ConnectedPlayer.java:871`) walks the old try list. The player's current (removed) server is skipped via `hasSameName(connectedServer...)`, but if two servers in the old try list were removed in the same reload and the player sits on one of them, `server.getServer(toTryName)` for the other returns empty and the player is kicked with `"could not move you to any fallback servers"` instead of trying the next entry. The same shape of issue exists with the replace path (and `serversToTry` is cached per player anyway).

proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
Rainnny7: ive gone ahead and fixed this. how is it now?

proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
Rainnny7: ive gone ahead and fixed this. how is it now?

proxy/src/main/java/com/velocitypowered/proxy/VelocityServer.java
Rainnny7: ive gone ahead and fixed this. how is it now?

Rainnny7:
> One minor edge case found by Claude (pre-existing pattern, not blocking). Dropping it here for future reference, it's out of scope for this PR. Not verified but seems legit.
> 
> Evacuation runs before `this.configuration = newConfiguration`, so `getNextServerToTry()` (`ConnectedPlayer.java:871`) walks the old try list. The player's current (removed) server is skipped via `hasSameName(connectedServer...)`, but if two servers in the old try list were removed in the same reload and the player sits on one of them, `server.getServer(toTryName)` for the other returns empty and the player is kicked with `"could not move you to any fallback servers"` instead of trying the next entry. The same shape of issue exists with the replace path (and `serversToTry` is cached per player anyway).

yes i believe you are correct, ive looked into this and it does appear that the edge case is possible. seeing as its out of scope for this PR, would you like me to open another PR to resolve this? or how should we go about that?

WouterGritter:
> yes i believe you are correct, ive looked into this and it does appear that the edge case is possible. seeing as its out of scope for this PR, would you like me to open another PR to resolve this? or how should we go about that?

If you want to spend some time on it, a separate PR is the way to go. I don't think this race is easy to hit at all, but for correctness sake a PR is always welcome.

Rainnny7:
> > yes i believe you are correct, ive looked into this and it does appear that the edge case is possible. seeing as its out of scope for this PR, would you like me to open another PR to resolve this? or how should we go about that?
> 
> If you want to spend some time on it, a separate PR is the way to go. I don't think this race is easy to hit at all, but for correctness sake a PR is always welcome.

alrighty sounds good, ill make another PR to resolve this

WouterGritter:
Awesome! And thanks for this fix! Seems like this has been an issue for a really long time.

Rainnny7:
> Awesome! And thanks for this fix! Seems like this has been an issue for a really long time.

not a problem! ive gone ahead and made a new PR with a proposed fix: https://github.com/PaperMC/Velocity/pull/1883

PR #1883, opened 2026-09-14
Fix reload evacuation fallback when try-list servers are removed

Rainnny7:
Follow up to an issue mentioned in #1882 (https://github.com/PaperMC/Velocity/pull/1882#issuecomment-5666398455):

During reload evacuation, fallback routing could fail when multiple try-list servers were removed or replaced in the same reload. Evacuation ran before the new configuration was applied, so getNextServerToTry() walked the old try list. If another removed entry appeared before a valid fallback, server.getServer() returned empty and the player was disconnected instead of continuing to the next server


Rainnny7:
just bumping this, hoping to have it merged 🙏🏼

PR #1885, opened 2026-09-16
Sanitize control characters in connection logs

### LINKED ISSUE #1827: Offline-mode login accepts arbitrary Unicode in username — no charset validation

HP-network:
Fixes #1827

### What changed
Escape control characters in the player representation used by connection logs. Newlines, tabs, carriage returns, and other ISO control characters are rendered as escaped text instead of changing the log layout.

### Why
Offline-mode usernames can contain control characters. Sanitizing the log representation removes log injection without rejecting names that integrations may legitimately use.

### Testing
- `./gradlew :velocity-proxy:test --tests com.velocitypowered.proxy.util.CharacterUtilTest --no-daemon --console=plain`
- `./gradlew :velocity-proxy:check --no-daemon --console=plain`
- `git diff --check`


electronicboy:
I fail to understand the legitimate use for this, I can understand why platforms like geyser will use characters not usable by normal names, but they are at least addressable, names with control characters are inaccessible to players, can cause protocol state issues elsewhere, I feel that we should just reject them entirely

WouterGritter:
Related:
- #1827
- #1849

I agree, I think it makes more sense to reject players joining with characters like `\n`, `\r`, `\t`, etc. We need to be careful that we're more lax than mojang's restrictions though, as disallowing players to join with `.` or `+` in their name would break plugins like Geyser. Though we'd need to draw the line somewhere... I believe geyser makes it possible to configure this prefix char. Servers could be using weird characters like `$`, `§`, `%`, I don't necessarily think we should block these. Maybe the line is at printable/non-white-space characters?

HP-network:
Thanks, I agree that escaping the log representation alone does not address the protocol-state issue. A login-time check for control and whitespace characters sounds like the right direction, while keeping printable symbols such as `.` and `+` available for integrations. Would limiting validation to control/whitespace characters match the project's intended boundary?

PR #1886, opened 2026-09-18
Fix stale absorption hearts on 1.8 server switches

EthanKeating:
Vanilla 1.8 clients can keep displaying absorption hearts from the previous backend after a server switch. For example, eating a golden apple and then switching to a hub with no absorption leaves stale yellow hearts above the hotbar until a later metadata update clears them.

The 1.8 client copies entity metadata during respawn. A destination backend whose absorption remains at its default zero may not send a changed absorption value, leaving the copied value visible.

Reset the destination player's absorption metadata to zero immediately after the switch packets, before destination play packets are forwarded. The reset applies only to protocol 1.8 and only to server switches, covers both switch paths, and allows subsequent destination metadata to supply the actual absorption value.

The reset is represented by `LegacyPlayerAbsorptionPacket`, an encode-only `MinecraftPacket` registered for protocol 1.8 in `StateRegistry`. `MinecraftEncoder` supplies the packet ID and serializes the payload; backend entity metadata continues to pass through without decoding.

Validation:

- `./gradlew build` on both the PvPRivals fork and current upstream base, including existing tests, Checkstyle, and Spotless.
- Local vanilla 1.8 golden-apple/server-switch testing with the backend workaround disabled; the tester confirmed the proxy fix resolved the visual issue.
- A protocol 1.8 connection verified login, `/hub`, and the zero-absorption metadata packet for the destination entity with the patch enabled. The same connection check succeeded without the patch and observed no standalone reset packet.

- Verified the typed packet through Netty's embedded encoder/decoder: expected wire payload, registration limited to 1.8, and unchanged backend metadata forwarding.


electronicboy:
We are 100% not sending raw bytes down the pipeline

